PR #9, opened 2024-08-06
[Fix] Image URL in README.md

piedcipher:
on pub, image from the readme in not loading due to short URL.

Screenshot_20240807-003055

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 3D library for Flutter.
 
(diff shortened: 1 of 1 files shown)


bdero:
This is already fixed in the [most recent version](https://pub.dev/packages/flutter_scene/versions/0.2.0) pushed to pub. The problem was the use of HTML for sizing/centering the image rather than markdown.

PR #11, opened 2024-08-12
Make `flutter_app` work on flutter `master` channel

domesticmouse:
The `flutter_app` example code doesn't compile and run as is on checkout. The dependencies fail to resolve, and the app requires the `DamagedHelmet.glb` model which isn't part of the checkout.

### DIFF
--- a/examples/flutter_app/pubspec.yaml
+++ b/examples/flutter_app/pubspec.yaml
@@ -12,7 +12,8 @@ dependencies:
     sdk: flutter
   flutter_scene:
     path: ../../
-  flutter_scene_importer: ^0.1.2-1
+  flutter_scene_importer:
+    path: ../../importer
   native_assets_cli: ^0.7.2
   vector_math: ^2.1.4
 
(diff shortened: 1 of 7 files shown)


domesticmouse (commented):
The `flutter_app` example code doesn't compile and run as is on checkout. The dependencies fail to resolve, and the app requires the `DamagedHelmet.glb` which isn't part of the checkout.

examples/flutter_app/pubspec.yaml
```diff
@@ -12,8 +12,7 @@ dependencies:
     sdk: flutter
   flutter_scene:
     path: ../../
-  flutter_scene_importer:
-    path: ../../importer
```
bdero, line 16: I'd rather keep the local path here so I don't keep having to update it.

bdero:
Ah yeah, this is a leak from my stage because I didn't want to check in large models. Now that I've replaced the importer scripts, maybe some CI stuff is in order soon.
I'm also considering adding a bootstrap script that downloads all the glTF sample assets.

domesticmouse:
Do you want me to trim this down to parts you are willing to land, or should I just close it?

bdero:
The former. Everything LGTM except the dependency pin. Apologies for the ambiguity.

domesticmouse:
Reverted the dependency pin.

bdero (approved):
LGTM

Issue #13, opened 2024-08-20
Document all the things


bdero:
Lots of docstrings are missing from the public API. Although the API is early (at the time of writing), we should make it a habit to add comprehensive docstrings everywhere.

PR #15, opened 2024-08-21
Provide doc comments for node.dart

kwill39:
Partially addresses #13

### DIFF
--- a/lib/node.dart
+++ b/lib/node.dart
@@ -60,12 +60,9 @@ base class Node implements SceneGraph {
   Node? get parent => _parent;
   bool _isRoot = false;
 
-  /// The collection of vertices, edges, and faces that represent the geometry of this node,
-  /// along with the material that defines the object's appearance.
+  /// The collection of [MeshPrimitive] objects that represent the 3D geometry and material properties of this node.
   ///
-  /// The mesh defines the shape of the object through its geometry, which is composed of vertices,
-  /// edges, and faces. The mesh may also include a material, which specifies how the surface
-  /// of the object interacts with light, including properties like color, texture, and reflectivity.
+  /// This property is `null` if this node does not have any associated geometry or material.
   Mesh? mesh;
 
   /// Whether this node is a joint in a skeleton for animation.
--- a/lib/node.dart
+++ b/lib/node.dart
@@ -13,12 +13,21 @@ import 'package:flutter_scene/scene_encoder.dart';
 import 'package:flutter_scene_importer/importer.dart';
 import 'package:flutter_scene_importer/flatbuffer.dart' as fb;
 
+/// A `Node` represents a single element in a 3D scene graph.
+///
+/// Each node can contain a transform (position, rotation, scale), a mesh (3D geometry and material),
+/// and child nodes. Nodes are used to build complex scenes by establishing relationships
+/// between different elements, allowing for transformations to propagate down the hierarchy.
 base class Node implements SceneGraph {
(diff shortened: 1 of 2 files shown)


kwill39:
Please determine the accuracy of the doc comment for the `render` method specifically.

kwill39:
Hey @bdero no rush to merge this. I just wanted to see if the format works for you though before I started documenting the other libraries.

bdero:
Whoops, sorry for the late reply. Taking a look now.

lib/node.dart
```diff
@@ -42,18 +54,39 @@ base class Node implements SceneGraph {
...
+  /// The parent node of this node in the scene graph.
   Node? get parent => _parent;
   bool _isRoot = false;
 
+  /// The collection of vertices, edges, and faces that represent the geometry of this node,
+  /// along with the material that defines the object's appearance.
+  ///
+  /// The mesh defines the shape of the object through its geometry, which is composed of vertices,
+  /// edges, and faces. The mesh may also include a material, which specifies how the surface
+  /// of the object interacts with light, including properties like color, texture, and reflectivity.
```
bdero, line 67: A `Mesh` is actually a list of `MeshPrimitives`, each of which contain a `Geometry` and a corresponding `Material` to apply. So this would actually be a pretty good description of `MeshPrimitive`, but isn't quite accurate for `Mesh`.

lib/node.dart
```diff
@@ -42,18 +54,39 @@ base class Node implements SceneGraph {
...
+  /// final node = await Node.fromAsset('path/to/asset.model');
+  /// ```
   static Future<Node> fromAsset(String assetPath) async {
     final buffer = await rootBundle.load(assetPath);
     return fromFlatbuffer(buffer);
   }
 
+  /// FlatBuffers are a compact binary serialization format, commonly used for efficiently storing complex 3D data.
+  ///
+  /// If you have a file containing your 3d model, consider using [fromAsset] instead.
```
bdero, line 89:
```suggestion
  /// Deserialize a model from Flutter Scene's compact model format.
  ///
  /// If you're using [Flutter Scene's offline importer tool](https://pub.dev/packages/flutter_scene_importer),
  /// consider using [fromAsset] to load the model directly from the asset bundle instead.
```

lib/node.dart
```diff
@@ -194,6 +240,15 @@ base class Node implements SceneGraph {
...
 
+  /// Prepares and renders this node and its children within the scene graph.
+  ///
+  /// This method calculates the final world transform for the node by combining the
+  /// parent's world transform with the node's local transform. It then renders the node's
+  /// mesh primitives using the provided `SceneEncoder`, and recursively renders all child nodes.
+  ///
+  /// The rendering process involves encoding the node's geometry and material data into commands
+  /// that are passed to the GPU for drawing. If the node has associated skinning data (for animated models),
+  /// this data is also passed along to the rendering pipeline.
```
bdero, line 251:
```suggestion
  /// Recursively records mesh draw operations for this node and all its children.
  ///
  /// To display this node in a dart:ui canvas, add this node to a `Scene` and call `Scene.render` instead.
```

Perhaps we should keep the docstring for this one simple. Users won't normally call this directly. We can guide them to using `Scene.render` instead.

bdero (commented):
Looking great. Just a few improvement suggestions. :)

kwill39:
@bdero, all set. Take another look and let me know what you think 🚀

bdero (approved):
LGTM

Issue #38, opened 2024-11-05
cant build flutter_scene


wangcker:
hello,
I try to build scene lib,but it throws:

```
Unhandled exception:
Exception: CMake generate step failed (exit code 1):
STDERR: [ 11%] Performing update step for 'flatbuffers-populate'
CMake Error at /Users/**/code/flutter_scene/importer/build/_deps/flatbuffers-subbuild/flatbuffers-populate-prefix/tmp/flatbuffers-populate-gitupdate.cmake:34 (message):
  Failed to get the hash for HEAD:

  fatal: not a git repository: '.git'
```

flutter doctor -v
```
[✓] Flutter (Channel master, 3.27.0-1.0.pre.352, on macOS 14.6.1 23G93 darwin-arm64, locale zh-Hans-CN)
    • Flutter version 3.27.0-1.0.pre.352 on channel master at /Users/wangcker/opt/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 8591d0c16a (13 hours ago), 2024-11-04 20:46:29 +0800
    • Engine revision 05cb5d7f79
    • Dart version 3.7.0 (build 3.7.0-95.0.dev)
    • DevTools version 2.40.1

[✓] Android toolchain - develop for Android devices (Android SDK version 34.0.0)
    • Android SDK at /Users/wangcker/Library/Android/sdk
    • Platform android-35, build-tools 34.0.0
    • Java binary at: /Applications/Android Studio.app/Contents/jbr/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 17.0.11+0-17.0.11b1207.24-11852314)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS (Xcode 16.1)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Build 16B40
    • CocoaPods version 1.15.2

[✓] Chrome - develop for the web
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[✓] Android Studio (version 2024.1)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Flutter plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/9212-flutter
    • Dart plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/6351-dart
    • Java version OpenJDK Runtime Environment (build 17.0.11+0-17.0.11b1207.24-11852314)

[...]

PR #43, opened 2024-11-07
fix can`t clone flatbuffers

wangcker:
#38

### DIFF
--- a/importer/CMakeLists.txt
+++ b/importer/CMakeLists.txt
@@ -16,14 +16,15 @@ FetchContent_Declare(
     GIT_REPOSITORY https://github.com/google/flatbuffers.git
     GIT_TAG        129ef422e8a4e89d87a7216a865602673a6d0bf3
 )
-FetchContent_MakeAvailable(flatbuffers)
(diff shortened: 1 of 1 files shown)


importer/CMakeLists.txt
```diff
@@ -16,14 +16,15 @@ FetchContent_Declare(
     GIT_REPOSITORY https://github.com/google/flatbuffers.git
     GIT_TAG        129ef422e8a4e89d87a7216a865602673a6d0bf3
 )
-FetchContent_MakeAvailable(flatbuffers)
+#FetchContent_MakeAvailable(flatbuffers)
```
bdero, line 19:
```suggestion
```

bdero:
Hi, just to clarify: Does this patch fully resolve https://github.com/bdero/flutter_scene/issues/38 for you?

I was unable to reproduce the failure against the cmake 3.28.3 homebrew build, so I wonder if this was caused by a bug present in 3.28.1. Either way, this change is harmless, so we should merge it if it fixes the build issue for you.

PR #67, opened 2025-04-06
Fix the discrepancy between an object's world coordinates and its displayed position.

wangcker:
```
Matrix4 get globalTransform {
    final parent = _parent;
    if (parent == null) {
      return localTransform;
    }
    return  localTransform * parent.globalTransform;
  }
```

```
void render(SceneEncoder encoder, Matrix4 parentWorldTransform) {
    if (!visible) {
      return;
    }

    if (_animationPlayer != null) {
      _animationPlayer!.update();
    }

    final worldTransform = parentWorldTransform * localTransform;
    if (mesh != null) {
      mesh!.render(
        encoder,
        worldTransform,
        _skin?.getJointsTexture(),
        _skin?.getTextureWidth() ?? 0,
      );
    }
    for (var child in children) {
      child.render(encoder, worldTransform);
    }
  }
```

### DIFF
--- a/examples/flutter_app/pubspec.yaml
+++ b/examples/flutter_app/pubspec.yaml
@@ -1,11 +1,11 @@
 name: example_app
 description: "flutter_scene example project."
-publish_to: 'none'
+publish_to: "none"
 
 version: 1.0.0+1
 
 environment:
-  sdk: '>=3.4.0-82.0.dev <4.0.0'
+  sdk: ">=3.4.0-82.0.dev <4.0.0"
 
 dependencies:
   flutter:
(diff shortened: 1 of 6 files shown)


bdero:
Hey @wangcker, could you please file a bug describing the issue you're trying to fix here? Or, please push a diff containing only the relevant changes so that they're easier to identify.

wangcker:
fix bug: node`s worldTransform get a error Translation. (node.globalTransform=localTransform * parent.globalTransform. but in rendder  node.worldTransform = parentWorldTransform * localTransform )
btw the  branch was can`s run that commit extra file let it running correct.

bdero:
Ah I see. I pushed a fix with tests here: https://github.com/bdero/flutter_scene/commit/962086b0841d34960e484a4b8b78dd3b97284d00

The fix is published to `0.9.2-0`!

Issue #82, opened 2025-09-13
STDERR: Unknown experiment: native-assets


cruvie:
```sh
cruvie@192 flutter_app % flutter run -d macos --enable-flutter-gpu
Launching lib/main.dart on macOS in debug mode...
Unhandled exception:
Exception: Failed to run flutter_scene_importer:import command in build hook (exit code 254):
STDERR: Unknown experiment: native-assets

STDOUT: 
#0      buildModels (package:flutter_scene_importer/build_hooks.dart:46:7)
#1      main.<anonymous closure> (file:///Users/cruvie/Documents/open-source/flutter_scene/examples/flutter_app/hook/build.dart:6:5)
#2      build (package:native_assets_cli/src/api/build.dart:107:16)
#3      main (file:///Users/cruvie/Documents/open-source/flutter_scene/examples/flutter_app/hook/build.dart:5:3)
#4      _delayEntrypointInvocation.<anonymous closure> (dart:isolate-patch/isolate_patch.dart:312:33)
#5      _RawReceivePort._handleMessage (dart:isolate-patch/isolate_patch.dart:193:12)

  Building assets for package:example_app failed.
  build.dart returned with exit code: 255.
  To reproduce run:
  /Users/cruvie/Documents/flutter/bin/cache/dart-sdk/bin/dart --packages=/Users/cruvie/Documents/open-source/flutter_scene/examples/flutter_app/.dart_tool/package_config.json /Users/cruvie/Documents/open-source/flutter_scene/examples/flutter_app/.dart_tool/hooks_runner/example_app/7ad245dadc068daf8ba9d7bb51a8ef8e/hook.dill --config=/Users/cruvie/Documents/open-source/flutter_scene/examples/flutter_app/.dart_tool/hooks_runner/example_app/7ad245dadc068daf8ba9d7bb51a8ef8e/input.json
  stderr:
  Unhandled exception:
Exception: Failed to run flutter_scene_importer:import command in build hook (exit code 254):
STDERR: Unknown experiment: native-assets

STDOUT: 
#0      buildModels (package:flutter_scene_importer/build_hooks.dart:46:7)
#1      main.<anonymous closure> (file:///Users/cruvie/Documents/open-source/flutter_scene/examples/flutter_app/hook/build.dart:6:5)
#2      build (package:native_assets_cli/src/api/build.dart:107:16)
[...]

PR #84, opened 2025-10-20
refactor: remove native_assets_cli dependency and add hooks package

M3BIONIX:
Updated the package to use hooks as it was depricated to use hooks which replaced native-assets

### DIFF
--- a/importer/hook/build.dart
+++ b/importer/hook/build.dart
@@ -1,7 +1,6 @@
 import 'dart:io';
-
 import 'package:logging/logging.dart';
-import 'package:native_assets_cli/native_assets_cli.dart';
+import 'package:hooks/hooks.dart';
 
 void main(List<String> args) async {
   await build(args, (config, output) async {
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -26,13 +26,14 @@ dependencies:
   #flutter_scene_importer: ^0.9.0-0
   flutter_scene_importer:
     path: ./importer
+  hooks: ^0.20.4
   native_assets_cli: '>=0.13.0 <0.14.0'
   vector_math: ^2.1.4
 
 dev_dependencies:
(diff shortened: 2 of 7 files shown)


M3BIONIX:
@bdero Could you please review this ?

pubspec.yaml
```diff
@@ -26,13 +26,14 @@ dependencies:
   #flutter_scene_importer: ^0.9.0-0
   flutter_scene_importer:
     path: ./importer
+  hooks: ^0.20.4
   native_assets_cli: '>=0.13.0 <0.14.0'
```
jbbjarnason, line 30: probably remove this line

importer/hook/build.dart
```diff
@@ -1,7 +1,6 @@
 import 'dart:io';
-
 import 'package:logging/logging.dart';
-import 'package:native_assets_cli/native_assets_cli.dart';
+import 'package:hooks/hooks.dart';
```
jbbjarnason, line 3: similar change needed in flutter_app/hook/build.dart

M3BIONIX: Tried this, but as flutter_gpu_shaders still uses native assets inside the package, we get a type error for BuildInput and BuildOutput

M3BIONIX: Added a PR for the same https://github.com/bdero/flutter_gpu_shaders/pull/11, once this gets merged will install the latest shaders and update it here

bdero:
Thanks for filing this @M3BIONIX! The same migration just landed in #91 (which had to be re-implemented after the pub-workspaces refactor in #90 reorganized the layout). Your diff was a useful reference for confirming the right shape. Closing as superseded — issue #82 is now also closed.

M3BIONIX:
Thanks @bdero Required this one for my project

PR #121, opened 2026-05-15
Add a Stress Tests menu that runtime-downloads Khronos sample assets

bdero:
Adds a Stress Tests entry to the example app's dropdown that opens a catalog of 11 Khronos glTF Sample Assets. Tapping a scene streams the GLB from raw.githubusercontent.com into the app's support directory (cached after first download) and renders it via `Node.fromGlbBytes`. No build hook, no `.model` file, no asset registration.

Three classes of coverage we don't otherwise exercise:

- PBR fidelity / scale: A Beautiful Game (chess set, 42 MB, many materials/textures), Antique Camera, Lantern, Water Bottle, Metal/Rough Spheres
- Skinning + animation: Recursive Skeletons (84 skins, deeply nested), Brain Stem, Fox
- Correctness suite: Alpha Blend Mode Test, Normal/Tangent Test, Orientation Test

The scene screen ships an FPS-style camera (WASD strafe, Q/E elevate, click-drag to look, Shift for 4x boost). Stress models vary hugely in scale, so a fixed orbit camera made it hard to tell whether something was offscreen or actually broken. Movement speed auto-scales to the model's AABB radius so the same controls feel right for a 10cm bottle and a 10m chess set.

Sponza and FlightHelmet are absent. Both ship only as multi-file glTF (separate `.bin` + textures) on the Khronos repo, and the runtime importer is GLB-only. Adding multi-file glTF support to the runtime importer would unlock them later.

## Notes

- Adds `package:http` + `package:path_provider` to the example app.
- macOS requires `com.apple.security.network.client` in the example's debug + release entitlements. That lives under the gitignored `macos/` scaffolding, so it's a one-time edit per checkout (not part of this commit).

## Known follow-ups

Two separate runtime-importer bugs surfaced while testing the skinned scenes:

- Skinned PBR materials render flat gray. The fragment shader's `FragInfo` uniform (base color, lighting) doesn't reach the shader when the vertex stage is `SkinnedVertex`.
- Animation channels bind to targets by node name. glTF nodes are often unnamed, so every channel collapses onto the same first nameless node and skin animation doesn't play.

## Test plan

- [x] Verified on macOS Impeller: catalog renders, all 11 GLBs download + parse.
- [ ] Verify on Pixel 10 Vulkan + Impeller GLES.
- [ ] Verify on iOS Metal.

### DIFF
--- a/examples/flutter_app/lib/example_stress_tests.dart
+++ b/examples/flutter_app/lib/example_stress_tests.dart
@@ -1,0 +1,651 @@
+// Stress-test catalog: downloads Khronos glTF Sample Assets at runtime
+// and renders them via the runtime GLB importer. Lets the renderer be
+// exercised against PBR fidelity, animation/skinning, and correctness
+// scenes without committing big binary blobs to the repo.
+
+import 'dart:async';
+import 'dart:io';
+import 'dart:math';
+
+import 'package:flutter/material.dart' hide Animation;
+import 'package:flutter/services.dart';
+import 'package:flutter_scene/scene.dart' hide Material;
+import 'package:http/http.dart' as http;
+import 'package:path_provider/path_provider.dart';
+import 'package:vector_math/vector_math.dart' as vm;
+
+class _StressTest {
+  const _StressTest({
+    required this.id,
+    required this.title,
+    required this.description,
+    required this.url,
+    required this.sizeBytes,
+    this.animationName,
+  });
+
+  final String id;
+  final String title;
+  final String description;
+  final String url;
+  final int sizeBytes;
+  // If set, the first matching animation is played on a loop after load.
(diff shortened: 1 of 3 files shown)


aki1770-del:
Saw #121 yesterday while we were shipping `glb_stream` v0.1.0 (pub.dev) — a pure-Dart streaming parser for the GLB container, sitting upstream of `Node.fromGlbBytes`. For the Khronos sample sizes in your Stress Tests menu the existing bytes-buffered path is the right answer; the streaming complement matters at the tail of the size distribution (large crash-scan / cabin / IVI scenes that don't fit in memory all at once). Mentioning it for awareness — happy to share specifics if useful.

bdero:
Thanks! Will consider it.

PR #129, opened 2026-05-17
Add Node.fromGlbStream factory for Stream<List<int>> input

aki1770-del:
Sibling to the existing `Node.fromGlbBytes` / `Node.fromGlbAsset` factories, accepting a `Stream<List<int>>` source instead of a materialised `Uint8List` or asset path.

Use cases this opens up:
- HTTP response bodies via `package:http` (`response.stream`)
- `dart:io` `File.openRead()` pipes
- Websocket frame sources
- Any caller with a chunked byte source but no full buffer up-front

Implementation drains the stream into a `BytesBuilder` and delegates to `fromGlbBytes` — peak memory equals the full GLB size, matching `fromGlbBytes` semantics. True incremental parsing of the GLB container is intentionally out of scope for this factory.

The factory accepts `Stream<List<int>>` (Dart-idiomatic; matches `dart:io` and `package:http` conventions); `Stream<Uint8List>` callers work transparently since `Uint8List` implements `List<int>`.

**On the test approach:** existing runtime-importer tests like `runtime_importer_byte_comparison_test.dart` require `flutter_gpu`. The new test guards the stream-collection code path directly (without `flutter_gpu`) so it remains runnable in pure-`dart:test` CI. Happy to migrate to a `flutter_gpu`-enabled test instead if that fits the test suite layout better.

*AI-assisted — authored with Claude, reviewed by Komada.*

### DIFF
--- a/packages/flutter_scene/lib/src/node.dart
+++ b/packages/flutter_scene/lib/src/node.dart
@@ -300,6 +300,35 @@ base class Node implements SceneGraph {
     return importGlb(bytes);
   }
 
+  /// Load a single-file `.glb` model from a [Stream] of byte chunks.
+  ///
+  /// Convenience wrapper for [fromGlbBytes] that drains the stream
+  /// into a single buffer before parsing. Useful when the caller has
+  /// a `Stream<List<int>>` (e.g. an `http` response body, a `dart:io`
+  /// `File.openRead()` pipe, or a websocket frame source) but not the
+  /// full byte buffer up-front.
+  ///
+  /// This factory buffers the entire stream in memory before
+  /// delegating to [fromGlbBytes] — peak memory equals the full GLB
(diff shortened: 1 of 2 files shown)


bdero:
Thanks for the contribution, and sorry for the slow response. I'm going to pass on this one. The loading surface on `Node` deliberately stays source-agnostic (the engine takes bytes, producing them is the app's job), and for a chunked source `package:async` already covers it in one line:

```dart
final node = await Node.fromGlbBytes(await collectBytes(response.stream));
```

I also want to keep the `fromGlbStream` name unclaimed for a possible real streaming loader that populates the scene progressively as data arrives, rather than shipping it now with buffer-everything semantics. If you have a concrete case the buffered path falls short on, an issue describing it would be genuinely useful. Thanks again for the clean PR.

PR #216, opened 2026-07-04
Add scene loading gate, pipeline warm-up, and off-thread glTF import

bdero:
Scenes can now hold rendering until their content is loaded, so they no longer flash half-built (objects at the origin, one group drawn before another) while assets stream in. This adds an opt-in loading gate, an offscreen pipeline warm-up, and off-UI-thread geometry packing.

`ResourceGroup` tracks a set of in-flight loads, reports aggregate progress as a `ValueListenable<double>`, and completes when they have all settled (recording any errors without aborting).

`SceneView` gains `loading`, `loadingBuilder`, and `revealMinDuration`. When either `loading` or `loadingBuilder` is set, the view holds the scene off-screen behind the loading widget until the engine's shared static resources and the tracked loads are ready, then reveals the fully assembled scene in one frame. While gated, per-frame ticking is paused and the first tick after reveal carries a single-frame delta. `revealMinDuration` keeps a fast load from flashing the loading widget. A view with no loading arguments renders exactly as before.

`Scene.warmUp` compiles the scene's render pipelines and uploads its GPU resources ahead of the first visible frame by encoding one frame offscreen and discarding it, so the first frame does not stall while shaders compile. It renders through the real view path at a tiny size, since a pipeline's identity does not depend on resolution, so it compiles exactly the variants the first frame uses. `SceneView` gains a `warmUp` flag that runs it once the loading group is ready, before reveal; it gates the view on its own.

The runtime glTF/GLB importer now packs mesh geometry on a background isolate, so loading a large model no longer stalls the UI thread with that work. The document is parsed and its buffers resolved on the calling thread, every primitive is packed on the isolate, and the packed buffers are uploaded to the GPU on the raster thread as before. The shared packer is unchanged, so imported geometry is byte-for-byte identical. On the web, where the isolate helper runs inline, this is transparent.

`Scene.isReadyToRender` exposes the shared static-resource state.

All of it is additive and non-breaking. 771 tests pass, and the changes are exercised end to end by a downstream app on Metal (loading screen, warm-up, and the off-thread import of a skinned, animated model). The render-path pieces still want a run across the full smoke matrix, which CI covers.

### DIFF
--- a/packages/flutter_scene/lib/scene.dart
+++ b/packages/flutter_scene/lib/scene.dart
@@ -172,6 +172,7 @@ export 'src/runtime_importer/gltf_resources.dart' show GltfResourceResolver;
 export 'src/scene_path.dart'
     show BezierPath, CatmullRomPath, PolylinePath, ScenePath, ScenePathFrame;
 export 'src/raycast.dart' show SceneRaycastHit, raycastNode, raycastNodeAll;
+export 'src/resource_group.dart' show ResourceGroup;
 export 'src/scene_pointer.dart' show ScenePointer;
 export 'src/scene.dart' show AntiAliasingMode, Scene, SceneGraph;
 export 'src/widget_texture.dart'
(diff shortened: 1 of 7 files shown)


bdero:
Docs follow-up for this API: bdero/fscene.dev#1 adds loading-screen and warm-up guidance to the Assets and loading guide. Its API deep links 404 until a release including these symbols is published, so it should merge/deploy after this lands and ships.

PR #219, opened 2026-07-05
Rotate per-frame transient host buffers on GPU completion

bdero:
HostBuffer recycles its internal storage a fixed number of resets after data is written. When frames queue deep on the GPU (heavy scenes), the renderer's per-frame transient buffer could be reset and rewritten while the GPU was still reading it, which showed up as flickering and corrupted draws under load.

This adds a GpuSubmissionTracker that records every command buffer the renderer submits and reports the id through which all submissions have completed, and a TransientsPool that hands out a host buffer only when the submissions that referenced it have completed (or a fresh buffer when none has). Buffer memory grows with actual GPU queue depth and shrinks back to a single spare when the queue drains. All render passes now allocate their per-frame transients through the pool.

Verified on a heavy voxel scene that corrupted reliably before the change and renders clean after it. Unit tests cover the tracker's out-of-order completion accounting and the pool's reuse, growth, and shrink behavior.


bdero:
Already landed on master as 9f2b6f9; this branch was a stale duplicate.

PR #256, opened 2026-07-22
Add a declarative scene widget API, KHR_materials_variants support, and a product configurator example

bdero:
Scenes can now be described in `build()` with widgets that own and reconcile retained scene-graph nodes, as an optional alternative to the imperative Scene/Node API (which is unchanged). Rebuilds apply only changed properties, structure reconciles through the element tree, and hot reload of scene structure works in place.

`SceneView.declarative` renders a view-owned `Scene` configured through constructor props. `SceneNode` declares a transform, components, and children; `SceneMesh` adds identity-diffed geometry and material; `SceneModel` loads a `.glb` asynchronously with placeholder/error subtrees, named-variant selection, and declarative animation playback via `SceneAnimationSpec` (playing/loop/weight/speed diffed onto clips as plain property writes). Equal-keyed models load once and share a refcounted template; each widget mounts its own clone, keeping GPU resources shared. `SceneSubtree`, `SceneNodeHost`, and `SceneNodeController` bridge the declarative and imperative styles in both directions.

`KHR_materials_variants` is supported by both import paths. Imported models carry a `MaterialsVariantsComponent` whose `select(name)` swaps mapped materials in place, and the `.fscene` format serializes variants as a component with no container change, so pre-converted assets keep them. Fixing live swaps surfaced a render-item gap (materials were captured at registration), now re-registered on selection with a regression test.

The new Configurator example live-downloads the Khronos MaterialsVariantsShoe (© Shopify, CC BY 4.0) into a showroom scene with colorway-reactive lighting, and the Animation example now blends dash.glb's clips declaratively. The declarative layer, model caching, animations, and format round-trip are covered by GPU-free tests via an injectable importer seam.

### DIFF
--- a/examples/flutter_app/lib/example_animation.dart
+++ b/examples/flutter_app/lib/example_animation.dart
@@ -9,6 +9,15 @@ import 'example_panel.dart';
 import 'example_settings.dart';
 import 'lighting_panel.dart';
 
+/// Skeletal animation blending, driven through the declarative API.
+///
+/// The model and its animation state are declared in `build()`: a
+/// [SceneModel] mounts dash.glb as a child of the app-owned [Scene] (the
+/// mixed mode, so the shared lighting panel keeps mutating the scene
+/// imperatively), and each blend slider just rebuilds with new
+/// [SceneAnimationSpec] weights. The widget diffs the specs onto the
+/// underlying clips as plain property writes, so dragging a slider costs a
+/// field write per frame and the blending itself runs engine-side.
 class ExampleAnimation extends StatefulWidget {
   const ExampleAnimation({super.key});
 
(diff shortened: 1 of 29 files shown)


kiddo4:
I reviewed #256 closely. Overall, I really like the architecture. The retained-node reconciliation, shared model templates, animation specs, and bridges between declarative and imperative ownership feel like a strong base for Glint.

I found two lifecycle cases worth addressing before release:

If a SceneModel source changes while its first load is pending, the stale generation still completes the shared load gate in finally. That could reveal or warm up the SceneView before the replacement model finishes. I think the gate should be generation-specific, or only settled by the current generation.

The model cache releases by cacheKey. After hot-reload eviction and reacquisition, an older holder can release the replacement entry because both generations share the same key. Returning an entry-specific lease from acquire would prevent stale holders from decrementing a newer entry.

A few API thoughts from the Glint side:
It would be useful to enforce SceneNodeController's single-owner contract instead of silently reassigning it.
A retry/reload surface for SceneModel would help with transient network sources.
Inline component construction can reset engine-side state on rebuild, so stronger guidance—or eventually a component widget/factory abstraction—could make that safer.

Beyond those points, the direction feels right. SceneSubtree and SceneNodeHost give us a clean interoperability boundary, and I can see a Glint-on-Scene prototype mapping naturally onto this API.

I went ahead and made a PR to fix the two lifecycle issues: #259 (branch `kiddo4/fix-declarative-model-lifecycles`), with regression tests for both. I'll turn the API thoughts into a follow-up separately, then start on the adapter prototype with you.

PR #259, opened 2026-07-23
Fix two SceneModel lifecycle races flagged in the #256 review

kiddo4:
Follow-up to the review on #256. Two lifecycle issues there, both fixed here:

**Stale generation completing the load gate early.** If a `SceneModel`'s source changes while its first load is still pending, the superseded load's `finally` block was completing the enclosing `SceneView`'s reveal/warm-up gate unconditionally. That could reveal (or warm up) the view before the replacement model actually finished loading. The gate logic is now pulled into a small `SceneModelLoadGate` helper that only settles for the generation that's still current; dispose still force-settles it unconditionally since no later generation will run.

**Cache release by key instead of by entry.** The model template cache released by `cacheKey`, so after a hot-reload evict + reacquire under the same key, an older holder's later release could decrement (and, at refcount zero, evict) the *new* entry instead of the one it actually held. `acquire` now returns a lease bound to the specific entry it incremented, and `release` only ever touches that entry.

Also added `debugEvictModelTemplateCache` (mirrors the existing `debugClearModelTemplateCache`) so the second issue is reproducible in a test without going through the full asset hot-reload path.

## Test plan
- New tests in `scene_model_test.dart`:
  - `SceneModelLoadGate` unit tests covering the generation-safety invariant directly (no GPU needed).
  - A widget test reproducing the evict+reacquire race, confirmed to fail against the old key-based release and pass with the fix.
- `flutter test` (master channel) — full suite green, 1007 passed / 26 skipped (GPU-only).
- `flutter analyze` clean on the changed files.

### DIFF
--- a/packages/flutter_scene/lib/src/widgets/declarative.dart
+++ b/packages/flutter_scene/lib/src/widgets/declarative.dart
@@ -43,6 +43,7 @@ import 'package:flutter_scene/src/hot_reload/hot_reload_coordinator.dart';
 import 'package:flutter_scene/src/material/material.dart';
 import 'package:flutter_scene/src/mesh.dart';
 import 'package:flutter_scene/src/node.dart';
+import 'package:flutter_scene/src/resource_group.dart';
 import 'package:flutter_scene/src/widgets/scene_view.dart';
 
 /// The declarative scene layer, widgets that own and reconcile [Node]s in a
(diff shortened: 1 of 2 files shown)


bdero (approved):
Verified both fixes against master, regression test fails on the old by-key release, full suite green. Pushed 645870f, a formatting pass with the stable dart_style CI uses.

One follow-up if you want it, the lease is never released on import failure. Harmless today (the failure path evicts the entry) but worth tightening.

LGTM, thanks for the careful work!

kiddo4:
Thanks for catching that and for the formatting pass. Went ahead and fixed the lease leak too: #268.

PR #268, opened 2026-07-24
Release the SceneModel template lease on import failure

kiddo4:
Follow-up from your review on #259 — the catch branch in `SceneModel`'s `_load` never released the lease it acquired, so a failed import left that entry's refcount stuck incremented.

You're right that it's harmless today, since `_import`'s own catch already evicts the entry regardless of refcount. But that's relying on the eviction-on-any-failure behavior as an implicit guarantee rather than having the lease accounting be correct on its own, so tightened it up: the catch branch now releases the lease unconditionally, same as the mismatched-generation branch in the try already does.

## Test plan
- `flutter analyze` clean.
- `flutter test` — full suite green (1024 passed). No new test added for this specific one since there's currently no path where it's observable from the outside (as you noted) — happy to add one if you'd rather have the invariant pinned down explicitly.

### DIFF
--- a/packages/flutter_scene/lib/src/widgets/declarative.dart
+++ b/packages/flutter_scene/lib/src/widgets/declarative.dart
@@ -1046,6 +1046,13 @@ class _SceneModelState extends State<SceneModel>
       _registerHotReload(source, modelRoot);
       widget.onLoaded?.call(modelRoot);
     } catch (e) {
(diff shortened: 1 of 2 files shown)


bdero (approved):
The unconditional release double-released when the try threw after the lease handoff (dispose releases the held lease too). Pushed 141ad3f with the guard and a regression test. LGTM with that, thanks!

PR #275, opened 2026-07-27
Handle quaternion double cover in slerp

aenriqu:
`QuaternionSlerp.slerp` doesn't account for quaternion double cover. When `dot(to)` is negative the two inputs sit in opposite hemispheres, and `atan2(sine, cosine)` then returns an angle over 90 degrees, so the interpolation sweeps the long way round.

This shows up in skeletal animation. glTF exporters are free to emit either `q` or `-q` for a keyframe, and `animation_builder.dart` passes accessor values through unchanged, so sign-flipped pairs reach the resolver as-is. A pair 21 degrees apart then interpolates as a 339 degree rotation over a single frame, which reads as a limb snapping backwards through the body.

`dash.glb` mostly avoids this. `Idle` and `Run` contain no sign flips at all, and the only ones in `Jump`/`JumpStart` are on `ToesPivot` and `KneeIK` at 1-2 degrees, where the artifact isn't visible. With a character pack from KayKit for example, I see 128 of 173 clips affected, mostly on `upperleg` during walk and run cycles.

The `.abs()` has to go too, since it was hiding the sign this needs to test. Widening the epsilon isn't an alternative: near-antipodal inputs would then take the lerp branch and sum to roughly the zero quaternion before normalising.

Same approach as three.js `Quaternion.slerp`, which negates one of the inputs when the dot product is negative.

### DIFF
--- a/packages/flutter_scene/CHANGELOG.md
+++ b/packages/flutter_scene/CHANGELOG.md
@@ -6,6 +6,7 @@
 * Declarative animation control via `SceneModel.animations` (`SceneAnimationSpec`), plus cached model templates so equal sources load once and share GPU resources.
 * `KHR_materials_variants` support in both import paths via `MaterialsVariantsComponent`, declaratively via `SceneModel.variant`, with a new Configurator example.
 * Fixed animation blending collapsing rigs with mirrored bones.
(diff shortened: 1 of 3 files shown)


bdero:
Thanks!

PR #277, opened 2026-07-29
Invalidate the lighting bind memo in the instanced draw path

Groseuros:
`EngineLightingUniforms` memoises its bind set and documents that "the scene encoder invalidates
the memo whenever it clears the pass's bindings". `_encode` does that (`scene_encoder.dart:379-380`).
`_encodeInstanced` clears at `:441` without invalidating, so from the second instanced draw onward
the engine lighting samplers are wiped and never rebound — those draws, and any non-instanced mesh
encoded after them, shade black while still writing depth.

```dart
_renderPass.clearBindings();
EngineLightingUniforms.invalidateBindMemo();   // added
```

Needs two or more `InstancedMesh`es sharing a pipeline. With one, the clear hits an empty pass and
the memo misses anyway, so everything binds — probably why it hasn't come up.

**Repro:** cuboid instances split across K `InstancedMesh`es with `PhysicallyBasedMaterial`, plus a
ground-plane `Node` and one `DirectionalLight`. Correct at K=1; at K≥2 only one type renders and the
ground disappears.

**Only visible on the GLES backend.** Windows/Vulkan, an Android flagship on Vulkan, and web/WebGL2
all render correctly *without* the patch, so they are tolerating the missing binds rather than
unaffected — which is likely why this survived, since Android only falls back to GLES where Vulkan
is unavailable. On GLES it also under-reports frame cost, because the unshaded geometry still writes
depth: the same scene measured 3.8 ms raster before the fix and 4.9 ms after.

No output or timing change on Vulkan or WebGL2 after the patch. Cost is ~12 binds per instanced mesh
per frame, scaling with mesh count rather than instance count.

Happy to instead mirror `_encode` exactly and only clear on an actual pipeline change, if you'd
prefer that over the minimal fix.

flutter_scene 0.20.0 and current master (`70125a88`); Flutter master `bc2162e270`.

### DIFF
--- a/examples/smoke_render/lib/smoke_scenes.dart
+++ b/examples/smoke_render/lib/smoke_scenes.dart
@@ -211,6 +211,57 @@ final List<SmokeScene> kSmokeScenes = <SmokeScene>[
     scene.add(node);
     return (scene: scene, camera: _camera());
   }),
+  // Two instanced meshes plus a plain mesh, all sharing a lit pipeline.
+  // Covers hardware instancing and, because the draws share a pipeline, the
+  // engine-lighting bind bookkeeping between instanced and non-instanced
+  // draws; a lost bind reads as black or missing geometry here.
+  SmokeScene('instanced_lighting', () {
+    final scene = Scene();
+    scene.add(
+      Node()..addComponent(
+        DirectionalLightComponent(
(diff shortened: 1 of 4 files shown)


bdero (approved):

bdero:
Thanks for this!

PR #280, opened 2026-07-31
Improve instanced rendering and directional shadow performance

bdero:
Adds retained geometry and instance-data paths, batch culling, and transient-buffer reuse to reduce per-frame CPU work and allocation churn. Adds per-node shadow casting, continuously changing directional-shadow controls, stable filtering, tighter cascade fitting, and fixes for cascade-edge striping. Migrates generated build outputs to managed DataAssets and switches material/display conversion to exact sRGB transfer functions.


packages/flutter_scene/lib/src/render/depth_prepass.dart
```diff
@@ -372,6 +434,9 @@ class _DepthPrepassEncoder {
         item.worldTransform,
         instances,
         nodeWindingFlipped: item.windingFlipped,
+        instanceWindingFlipped: item.instanceWindingFlipped,
+        indices: item.visibleInstanceIndices,
+        scratch: transientInstancePackingScratch,
```
bdero: The instance data bound here no longer matches the pipeline layout.

At line 335 the pipeline resolves with `depthVertex?.layout ?? geometry.instancedVertexLayout`. When `depthVertex` is null (`_writeNormals || masked`) that is `kUnskinnedInstancedLayout` or `kUnskinnedSoAColorLayout`, both of which this PR widened to an 80-byte instance stride with `instance_color` at offset 64. But `packInstanceTransforms` emits 64 bytes per instance, so instance `i` fetches `model_transform_*` at byte `i * 80` while the data sits at `i * 64`. Everything past the first instance gets a garbage transform and the fetch runs past the end of the buffer.

`ShadowEncoder` handles this correctly, it picks `packInstanceData` vs `packInstanceTransforms` on `depthVertex == null` and threads the choice through `_drawPacked`. The same split is needed here.

`bindSingleInstanceTransform` at line 460 has the same problem in the non-instanced case, 64 bytes into an 80-stride slot.

The new `flush()` batching widens the blast radius well beyond `InstancedMesh`, since it now merges N plain meshes sharing geometry and material into one synthetic instanced item. DoF alone keeps `writeNormals` false, which is why the new `depth_post` smoke scene does not catch this.

packages/flutter_scene/lib/src/render/depth_prepass.dart
```diff
@@ -218,30 +229,76 @@ class _DepthPrepassEncoder {
...
+          first.jointsTexture == null) {
+        while (end < _records.length &&
+            identical(first.geometry, _records[end].geometry) &&
+            identical(first.material, _records[end].material) &&
+            _records[end].jointsTexture == null) {
+          end++;
+        }
       }
+      if (end > index + 1) {
+        final transforms = <Matrix4>[];
```
bdero: This is the one place the PR adds per-frame allocation. Per batch it builds a growable `List<Matrix4>`, then a `Matrix4.copy` per non-instanced item and a fresh `Matrix4` per instance from `item.worldTransform * instances[i]`, then a throwaway `RenderItem` (which itself allocates `_visibleInstanceScratch`).

`ShadowEncoder.flush` already solves the identical problem allocation-free with `InstanceDataBatch` plus `packInstanceTransformBatches`. Reusing that here would fix both this and the stride mismatch below in one move.

packages/flutter_scene/lib/src/scene_encoder.dart
```diff
@@ -194,21 +261,29 @@ base class SceneEncoder {
   late final double? _lodFovRadiansY;
   final List<_OpaqueRecord> _opaqueRecords = [];
   final List<_TranslucentRecord> _translucentRecords = [];
+  static final List<_OpaqueRecord> _opaqueRecordPool = [];
```
bdero, line 304: These pools are static and records are returned without clearing their fields, so up to 8192 records keep strong references to a `RenderItem`, `Geometry`, `Material`, `RenderPipeline`, `Matrix4`, and the joints `gpu.Texture`. After a `Scene` is torn down that pins its geometry, materials, and GPU textures for the process lifetime.

Either null the reference fields on release, or make the pools per-encoder so they die with the frame.

packages/flutter_scene/lib/src/scene_encoder.dart
```diff
@@ -281,31 +349,37 @@ base class SceneEncoder {
...
+      final bounds = item.worldBounds;
+      _translucentRecords.add(
+        _obtainTranslucentRecord(
+          item,
+          item.worldTransform,
+          geometry,
+          material,
+          fade,
+          pipeline,
+          bounds == null
```
bdero, line 405: This changes how translucent instanced meshes composite. Previously each instance became its own translucent record and sorted independently against other translucent geometry; now the whole batch sorts as one unit at its aggregate-bounds center. Instances still sort back to front within the batch, but a translucent instanced mesh interleaved with other translucent objects will render differently than before.

Reasonable tradeoff for the draw-call win, but it needs a CHANGELOG line since it is a visible behavior change.

packages/flutter_scene/lib/src/render/instance_packing.dart
```diff
@@ -14,16 +58,371 @@ import 'package:vector_math/vector_math.dart';
...
+  final cw = scratch?.cw(cwLength) ?? Float32List(cwLength);
+  final world = scratch?.world ?? Matrix4.zero();
+  var ccwIndex = 0, cwIndex = 0;
+  flatIndex = 0;
+  for (final batch in batches) {
+    final packedData = batch.packedWorldData;
+    if (packedData != null) {
+      final indices = batch.indices;
+      if (indices == null) {
+        final winding = _uniformWinding(batch.packedWindingFlipped!);
```
bdero: `_uniformWinding` rescans the whole `packedWindingFlipped` array, but the first pass over the batches already computed exactly this. A batch is uniform iff its contribution to `cwCount` is either 0 or `batch.length`, so recording the per-batch delta in pass one removes this O(n) scan from the hot path.

packages/flutter_scene/lib/src/render/instance_packing.dart
```diff
@@ -14,16 +58,371 @@ import 'package:vector_math/vector_math.dart';
...
+/// One retained instance group consumed by [packInstanceDataBatches].
+class InstanceDataBatch {
+  InstanceDataBatch({
+    required this.nodeTransform,
+    required List<Matrix4> instances,
+    required List<Vector4> colors,
+    required this.nodeWindingFlipped,
+    List<bool>? instanceWindingFlipped,
+    this.indices,
+  }) : instances = instances,
```
bdero, line 112: `instances = instances`, `colors = colors`, and `instanceWindingFlipped = instanceWindingFlipped` are redundant. These can be `required this.instances` and friends; the asserts below can still reference the parameter names.

packages/flutter_scene/CHANGELOG.md
```diff
@@ -7,7 +7,20 @@
 * Debug builds check a custom material's uniform blocks against the shader's reflection, so a misspelled or optimized-out block name and a buffer too small for its std140 layout raise named exceptions instead of failing at bind time or reading past the end.
 * Debug builds check a caller-declared vertex layout against the buffers actually bound, so a slot count mismatch raises instead of feeding the pipeline undefined vertex data.
 * `IndexType`, `VertexFormat`, and `VertexStepMode` are exported, so 32-bit indices and caller-declared layouts no longer need `lib/src` imports.
-
+* Made first-party shader/material hooks and examples DataAssets-only, with migration guidance for legacy assets.
```
bdero, line 10: Retitle this section `## 0.21.0` and bump `packages/flutter_scene/pubspec.yaml` to match. There are at least three breaking changes in here.

- The instance-rate vertex buffer stride went 64 to 80 with a new required `instance_color` attribute, so any hand-written `ShaderMaterial` vertex shader using the instanced layout has to add `in vec4 instance_color`. This PR patches `example_ripple.vert` and `raw_pair.vert` for exactly that reason.
- DataAssets are now required for the first-party hooks, the `build/shaderbundles/*` asset keys are gone, and `flutter config --enable-dart-data-assets` becomes mandatory.
- Output color changes everywhere with the exact sRGB transfer.

A minor bump also keeps the `flutter_scene_rapier` lockstep release honest, since its `^0.20.0` caret would otherwise silently pick this up.

packages/flutter_scene/CHANGELOG.md
```diff
@@ -7,7 +7,20 @@
...
+* Added `InstancedMesh.sortTransparentInstances` to skip unnecessary particle sorting.
+* Added `DirectionalLight.cacheStaticShadows` for continuously changing shadow lights.
+* Added `DirectionalShadowFilter.fixedPcf` for stable lower-ALU shadow filtering.
+* Tightened stable cascade fitting to increase effective shadow resolution.
+* Reduced shadow-map color bandwidth with single-channel fp32 tiles.
+* Fixed cascade-edge striping from normal-biased receiver positions.
+* Cached instanced bounds and render data to keep moving-camera culling allocation-light.
+* Skipped per-instance bounds packing when instance culling is disabled.
+* Skipped per-instance culling when an entire batch is inside every active plane.
+* Reused large transient buffers across changing instance counts.
```
bdero, line 28: Several public additions from this PR are missing here.

- Per-instance colors (`InstancedMesh.addInstance(color:)`, `setInstanceColor`) and, more importantly, the `instance_color` vertex attribute plus the 64 to 80 byte instance stride, which is the breaking half
- `InstancedMesh.cullInstances`
- `RenderView.cullingPlanes`
- `Scene.warmUp(includeOffscreen:)`
- `SunLight.cacheStaticShadows` and `SunLight.shadowFilter`
- `GeometryBuilder.build({bufferArena, retainCpuData})`
- The switch to exact sRGB transfer functions, which shifts output color in every material and the resolve pass
- Translucent instanced batches now sorting as one unit

packages/flutter_scene/lib/src/fmat/build_materials.dart
```diff
@@ -162,6 +162,20 @@ Future<void> buildMaterials({
   if (materialPaths.isEmpty) {
     return;
   }
+  if (assetMode == MaterialAssetMode.legacyOnly) {
+    stderr.writeln(
+      'flutter_scene warning, buildMaterials is using legacy generated '
```
bdero, line 168: `assetMode` still defaults to `MaterialAssetMode.legacyOnly` at line 150, so this warns anyone who takes the default for taking the default. Either flip the default to `dataAssetsIfAvailable` in this release, or leave the default alone and drop the unconditional warning for it.

packages/flutter_scene/shaders/material_lighting.glsl
```diff
@@ -97,49 +95,50 @@ float SampleCascade(int cascade, int count, mat4 cascade_matrix, float box,
...
-  // A per-fragment rotation hides the 16-tap pattern as a smooth edge.
+  // Select the tap positions without duplicating the texture samples in both
+  // branches. Duplicating both kernels here expands to 33 samples per cascade
+  // in the generated GLES source even though the choice is uniform.
+  float fixed_filter = step(0.5, frag_info.directional_light_direction.w);
   float noise = fract(
       52.9829189 *
       fract(dot(gl_FragCoord.xy, vec2(0.06711056, 0.00583715))));
-  float angle = noise * 6.28318530718;
+  float angle = noise * 6.28318530718 * (1.0 - fixed_filter);
```
bdero, line 105: The default `rotatedPoisson` path now pays for the branch it does not use, a `mix()` per tap across 17 taps on every shadowed fragment. The `fixedPcf` path still evaluates `cos`/`sin` on a forced-zero angle too.

Small either way, and the comment explaining why this is uniform-but-not-branched (33 samples in the generated GLES source) is worth keeping. Just flagging that the cost lands on the default.

packages/flutter_scene/lib/src/render/shadow_pass.dart
```diff
@@ -113,14 +113,11 @@ class ShadowPass extends RenderGraphPass {
...
     // That made the flat distant ground self-shadow in moire bands.
-    //
-    // TODO(bdero): Only the red channel is used. Add r32Float to Flutter
-    // GPU and use it here instead.
     final color = context.texturePool.acquire(
       TransientTextureDescriptor.color(
         width: atlasWidth,
         height: _tileResolution,
-        format: gpu.PixelFormat.r32g32b32a32Float,
+        format: gpu.PixelFormat.r32Float,
```
bdero, line 120: Single-channel float has narrower color-renderable support than RGBA32F on GLES and WebGL2 (`EXT_color_buffer_float`), and CI's software stacks expose a different extension set than real GPUs, so a device can pass while CI fails or the reverse.

Per `notes/ci/local_smoke_render_matrix.md` this wants a local Metal/GLES/Vulkan/WebGL2 pass before merge. Worth confirming in the PR body whether that ran.

packages/flutter_scene/lib/src/render/scene_pass.dart
```diff
@@ -108,6 +119,8 @@ class ScenePass extends RenderGraphPass {
   final bool _captureOpaqueColor;
   final bool _bindSceneDepth;
   final double _time;
+  late final List<Plane> _cullingPlanes;
```
bdero: `late` is redundant, this is assigned in the initializer list. Every neighbouring field is a plain `final`.

packages/flutter_scene/lib/src/node.dart
```diff
@@ -83,6 +83,13 @@ base class Node implements SceneGraph {
   /// camera-dependent displacement.
   bool shadowStatic = false;
 
+  /// Whether this node's meshes cast shadows. Defaults to `true`.
+  ///
+  /// This does not affect whether the meshes receive shadows. The value is not
+  /// inherited by children.
+  /// {@category Rendering}
```
bdero: A member-level `{@category Rendering}` on a `{@category Scene graph}` class splits this symbol into a different doc topic from the class it lives on, and the adjacent `shadowStatic` carries none. Drop it here, or apply the pattern to the neighbouring members too.

packages/flutter_scene/lib/src/render/render_graph.dart
```diff
@@ -1,6 +1,14 @@
 import 'package:flutter_scene/src/gpu/gpu.dart' as gpu;
 import 'package:flutter_scene/src/render/frame_transients.dart';
 
+const bool _profileRenderGraph = bool.fromEnvironment('FLUTTER_SCENE_PROFILE');
+
+class _PassProfile {
```
bdero: The `FLUTTER_SCENE_PROFILE` instrumentation is duplicated three times, here plus `scene_pass.dart` and `scene_encoder.dart`, each with its own near-identical 120-sample accumulate-and-`print` block over static mutable state. It ships in the published package and is undocumented.

Worth factoring the accumulator into one small helper and noting the flag in `notes/`. The static counters in `ScenePass`/`SceneEncoder` also conflate across multiple `RenderView`s, which is fine for a dev tool but deserves a one-line comment.

bdero (commented):
Read through the whole change. The engine work is strong, the cascade-fit derivation and the biased-position cascade selection fix are both well argued, and the new packing tests are the right shape. `flutter analyze` is clean and all 1097 unit tests pass locally.

One blocking item, the depth prepass still binds 64-byte instance records into the widened 80-byte pipeline layout, so every instance past the first fetches a garbage transform whenever SSR, custom normals, or an alpha-masked caster is active. `ShadowEncoder` got the matching fix and `DepthPrepass` did not. Details inline.

The rest is a mix of memory retention, release hygiene, and smoke coverage notes. Also worth widening the PR description, it currently covers the instanced and shadow work but says nothing about the DataAssets migration or the exact sRGB transfer switch, which are the two parts with real consumer-facing impact. Landing the DataAssets migration as its own PR would make both easier to review and to revert.

examples/smoke_render/lib/smoke_scenes.dart
```diff
@@ -273,13 +326,14 @@ final List<SmokeScene> kSmokeScenes = <SmokeScene>[
...
   SmokeScene('directional_shadow', () {
-    final scene = Scene();
+    final scene = _smokeScene(hasAnalyticLight: true);
     scene.add(
       Node()..addComponent(
         DirectionalLightComponent(
           DirectionalLight(
             direction: vm.Vector3(-0.4, -1.0, -0.35),
             castsShadow: true,
+            shadowFilter: DirectionalShadowFilter.fixedPcf,
```
bdero: `directional_shadow` is the only shadow smoke scene, so switching it to `fixedPcf` leaves the default `rotatedPoisson` path with no cross-backend coverage at all. Given the one-scene-per-area budget this is a fair place to spend it, but it should be the default that stays covered.

examples/smoke_render/integration_test/smoke_test.dart
```diff
@@ -66,9 +65,13 @@ void main() {
         ),
       );
 
-      // Let the post-ready repaint and GPU frames settle.
-      for (var i = 0; i < 20; i++) {
-        await tester.pump(const Duration(milliseconds: 50));
+      // Let the post-ready repaint and GPU frames settle. Android software
+      // renderers need fewer, longer frames to stay below emulator watchdogs.
+      final settleFrames =
```
bdero, line 72: Android now settles with a single 1000ms frame instead of 20. That will hide anything that only stabilizes after a few frames, which is most of what this PR touches (bind memoization, the static shadow cache, transient pool reuse), on the two backends where the change is largest.

At minimum leave a `TODO(smoke)` naming the emulator watchdog constraint and what it costs.

.github/workflows/smoke_render_jobs.yml
```diff
@@ -119,15 +122,21 @@ jobs:
...
     timeout-minutes: 45
+    env:
+      GRADLE_OPTS: -Dorg.gradle.daemon=false
     strategy:
       fail-fast: false
       matrix:
         include:
           - backend: gles
             expected_backend: opengles
+            lightweight_environment: false
```
bdero: Both matrix entries set `lightweight_environment: false` and `render_size: 512`, which are the existing defaults, so `SMOKE_USE_LIGHTWEIGHT_ENVIRONMENT`, `SMOKE_RENDER_SIZE`, and the whole `allowsDarkForeground` plus `_smokeScene()` mechanism in `smoke_scenes.dart` are inert as committed. Either wire one lane to actually use them or drop the machinery.

.github/workflows/smoke_render_jobs.yml
```diff
@@ -168,30 +177,49 @@ jobs:
             (adb devices -l > examples/smoke_render/build/android-diagnostics/adb_devices_before_${{ matrix.backend }}.txt 2>&1 || true) &&
             (adb shell getprop > examples/smoke_render/build/android-diagnostics/getprop_${{ matrix.backend }}.txt 2>&1 || true) &&
             (adb shell dumpsys SurfaceFlinger > examples/smoke_render/build/android-diagnostics/surfaceflinger_${{ matrix.backend }}.txt 2>&1 || true) &&
+            (host_monitor() { while true; do
```
bdero, line 178: The 2-second `host_monitor` polling loop, `sudo dmesg`, and the `memory.events`/`meminfo` dumps read as bisecting scaffolding for the Android hangs rather than permanent diagnostics. If they stay, a line saying what they are for would help; otherwise worth dropping with the rest of the debugging pass.

packages/flutter_scene/test/instance_packing_test.dart
```diff
@@ -59,5 +61,187 @@ void main() {
...
+
+      expect(packed.ccwCount, 1);
+      expect(packed.cwCount, 1);
+      expect(packed.ccw[12], 8);
+      expect(packed.ccw[16], 2);
+      expect(packed.cw[12], 4);
+      expect(packed.cw[16], 1);
+    });
+
+    test('copies homogeneous cached records contiguously', () {
```
bdero, line 212: The packing primitives are well covered here, but nothing tests that the encoders assemble the right batches. `_canBatchOpaque`, the run-detection loops in `SceneEncoder.flushOpaque`, `ShadowEncoder.flush`, and `DepthPrepass.flush`, and the cached-vs-uncached `InstanceDataBatch` selection are the riskiest new code in the PR and are untested.

A test asserting that N single meshes sharing geometry and material collapse into one batch, and that a differing `lightListOffset`, `fade`, or `jointsTexture` splits them, would have caught the depth prepass stride mismatch.

bdero:
Addressed the depth-prepass correctness issue, pool retention, release notes, smoke coverage, and PR description in `5d07740` and the follow-up CI commits. I kept the DataAssets migration integrated in this branch for now.

packages/flutter_scene/lib/src/scene_encoder.dart
bdero: Documented in `5d07740`. The 0.21.0 notes now call out batch-level translucent sorting.

packages/flutter_scene/lib/src/render/depth_prepass.dart
bdero: Fixed in `5d07740`. The prepass now selects 80-byte transform/color records for full layouts and 64-byte transform records only for depth-only layouts, including batched, cached, and single draws.

packages/flutter_scene/lib/src/scene_encoder.dart
bdero: Fixed in `5d07740`. Pooled records clear every scene and GPU reference before entering the static pools.

packages/flutter_scene/lib/src/render/depth_prepass.dart
bdero: Fixed in `5d07740`. Depth batching now reuses `InstanceDataBatch` and the shared packers instead of allocating flattened matrices and a synthetic `RenderItem`.

packages/flutter_scene/lib/src/render/instance_packing.dart
bdero: Done in `5d07740`.

packages/flutter_scene/lib/src/render/instance_packing.dart
bdero: Fixed in `5d07740`. The first winding pass records each batch’s homogeneous or mixed state, removing the second scan.

packages/flutter_scene/CHANGELOG.md
bdero: Added in `5d07740`. The 0.21.0 notes now cover the missing APIs, 80-byte stride break, exact sRGB conversion, and translucent sorting.

packages/flutter_scene/CHANGELOG.md
bdero: Fixed in `5d07740`. The core version and workspace constraints are now 0.21.0.

packages/flutter_scene/lib/src/fmat/build_materials.dart
bdero: Fixed in `5d07740`. `dataAssetsIfAvailable` is now the default, so the warning only appears for an actual fallback or explicit legacy choice.

packages/flutter_scene/lib/src/render/shadow_pass.dart
bdero: Verified locally on Metal, GLES, Vulkan, and WebGL2. The GitHub matrix is also green on Linux, Windows, web, Android GLES, and Android Vulkan.

packages/flutter_scene/lib/src/render/scene_pass.dart
bdero: Fixed in `5d07740`.

packages/flutter_scene/shaders/material_lighting.glsl
bdero: Kept the arithmetic selection and documented why in the shader. Experiments with a uniform ternary drove the GLES emulator host to about 4 GB before termination, while a source-level branch exceeded the 20-minute software GLES drive timeout, so the small default-path `mix` cost is the bounded option.

examples/smoke_render/lib/smoke_scenes.dart
bdero: Fixed in `5d07740`. `directional_shadow` now covers the default rotated filter.

packages/flutter_scene/lib/src/node.dart
bdero: Fixed in `5d07740`.

examples/smoke_render/integration_test/smoke_test.dart
bdero: Added in `5d07740`.

packages/flutter_scene/lib/src/render/render_graph.dart
bdero: Fixed in `5d07740`. The three sites share one accumulator, the cross-view behavior is documented, and the profile flag is noted in the README.

.github/workflows/smoke_render_jobs.yml
bdero: Fixed in `c6634b2` and `c512ba0`. Removed the unused environment and variable-size smoke machinery, keeping one fixed 512-pixel corpus.

packages/flutter_scene/test/instance_packing_test.dart
bdero: Added in `5d07740` and `d72983f`. Tests exercise the shared run planners and cached/uncached selection, while `depth_post` requests normals through a no-op custom pass to cover merged full-layout depth draws without unrelated SSR work.

.github/workflows/smoke_render_jobs.yml
bdero: Documented in `5d07740`. The workflow now explains that these diagnostics preserve host-pressure and process evidence when the emulator exits without an Android crash report.

bdero (commented):
Second pass over `c512ba0`. 17 of the 19 earlier items are cleanly resolved. The stride fix covers the batched, cached, uncached, and single-draw paths; the record pools release their references before pooling; the winding rescan is gone; the profiling triplication collapsed into `render_profile.dart`; and 0.21.0 landed across the pubspec and every workspace constraint. Pulling the run planners out into `instance_batching.dart` was a better answer than what I suggested, and `instance_batching_test.dart` is the coverage that was missing.

`flutter analyze` is clean and tests pass at 1101, up from 1097.

One blocking item came out of the refactor. `instanceDataBatchFor` unified four call sites that were not actually identical, so the shadow encoder now culls its casters against the camera frustum. Details inline.

One of the smaller notes is a correction to myself. My earlier nit about the redundant `instances = instances` initializer was wrong, and acting on it traded a compile-time guarantee for a debug assert. Also flagged inline.

packages/flutter_scene/lib/src/render/instance_batching.dart
```diff
@@ -0,0 +1,83 @@
...
+      nodeWindingFlipped: item.windingFlipped,
+    );
+  }
+  final packedWorldData = item.instanceWorldData;
+  final packedWinding = item.instanceWorldWindingFlipped;
+  if (packedWorldData != null && packedWinding != null) {
+    return InstanceDataBatch.cached(
+      packedWorldData: packedWorldData,
+      packedWindingFlipped: packedWinding,
+      indices: item.visibleInstanceIndices,
```
bdero: This helper always applies `item.visibleInstanceIndices`, but only two of its three callers should.

`visibleInstanceIndices` is written by `SceneEncoder.submit` and `DepthPrepass.submit`, both of which cull against the camera. `ShadowPass` is added to the graph first (`scene.dart:1355`, ahead of `DepthPrepass` at :1426 and the scene pass), so when `ShadowEncoder.flush` reaches this helper the list holds the previous frame's camera selection, or null on the first frame.

With `cullInstances = true` that drops instances outside the camera frustum from the shadow map, which is exactly the geometry a shadow map needs, and it does so one frame late.

Suggest an explicit indices argument or an `applyInstanceCulling` flag, so the color and prepass encoders pass `item.visibleInstanceIndices` and the shadow encoder passes null. Same applies to the uncached branch at line 81.

Nothing in the smoke matrix covers this. `instanced_lighting` leaves `cullInstances` at its default false and `directional_shadow` has no instanced meshes.

packages/flutter_scene/lib/src/render/shadow_encoder.dart
```diff
@@ -78,34 +79,74 @@ class ShadowEncoder {
...
+      ).compareTo(identityHashCode(b.geometry));
+    });
+    var index = 0;
+    while (index < _records.length) {
+      final first = _records[index];
+      final end = depthBatchEnd(_records, index);
+      if (end > index + 1) {
+        final batches = <InstanceDataBatch>[];
+        for (var batchIndex = index; batchIndex < end; batchIndex++) {
+          batches.add(instanceDataBatchFor(_records[batchIndex]));
```
bdero: This is the call site that picks up the camera-frustum instance selection described on `instance_batching.dart`. Before the refactor this loop built `InstanceDataBatch.cached` with no indices, so every instance rendered into the shadow map.

packages/flutter_scene/lib/src/render/shadow_encoder.dart
```diff
@@ -208,21 +265,43 @@ class ShadowEncoder {
...
-        bindInstanceTransforms(_renderPass, packed.cw, slot: instanceSlot);
-        _renderPass.setWindingOrder(gpu.WindingOrder.clockwise);
-        geometry.draw(_renderPass, instanceCount: packed.cwCount);
-      }
+      final packedWorldData = item.instanceWorldData;
+      final packedWinding = item.instanceWorldWindingFlipped;
+      final cached = packedWorldData == null || packedWinding == null
+          ? null
+          : [
+              InstanceDataBatch.cached(
```
bdero, line 275: The non-batched path still builds `InstanceDataBatch.cached` without indices, and the two uncached fallbacks below also omit them, so this branch keeps the old all-instances behavior while `flush` at line 137 now filters.

That means one `InstancedMesh` draws every instance into the shadow map when it encodes alone, and only the camera-visible subset when it happens to share geometry and material with a neighbour. Content-dependent and silent. Whichever way the helper is fixed, these two paths should agree.

packages/flutter_scene/test/instance_batching_test.dart
```diff
@@ -0,0 +1,162 @@
...
+      final single = _item(geometry, material);
+      expect(instanceDataBatchFor(single).instances, isNull);
+
+      final item = _item(geometry, material)
+        ..instanceTransforms = [Matrix4.identity(), Matrix4.identity()]
+        ..instanceColors = [Vector4.all(1), Vector4.all(1)]
+        ..visibleInstanceIndices = [1];
+      final uncached = instanceDataBatchFor(item);
+      expect(uncached.instances, same(item.instanceTransforms));
+      expect(uncached.indices, [1]);
```
bdero, line 152: This assertion pins the behavior that breaks the shadow pass. Once `instanceDataBatchFor` takes the indices explicitly, this case becomes the color/prepass one, and it would be worth a sibling asserting the shadow call keeps every instance.

packages/flutter_scene/lib/src/render/instance_packing.dart
```diff
@@ -14,16 +64,374 @@ import 'package:vector_math/vector_math.dart';
...
+  int get ccwCount => ccw.length ~/ 20;
+  @override
+  int get cwCount => cw.length ~/ 20;
+}
+
+/// One retained instance group consumed by [packInstanceDataBatches].
+class InstanceDataBatch {
+  InstanceDataBatch({
+    required this.nodeTransform,
+    required this.instances,
```
bdero: My earlier nit here was wrong, sorry. The field has to stay nullable for the `.single` and `.cached` constructors, so `required List<Matrix4> instances` plus an `instances = instances` initializer was narrowing rather than redundancy.

As written now, `required this.instances` accepts null at compile time and only the debug assert below catches it. Worth reverting to the non-nullable parameters and dropping the two null asserts they made necessary.

packages/flutter_scene/hook/build.dart
```diff
@@ -1,17 +1,20 @@
+import 'package:data_assets/data_assets.dart';
 import 'package:hooks/hooks.dart';
 
 import 'package:flutter_gpu_shaders/build.dart';
 
 void main(List<String> args) async {
   await build(args, (config, output) async {
+    if (!config.config.buildDataAssets) {
+      return;
```
bdero, line 11: This guard is right for code-only invocations, but `buildDataAssets` resolves to `buildAssetTypes.contains(DataAssetType.type)`, which is also false for every invocation when the project has not run `flutter config --enable-dart-data-assets`. So it returns before `buildShaderBundleJson` can raise its setup guidance.

That would be survivable, except this PR also drops `build/shaderbundles/base.shaderbundle` from `flutter.assets`, and `resolveBaseShaderBundleKey` falls back to exactly that key when the data asset is missing. The user ends up at `Failed to load base shader bundle! (packages/flutter_scene/build/shaderbundles/base.shaderbundle)` from `Scene.initializeStaticResources()`, which names a path instead of the config step they skipped.

The hook cannot tell the two cases apart, so the fix probably belongs in `loadBaseShaderLibrary`. Have it name `flutter config --enable-dart-data-assets` when neither key resolves, and leave a comment here saying the guard deliberately swallows that case.

examples/smoke_render/hook/build.dart
```diff
@@ -8,13 +8,15 @@ void main(List<String> args) {
       buildInput: config,
       buildOutput: output,
       materials: ['assets/custom_material.fmat', 'assets/noise_parity.fmat'],
+      assetMode: MaterialAssetMode.dataAssetsRequired,
```
bdero, line 15: This hook and `examples/flutter_app/hook/build.dart` both pass `dataAssetsRequired` but have no `buildDataAssets` guard, unlike `packages/flutter_scene/hook/build.dart`.

If the code-only invocation behind `a1b00e9` reaches dependency hooks in general, these two should be throwing the same way. Either they need the same guard or something else distinguishes them, and it would be good to know which before this lands.

packages/flutter_scene/CHANGELOG.md
```diff
@@ -7,7 +7,25 @@
 * Debug builds check a custom material's uniform blocks against the shader's reflection, so a misspelled or optimized-out block name and a buffer too small for its std140 layout raise named exceptions instead of failing at bind time or reading past the end.
 * Debug builds check a caller-declared vertex layout against the buffers actually bound, so a slot count mismatch raises instead of feeding the pipeline undefined vertex data.
 * `IndexType`, `VertexFormat`, and `VertexStepMode` are exported, so 32-bit indices and caller-declared layouts no longer need `lib/src` imports.
-
+* Made first-party shader/material hooks and examples DataAssets-only, with migration guidance for legacy assets.
```
bdero, line 10: `buildMaterials` also flipped its default `assetMode` from `legacyOnly` to `dataAssetsIfAvailable`, which changes behavior for third-party hooks, not just the first-party ones this line covers. Low risk in practice, since the generated files still land in `build/shaderbundles/` either way so existing pubspec entries keep resolving, but what gets registered changes and it deserves its own bullet.

.github/workflows/smoke_render_jobs.yml
```diff
@@ -65,6 +68,8 @@ jobs:
             xvfb libgl1-mesa-dri libegl-mesa0 libgles2 mesa-vulkan-drivers
       - run: flutter config --enable-linux-desktop --enable-native-assets
       - run: flutter pub get
+      # Keep host pressure and process evidence when SwiftShader or the emulator
```
bdero: This comment landed in the `linux` job, above `Render smoke scenes`, but the diagnostics it describes (`host_monitor`, the cgroup and dmesg dumps) live in the `android` job around line 178. As written it talks about an emulator and an Android crash report from inside the Linux desktop lane.

packages/flutter_scene/lib/src/scene_encoder.dart
```diff
@@ -150,6 +246,11 @@ void evictPipelinesForShaders(Set<gpu.Shader> shaders) {
 /// their `bind` callbacks, which receive the `gpu.RenderPass` and
 /// `TransientWriter` directly.
 base class SceneEncoder {
+  static final RenderProfileAccumulator _profile = RenderProfileAccumulator();
+  int _instancePackMicros = 0;
```
bdero, line 250: Now that these are instance fields reset inside `_recordProfile`, and `_recordProfile` only runs from `flushOpaque`, any packing done during `flushTranslucent` is never reported. The encoder is discarded before the next sample, so those microseconds and bytes are simply dropped rather than deferred.

Dev tooling only, and arguably fine, but the old statics did fold them into the following window.

examples/smoke_render/lib/smoke_scenes.dart
```diff
@@ -190,6 +194,20 @@ MeshGeometry _phaseGrid() {
     ..setCustomAttribute('phase', phase, components: 1);
 }
 
+class _NormalsProbePass extends CustomRenderPass {
```
bdero, line 197: Nice way to force the full-layout prepass without dragging in SSR.

Noting for the record that this moves `depth_post` off the depth-only 64-byte prepass path it used to cover. That path is still exercised through `directional_shadow`, since the shadow encoder takes `geometry.depthOnlyVertex`, so coverage is intact. Just worth knowing that the two layouts are now covered by different scenes.

packages/flutter_scene/lib/src/render/shadow_encoder.dart
bdero: Fixed in bf59965. Batched and non-batched shadow paths now both ignore camera-frustum instance indices.

packages/flutter_scene/lib/src/render/instance_packing.dart
bdero: Reverted in bf59965. The main constructor again takes non-nullable lists and narrows them into the nullable fields.

packages/flutter_scene/CHANGELOG.md
bdero: Added in bf59965.

packages/flutter_scene/test/instance_batching_test.dart
bdero: Added in bf59965. Cached and uncached tests now cover camera-selected batches and all-instance shadow batches.

packages/flutter_scene/hook/build.dart
bdero: Fixed in bf59965. The guard documents the ambiguous false case, and runtime loading names `flutter config --enable-dart-data-assets` when the bundle is unavailable.

.github/workflows/smoke_render_jobs.yml
bdero: Moved in bf59965. The diagnostics comment now sits in the Android job it describes.

packages/flutter_scene/lib/src/render/instance_batching.dart
bdero: Fixed in bf59965. Callers now pass instance indices explicitly, so color/depth use the camera selection while shadows pass null.

packages/flutter_scene/lib/src/render/shadow_encoder.dart
bdero: Fixed in bf59965. Batched shadow records now pass null indices and retain every caster instance.

examples/smoke_render/lib/smoke_scenes.dart
bdero: Agreed. The full-layout and depth-only paths remain covered by `depth_post` and `directional_shadow`, respectively.

examples/smoke_render/hook/build.dart
bdero: Fixed in bf59965. Both example hooks now use the same `buildDataAssets` guard.

packages/flutter_scene/lib/src/scene_encoder.dart
bdero: Fixed in bf59965. Profiling now records after the translucent phase and includes both phases in the frame totals.

PR #288, opened 2026-08-02
Predict the owned player with input-replay reconciliation

bdero:
The owned entity now predicts from local input and reconciles by replaying the inputs the server has not yet applied, replacing the error-smoothing stopgap that lagged by the round trip under real latency. `SceneReplication` grows a `localPrediction` seam supplying a `PredictedController`, and the Multiplayer example consumes input commands server-side with a prediction toggle and a simulated-latency control so the difference is visible on one machine.

Draft until the next `dashwire`/`dashwire_replication` minor publishes (bdero/dashwire#1); the constraint bump from `^0.1.0` lands here once it does, so CI stays red in the meantime.

### DIFF
--- a/packages/flutter_scene_net/lib/src/predicted_transform.dart
+++ b/packages/flutter_scene_net/lib/src/predicted_transform.dart
@@ -44,6 +44,7 @@ final class PredictedTransformComponent extends Component {
     required int tickRate,
     this.smoothing = const Duration(milliseconds: 120),
     this.correctionThreshold = 0.05,
+    this.maxCatchUpTicks = 32,
     NowMicros now = defaultNowMicros,
   }) : _dt = 1 / tickRate,
        _now = now {
@@ -65,6 +66,16 @@ final class PredictedTransformComponent extends Component {
   /// Positional divergence (metres) beyond which a snapshot triggers a replay
   /// correction.
   final double correctionThreshold;
+
+  /// Most ticks predicted in one [update] before the prediction snaps to
+  /// authority instead of catching up.
+  ///
+  /// The target tick comes from the wall clock, so a stall (a backgrounded
+  /// tab, a long hitch, a resume from pause) would otherwise sample, step and
+  /// send every missed tick in a single frame. The server's input buffer
+  /// drops commands older than its window anyway, so most of that catch-up is
+  /// discarded on arrival.
+  final int maxCatchUpTicks;
 
   final double _dt;
   final NowMicros _now;
@@ -99,6 +110,16 @@ final class PredictedTransformComponent extends Component {
       ));
       _reconciledTick = acked;
       if (corrected) _error.setFrom(rendered - _predictor.current.$1);
+    }
+
+    // Too far behind to catch up tick by tick, so snap to the newest
+    // authoritative pose and resume predicting from there.
+    if (target - _predictor.currentTick > maxCatchUpTicks) {
+      _predictor.reset(target - 1, (
+        replica.positionVector,
+        replica.rotationQuaternion,
+      ));
+      _error.setZero();
     }
 
     // Predict forward to the target tick, one input sampled and sent per tick.
--- a/packages/flutter_scene_net/lib/src/predicted_transform.dart
+++ b/packages/flutter_scene_net/lib/src/predicted_transform.dart
@@ -97,9 +97,15 @@ final class PredictedTransformComponent extends Component {
       ));
     }
 
-    // Reconcile against the authoritative pose the server has confirmed
-    // applying input through.
-    final acked = client.lastAppliedInputTick;
+    // Reconcile against the authoritative pose, at the tick that pose
+    // actually belongs to. The input ack goes out every tick while snapshots
+    // are priority-packed against a byte budget and can be deferred or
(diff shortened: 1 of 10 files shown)


bdero (commented):
Read the whole prediction path against the published `dashwire_replication` 0.2.0 sources. The seam is clean and the split between `localPrediction` and the interpolated default is the right shape. Three things inline look like real problems, and the first two show up as soon as the link is anything but perfect.

A few notes that do not fit on a diff line.

- `network_transform.dart:15` still carries `TODO(prediction): owned replicas currently render on the same delayed timeline as everything else; input-replay prediction needs the tick infrastructure wired through here.` This PR is what resolves it, so it should go with it.
- CI never runs `packages/flutter_scene_net`. The test matrix in `.github/workflows/flutter.yml` is `packages/scene` and `packages/flutter_scene` only, so both new tests here never execute on a push. Worth adding the package to the matrix in this PR, otherwise the prediction suite is write-only.
- The description says this is draft until the next `dashwire`/`dashwire_replication` minor publishes and that CI stays red meanwhile. `dashwire_replication` 0.2.0 is published, the pubspec already carries `^0.2.0`, and every check on this PR is green, so that blocker is gone and the description is stale.

packages/flutter_scene_net/lib/src/predicted_transform.dart
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    // Reconcile against the authoritative pose the server has confirmed
+    // applying input through.
+    final acked = client.lastAppliedInputTick;
+    if (acked > _reconciledTick &&
+        acked > 0 &&
+        acked <= _predictor.currentTick) {
+      final rendered = _predictor.current.$1 + _error;
+      final corrected = _predictor.reconcile(acked, (
```
bdero, line 120: `acked` is an input tick, but `replica.positionVector` is just the newest pose that has arrived, and the two only line up when the snapshot carrying this replica went out on that same server tick.

Upstream, `_flushInputAck` runs every tick for every active peer, while `_flushSnapshot` priority packs against a per peer byte budget and rides the unreliable channel. A deferred or dropped snapshot leaves the newest pose several ticks behind `lastAppliedInputTick`.

When that happens this call overwrites the predicted state at `acked` with a pose from `acked - k`, then replays only the inputs after `acked`, so k ticks of motion are silently discarded and re-predicted. The next snapshot repeats it, which reads as a steady backward tug rather than a one-off pop. The example's simulated link sets `unreliableLoss: 0.02`, so it is reachable in the demo, not just in theory.

The snapshot message already carries its tick and the client reads it to build the ack. Exposing it upstream (a `ReplicationClient.lastSnapshotTick`, say) and reconciling at that tick when it is not ahead of `lastAppliedInputTick` would pin the authoritative pose to the tick it actually belongs to.

packages/flutter_scene_net/lib/src/predicted_transform.dart
```diff
@@ -0,0 +1,126 @@
...
+
+  late final Predictor<Uint8List, (Vector3, Quaternion)> _predictor;
+  final Vector3 _error = Vector3.zero();
+  int _reconciledTick = -1;
+
+  @override
+  void update(double deltaSeconds) {
+    final clock = client.session.clock;
+    if (!clock.isSynchronized) return;
+    final target = clock.inputTickAt(_now());
```
bdero: This takes `inputTickAt`'s default 1.5 tick margin, but `ClientInputSender` keeps its own `_lead` and only spends it in `nextTick()`. Passing an explicit `tick:` to `sendInput` bypasses that, so the server's buffer depth ack still drives `_lead` up and down every tick and nothing ever reads it. The pacing control loop upstream never closes for a predicted entity, and `ReplicationClient.inputTargetDepth` has no effect on one.

The practical cost is that the send ahead cannot deepen under jitter, so `ServerInputBuffer` starves and holds the last input instead of buffering. Driving `target` from `client.nextInputTick()` would close the loop and cost nothing here, since the loop below already tolerates a target that does not advance.

packages/flutter_scene_net/lib/src/predicted_transform.dart
```diff
@@ -0,0 +1,126 @@
...
+      final corrected = _predictor.reconcile(acked, (
+        replica.positionVector,
+        replica.rotationQuaternion,
+      ));
+      _reconciledTick = acked;
+      if (corrected) _error.setFrom(rendered - _predictor.current.$1);
+    }
+
+    // Predict forward to the target tick, one input sampled and sent per tick.
+    while (_predictor.currentTick < target) {
```
bdero, line 139: `target` comes from the wall clock, so any stretch where `update` does not run (a backgrounded web tab, a long hitch, an app resumed from pause) makes this loop replay every missed tick in a single frame, sampling and sending an input command for each one. At 60Hz a few seconds away is hundreds of iterations, and in the physics subclass each iteration is a world step plus a full world snapshot.

`ServerInputBuffer.window` is 64 and drops the oldest beyond that, so most of that catch up is discarded on arrival anyway. Clamping the loop, and falling back to `reset(target - 1, ...)` when the gap exceeds the retained history, would bound both the hitch and the send burst.

PR #289, opened 2026-08-02
Add physics rollback prediction and the arena example

bdero:
Stacked on #287 and #288; it carries their commits and rebases to just its own once they merge.

`PredictedPhysicsComponent` rollback-predicts an owned body in a dedicated snapshot-capable world, restoring the acked tick, adopting the authoritative pose and velocity, and replaying pending inputs, with `onWorldRestored` telling the controller to revalidate bodies it manages. `PhysicsWorldHistory` rewinds server-side hit queries to the client-rendered tick. Remote players mirror into the prediction world as kinematic proxies at their rendered poses, and the interpolation delay now adapts to measured arrival cadence and jitter. The Multiplayer example becomes a 60Hz physics arena (players are balls that shove each other) with a headless `dart run` host tool, and prediction runs on the web behind a runtime capability probe.

### DIFF
--- a/examples/flutter_app/lib/example_multiplayer.dart
+++ b/examples/flutter_app/lib/example_multiplayer.dart
@@ -2,12 +2,12 @@ import 'dart:async';
 
 import 'package:dashwire/dashwire.dart';
 import 'package:dashwire_replication/dashwire_replication.dart' hide Authority;
-import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 import 'package:flutter_scene/physics.dart';
 import 'package:flutter_scene/scene.dart';
 import 'package:flutter_scene_net/flutter_scene_net.dart';
+import 'package:flutter_scene_rapier/flutter_scene_rapier.dart';
 import 'package:vector_math/vector_math.dart' as vm;
 
 import 'example_overlay.dart';
@@ -125,9 +125,28 @@ class _ExampleMultiplayerState extends State<ExampleMultiplayer> {
     ),
   );
 
+  /// Whether the physics backend on this platform can rollback-predict.
+  /// True on native; on the web it needs a wasm module carrying the
+  /// snapshot exports (an older published module degrades to interpolation).
+  Future<bool> _predictionSupported() async {
+    try {
+      await RapierWorld.ensureInitialized();
+      final probe = RapierWorld();
+      try {
+        probe.snapshot();
+      } finally {
+        probe.dispose();
+      }
+      return true;
+    } catch (_) {
+      return false;
+    }
+  }
+
   Future<void> _start(Future<WireConnection> Function() connect) async {
     setState(() => _status = 'connecting');
     try {
+      final canPredict = await _predictionSupported();
       final session = await connectSession(
         await connect(),
         schemaHash: multiplayerRegistry().schemaHash,
@@ -142,10 +161,8 @@ class _ExampleMultiplayerState extends State<ExampleMultiplayer> {
         // the world back to the acked tick and replay unacked inputs. Remote
         // players stay interpolated. When prediction is off the local player
         // interpolates too, so it visibly lags input under latency.
-        // TODO(wasm): enable on web once the rapier wasm carries the
-        // snapshot/restore exports and the Dart-side marshalling lands.
         localPrediction: (replica) {
-          if (!_predict || kIsWeb) return null;
+          if (!_predict || !canPredict) return null;
           final controller = _PlayerController(_pressed, replica as NetPlayer);
           _clientWorlds.add(controller.simulation);
           return controller;
--- a/examples/flutter_app/lib/example_multiplayer.dart
+++ b/examples/flutter_app/lib/example_multiplayer.dart
@@ -129,6 +129,19 @@ class _ExampleMultiplayerState extends State<ExampleMultiplayer> {
     ),
   );
 
+  /// Rendered poses of every player this client does not own, fed to the
+  /// prediction world as collision proxies.
+  Iterable<(Object, vm.Vector3)> _remotePlayerPoses() {
+    final replication = _replication;
+    if (replication == null) return const [];
+    return [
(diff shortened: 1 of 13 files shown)


bdero (commented):
Reviewed the rollback layer, the arena, and the rewind ring. The proxy comment is the one I would look at first, since I think it quietly disables the collision feature the last few commits were added for.

Ran `flutter test` in `packages/flutter_scene_net` twelve times on macOS, one failure, details on the test line.

Everything on #288 applies here too, in particular the reconcile tick pairing, which `PredictedPhysicsComponent` inherits verbatim and which matters more with real physics, since a stale base pose is replayed through a whole world step rather than an analytic one.

Smaller notes.

- `SceneReplication` still forwards only `interpolationDelay` as `NetworkTransformComponent.delay`. `minDelay` and `adaptive` are unreachable through the binding, so an app cannot floor the adaptive delay or turn it off without constructing the component itself.
- The CHANGELOG entry says `inputTargetDepth` at 1 trades jitter headroom for a tick less input latency, and the example repeats it in a comment. Per the pacing comment on #288, the prediction components never read the adapted lead, so as wired that knob does nothing for a predicted entity and the claim is not true yet.
- `PhysicsWorldHistory` is exported and unit tested, but nothing in the repo or the example uses it, so the server side rewind path has never run against a real backend or a real hit query. Worth either wiring it into the arena (pellet pickup rewound to the shooter's rendered tick) or holding it back until something consumes it.

examples/flutter_app/lib/example_multiplayer.dart
```diff
@@ -294,3 +452,128 @@ class _ExampleMultiplayerState extends State<ExampleMultiplayer> {
...
+  @override
+  void onWorldRestored() => _proxiesStale = true;
+
+  // Mirrors the other players as kinematic proxies at their rendered
+  // poses. Runs on fresh ticks only; world snapshots retain the proxies,
+  // so rollback replays collide against their historical poses.
+  void _syncProxies() {
+    if (_proxiesStale) {
+      _proxiesStale = false;
+      for (final handle in _everCreated) {
```
bdero: I do not think this holds on the rapier backend, and I think it turns the feature off after the first correction rather than protecting it.

`RapierWorld.restore` calls `_reconcileBodies`, which drops registry records for bodies the snapshot does not contain and then destroys any body the snapshot brought back that has no record. Its own docstring is explicit that a body destroyed after the snapshot "is removed again rather than left simulating with no owner" and that recreated handles are new. So a restore does not resurrect the proxies into a usable state; it reconciles them away.

Walking it through. A correction restores a snapshot, `onWorldRestored` marks the proxies stale, and the next fresh tick destroys them and recreates them with new handles. The next correction targets an acked tick that trails the predicted tick by the send ahead, so it restores a snapshot taken before that rebuild. That snapshot holds the old handles while the registry holds the new ones, `_reconcileBodies` drops the new records and destroys the old bodies, and the replay from the acked tick runs with no remote proxies at all. Since acked always trails, this repeats on every correction after the first.

The per tick recreation also resets kinematic target interpolation, so a proxy has no target delta to derive velocity from and imparts no push on contact even on the fresh ticks.

A fixed proxy pool allocated in the controller constructor, before the first snapshot is ever taken, with unused slots parked out of reach, sidesteps the whole thing. Nothing is created or destroyed after the first snapshot, so restores never touch the proxy set and `onWorldRestored` is not needed for this at all. That also retires the `_everCreated` list and its 256 cap, which currently depends on an unstated invariant that a forgotten handle is always older than `historyTicks`.

packages/flutter_scene_net/test/predicted_physics_test.dart
```diff
@@ -0,0 +1,226 @@
...
+          .getTranslation()
+          .x;
+      // Local input moved the predicted body right immediately.
+      expect(predictedX, greaterThan(1));
+      // The unpredictable shove happened and the rollback correction adopted
+      // it; the prediction tracks authority within the send-ahead lead, and
+      // the restore hook told the controller about the rollback.
+      expect(shoved, isTrue);
+      expect((predictedX - pawn.position.value.$1).abs(), lessThan(3));
+      expect(controller!.worldRestores, greaterThan(0));
```
bdero: This one is flaky. Twelve runs of `flutter test test/predicted_physics_test.dart` on macOS gave one failure here, `Expected: a value greater than <0>` against `Actual: <0>`, meaning no rollback happened inside the 800ms wall clock window. The looser assertions above survive the same run, so this is the tight one.

The test drives real time through `Stopwatch` and `Future.delayed`, and both prediction components already take a `NowMicros now` for exactly this reason. Stepping the clock and the room by hand would make the shove, the ack, and the correction deterministic instead of racing the scheduler. Worth fixing before this lands rather than after, since the package is not in the CI test matrix yet (see the note on #288) and the flake will surface the moment it is added.

packages/flutter_scene_net/lib/src/physics_world_history.dart
```diff
@@ -0,0 +1,52 @@
...
+  void record(int tick) {
+    _snapshots[tick] = simulation.snapshot();
+    _newestTick = tick;
+    _snapshots.removeWhere((t, _) => _newestTick - t > maxRewindTicks);
+  }
+
+  /// Runs [query] against the world as recorded at [tick] and returns its
+  /// result, or null when [tick] is outside the retained window. The present
+  /// world is restored before returning.
+  T? rewind<T>(int tick, T Function(PhysicsSimulation simulation) query) {
```
bdero: `rewind` returns null both for a tick outside the retained window and for a query that legitimately returned null, so a caller cannot tell a miss from a clean result without re-deriving the window itself. For the documented use, hit registration, "the raycast found nothing" and "this peer is too far behind to compensate" want different handling, so a record return or an explicit out of window signal would earn its keep.

Separately, each call takes a full extra world snapshot to hold the present, so N queries on one tick pay N serialize plus deserialize round trips of the whole world. When `record` is called once per tick after stepping, as the doc prescribes, the present is already `_snapshots[_newestTick]`, so restoring from that instead would make repeated queries on a tick close to free.

PR #296, opened 2026-08-04
Complete scene realization and rendering support for interactive demos

bdero:
Adds linked FScene realization, authored material loading, filtered physics queries, safe graph mutation, and imported mesh variant handling. Improves ambient occlusion, tone mapping, post-processing stage application, networking, and native asset compatibility for large interactive scenes.


packages/flutter_scene_rapier/lib/src/ffi/wasm_rapier_bindings.dart
```diff
@@ -1034,6 +1035,7 @@ class WasmRapierBindings extends RapierBindings {
       _f(maxDistance),
       _i(1),
       _i(flags),
+      _i(layerMask),
```
bdero: This is the most serious issue in the PR. `wasm_release.dart` still points at `flutter_scene_rapier-0.4.0`, whose exports take `(..., filter_flags, out)`. The extra argument is dropped at the JS boundary, so `layerMask` binds where `out` belongs and `_hitScratch` never arrives. Every raycast and shape cast then writes the hit struct to whatever address the mask happens to be.

The native side is gated by `abi_version`, but nothing gates the wasm. Worth adding an equivalent guard so a stale module fails loudly instead of scribbling on the heap.

packages/flutter_scene_rapier/native_binaries.json
```diff
@@ -1,5 +1,6 @@
 {
   "version": "0.4.0",
+  "abi_version": 1,
```
bdero: This says `1` while `hook/build.dart` now requires `2`, so the gate rejects every published prebuilt and falls through to a Rust source build. That needs cargo, and Android source builds currently fail on a null `code.cCompiler`.

The manifest has to be regenerated from a new tag before this lands.

packages/flutter_scene_rapier/hook/build.dart
```diff
@@ -32,6 +32,7 @@ import 'package:flutter_scene_rapier/src/prebuilt_binaries.dart';
 import 'package:hooks/hooks.dart';
 
 const _nativeLibraryName = 'flutter_scene_rapier_native';
+const _nativeAbiVersion = 2;
```
bdero: This constant is duplicated as a literal `2` in the release workflow's Python. They will drift on the next ABI change. Worth having the workflow read it from here, or at minimum cross-referencing both sides in a comment.

.github/workflows/rapier_native_binaries.yml
```diff
@@ -211,6 +211,7 @@ jobs:
 
           manifest = {
               "version": version,
+              "abi_version": 2,
```
bdero: The other half of the duplicated ABI version (see `hook/build.dart`). A mismatch here silently disables prebuilts for every consumer, which is a slow failure to notice.

packages/flutter_scene/lib/src/importer/src/fscene_emitter/fscene_emitter.dart
```diff
@@ -402,6 +421,29 @@ ComponentSpec _meshComponent(List<(LocalId, LocalId)> pairs) {
   );
 }
 
+ComponentSpec? _lightComponent(GltfPunctualLight light) {
+  final properties = <String, PropertyValue>{
+    'color': Vec3Value(light.color.clone()),
+    'intensity': DoubleValue(light.intensity),
```
bdero: `KHR_lights_punctual` specifies candela for point and spot, and lux for directional. `PointLight.intensity` is documented as the radiance at unit distance, and `DirectionalLightCodec` defaults to `3.0`. A sun exported at 100000 lux imports four orders of magnitude too bright.

This needs a unit conversion, or a `TODO(light-units)` saying what is missing and how to finish it. The new test asserts the raw property values, so it will not catch the mismatch either way.

packages/flutter_scene/lib/src/importer/src/fscene_emitter/fscene_emitter.dart
```diff
@@ -402,6 +421,29 @@ ComponentSpec _meshComponent(List<(LocalId, LocalId)> pairs) {
   );
 }
 
+ComponentSpec? _lightComponent(GltfPunctualLight light) {
```
bdero, line 425: Two more things about turning this on.

It is unconditional, so every existing `buildScenes` user gains the authored lights on their next build. A scene that already sets up its own lighting silently doubles up. Worth a CHANGELOG entry at least, possibly a flag.

And `castsShadow` is never emitted, so imported lights never cast shadows.

packages/flutter_scene/lib/src/runtime_importer/runtime_importer.dart
```diff
@@ -76,30 +76,57 @@ Future<Node> importGltf(
...
   final doc = input.doc;
+  final skinnedMeshes = <int>{};
+  final unskinnedMeshes = <int>{};
+  for (final node in doc.nodes) {
+    final mesh = node.mesh;
+    if (mesh == null) continue;
+    (node.skin == null ? unskinnedMeshes : skinnedMeshes).add(mesh);
+  }
+
+  _PackedPrimitiveVariants pack(int meshIndex, GltfMeshPrimitive primitive) {
```
bdero, line 104: The mesh-variant fix is mirrored between the two import paths here, which is right. Lights are not, so a `.glb` loaded through `importGltf` has no lights while the same file cooked through `buildScenes` does. Worth closing that gap in the same pass, or noting why the two differ.

packages/flutter_scene/lib/src/material/material.dart
```diff
@@ -184,6 +184,12 @@ abstract class Material {
   /// material.
   bool doubleSided = false;
 
+  /// Clip-space offset toward the camera used for coplanar surface details.
+  /// Zero leaves depth unchanged. Small positive values such as `0.00001`
+  /// keep an overlay in front of its supporting surface without moving its
+  /// world-space geometry.
+  double depthBias = 0.0;
```
bdero, line 195: The unit here is the part I would most want settled before it is public, since it also becomes an authored `.fscene` property.

As a raw NDC offset it is not portable in three ways. `camera.dart` builds a `[0,1]` clip depth projection, so on GLES and WebGL2 the viewport transform lands it in window `[0.5,1]` and the same value covers roughly half as many depth-buffer units as on Metal. The right magnitude also depends on near/far and on whether the attachment is 16, 24, or 32 bit. And a world in centimetres needs a different value from the same world in metres.

A view-space offset the encoder converts per draw, or a normalized count of depth-buffer units, keeps one authored value correct across backends, formats, and scales. Worth confirming the current behavior on the WebGL2 and llvmpipe smoke lanes before deciding.

Also, `0.00001` in the doc reads as a magic number tuned for one scene at one near/far.

packages/flutter_scene/shaders/flutter_scene_unskinned_body.glsl
```diff
@@ -46,6 +47,7 @@ void main() {
 
   v_position = vertex.world_position;
   gl_Position = frame_info.camera_transform * vec4(vertex.world_position, 1.0);
+  gl_Position.z -= frame_info.depth_bias * gl_Position.w;
```
bdero: Since the bias is applied in the shared vertex body, the shim cannot normalize it per backend, which is what makes the clip depth convention leak into the public API. If the conversion moves to the encoder this line becomes backend-independent.

packages/flutter_scene/lib/src/render/shadow_encoder.dart
```diff
@@ -193,6 +193,7 @@ class ShadowEncoder {
     // fragment shader ignores camera_position, but a material's Vertex() hook
     // reads it, so the real camera position is bound).
     void bindDraw(Matrix4 worldTransform) {
+      geometry.depthBias = item.material.depthBias;
```
bdero: `_lightSpaceMatrix` is passed as `cameraTransform` here, so a property documented as an offset toward the camera biases toward the light in the shadow map, changing acne and peter-panning for any biased material.

The depth prepass and the object filter clearly want it. This one reads unintentional. If it is deliberate, worth a line saying so.

packages/flutter_scene/lib/src/geometry/geometry.dart
```diff
@@ -691,6 +691,10 @@ abstract class Geometry {
   /// nodes carries the correct skeleton for every draw.
   void setJointsTexture(gpu.Texture? texture, int width) {}
 
+  /// Depth bias for the current draw, set by the active render encoder.
+  @internal
+  double depthBias = 0.0;
```
bdero: Two mechanisms carry one value now. Skinned geometry reads this field, unskinned takes a `depthBias:` parameter, and every encoder has to remember to set the field before `bind`. All four call sites do it correctly today, but the next encoder someone writes will silently inherit a stale bias.

A parameter on `Geometry.bind` would make it impossible to forget.

packages/flutter_scene/lib/src/ambient_occlusion.dart
```diff
@@ -27,16 +24,32 @@ class AmbientOcclusionSettings {
   /// contact creases.
   double radius = 0.33;
 
-  /// Strength of the darkening, applied as a power on the occlusion
-  /// factor. `1.0` is the raw estimate; higher values deepen the
-  /// occlusion and `0.0` removes it.
-  double intensity = 1.22;
+  /// Scalar applied to the averaged obscurance before it is converted to
+  /// visibility. `0.0` removes screen-space occlusion.
+  double intensity = 2.0;
```
bdero, line 34: `intensity` changed meaning (power on the factor, now a scalar on the obscurance) and default (`1.22` to `2.0`), with the old role moving to the new `power`. Anyone who tuned AO on 0.20.x gets a different image with no diagnostic. Fine for a pre-1.0 minor, but it needs an explicit CHANGELOG line naming the rename.

There is also no documented upper bound. The shader clamps at `min(intensity * obscurance, 0.98)`, so cranking it stops doing anything past a point.

packages/flutter_scene/lib/src/ambient_occlusion.dart
```diff
@@ -27,16 +24,32 @@ class AmbientOcclusionSettings {
...
   double bias = 0.07;
 
+  /// Minimum sine of the sample angle above the surface horizon. Samples
+  /// closer to tangent are rejected to prevent shallow surfaces from
+  /// self-occluding because of depth precision and coarse tessellation.
+  double horizonAngle = 0.06;
+
+  /// Fraction of screen-space occlusion applied to analytic direct lights.
+  /// `0.0` affects indirect lighting only, while `1.0` affects both equally.
+  double directLightAffect = 0.0;
```
bdero, line 56: `EnvironmentSettings` captures, applies, and lerps only `enabled`, `radius`, `intensity`, and `bias`, so environment-volume blending silently drops these four plus `specularMode`. The gap predates this PR but four more knobs makes it worth closing.

Defaulting `directLightAffect` to `0.0` is the right call, screen-space occlusion on analytic direct light is not physical and opt-in keeps the default correct.

Separately, the SAO citation was dropped from this file and from `flutter_scene_ssao.frag` while `flutter_scene_depth_downsample.frag` still cites the reference minifier and the new kernel comment still cites the reference kernel. The technique is still substantially SAO-derived, so the removal looks accidental.

bdero (commented):
`dart analyze` is clean and all 1121 `flutter_scene` tests pass locally. The mesh-variant fix, the nested override grammar, and the layer-mask query coverage are all solid.

Three things look like blockers.

1. The rapier ABI change ships without the release that makes it safe. `wasm_release.dart` still pins the 0.4.0 module, and the web path has no ABI gate, so every query passes `layerMask` where the old export expects the out pointer. `native_binaries.json` still declares `abi_version: 1` against the hook's new `2`, so native consumers fall through to a source build that is currently broken on Android. The package is still 0.4.0 with no CHANGELOG entry. This wants to be its own PR with its own release cycle.
2. glTF punctual light intensity is imported raw. The extension specifies candela and lux, while `PointLight.intensity` is an artistic radiance-at-unit-distance scalar, so authored lights land orders of magnitude too bright. It also turns on for every existing `buildScenes` user with no opt-out, and the runtime importer does not do it, so the two import paths now disagree about what a `.glb` contains.
3. Most of the new public API is missing from the CHANGELOG. `dart_apitool` will list all of it, worth running now rather than at release.

On the API design, `Material.depthBias` is the one I would most want to change before it is public. It is a raw NDC offset, so its meaning varies with the backend's clip depth convention, the depth attachment precision, and the scene's world scale. It also goes into `.fscene` documents as an authored property, so the unit is expensive to change later. A view-space offset converted per draw, or a normalized count of depth-buffer units, would stay meaningful across all three.

The rest is inline. Most of it is small.

packages/flutter_scene/lib/src/fscene/realize/stage.dart
```diff
@@ -118,6 +119,19 @@ Future<void> realizeStage(
   scene.environmentVolumes.clear();
 }
 
+void _applyStageLook(EnvironmentSettings settings, Scene scene) {
```
bdero: The underlying fix is right, `EnvironmentResource` carries no post-process or AO fields so applying a full snapshot was clobbering app settings with constructor defaults.

But this reimplements the first block of `EnvironmentSettings.applyTo`, and the two are already out of sync on day one, since `applyTo` sets `agxWhite`/`agxContrast` and this does not. Putting the split on `EnvironmentSettings` itself, with `applyTo` calling the look subset plus post-process, gives one definition and makes the contract legible from the class.

packages/flutter_scene/lib/src/scene.dart
```diff
@@ -571,6 +571,12 @@ base class Scene implements SceneGraph {
   /// display image. Defaults to [ToneMappingMode.pbrNeutral].
   ToneMappingMode toneMapping = ToneMappingMode.pbrNeutral;
 
+  /// AgX reference white. Only used by [ToneMappingMode.agx].
+  double agxWhite = 16.29;
```
bdero, line 575: `EnvironmentResource` has no fields for these, and stage serialization only writes `scene.toneMapping.name`, so a stage can select AgX but its curve tuning is lost on every editor save. Same shape as the `depthBias` gap on fmat materials.

packages/flutter_scene/lib/src/node.dart
```diff
@@ -963,8 +963,14 @@ base class Node implements SceneGraph {
         instancedMeshComponent.hideRenderItem();
       }
     }
-    for (final child in children) {
+    var childIndex = 0;
```
bdero: The guard handles detaching self, detaching an earlier sibling, and appending. It does not handle inserting before the current index, which re-ticks an already-ticked node, and `children` is a public mutable `List`, so `children.insert(0, x)` is reachable.

Since spawn and despawn during update is about the most common source of iteration bugs, this wants a stated guarantee, ideally on `Component.update` and `fixedUpdate` where users read it. Note the component loop right above carries a comment explaining what it tolerates while this stronger guard has none.

The guarantee is also incomplete in two places. `Node._mount` iterates `_components` and `children` with `for`-in while calling `component.mount()`, so a component that spawns a child on mount throws `ConcurrentModificationError`, which is the same bug in the spot most likely to hit it. And the component index loops still skip one when a component removes an earlier sibling component.

packages/flutter_scene/lib/src/render/ssao_pass.dart
```diff
@@ -21,9 +21,9 @@ const String kSsaoTextureBlackboardKey = 'ssao_texture';
 // [SsaoBlurPass].
 const String _kSsaoRawBlackboardKey = 'ssao_raw';
 
-// Single-channel, filterable, and color-renderable on every backend, which
-// is all the occlusion factor needs.
-const gpu.PixelFormat _aoFormat = gpu.PixelFormat.r8UNormInt;
+// Android GLES does not guarantee renderable single-channel normalized color
+// attachments. The shader only reads red, but RGBA8 keeps the pass portable.
+const gpu.PixelFormat _aoFormat = gpu.PixelFormat.r8g8b8a8UNormInt;
```
bdero, line 27: R8 is required color-renderable in GLES 3.0 core and in WebGL2, and the web shim already maps it in `web/texture.dart`, so the justification does not hold in general even if a specific driver misbehaves.

This is renderer-level backend handling paid on every backend, and it is not cheap. Four times the bytes per AO texel, plus the separable blur turns one target into two, so at 1080p that is roughly three RGBA8 transients where there was one R8. The main pass also gained nine depth fetches per pixel for the reconstructed normal and the detail taps, where the normal used to come free from `dFdx`/`dFdy`.

Given the `performance` label, worth either measuring or keeping R8 behind a shim-side or capability-probed fallback. If there is a real device behind this, naming it in a `TODO` would help the next person.

packages/flutter_scene/shaders/flutter_scene_ssao.frag
```diff
@@ -23,11 +21,14 @@ uniform SsaoInfo {
   vec4 viewport;
   // x: tan(fovX / 2). y: tan(fovY / 2). z: near plane. w: far plane.
   vec4 proj;
-  // x: radius (world units). y: bias (world units). z: intensity (final
-  // contrast power). w: projection scale (pixels per world unit at depth 1).
+  // x: radius (world units). y: bias (world units). z: obscurance intensity.
+  // w: projection scale (pixels per world unit at depth 1).
   vec4 params;
-  // x: sample count. y: mip level count (1 disables the chain).
+  // x: sample count. y: mip level count. z: minimum horizon cosine term.
```
bdero: This says cosine, but `SampleObscurance` computes `dot(normal, delta)/|delta|`, which is the sine of the elevation above the tangent plane. The Dart doc on `horizonAngle` has it right.

packages/flutter_scene/shaders/flutter_scene_ssao.frag
```diff
@@ -83,45 +140,36 @@ void main() {
...
-    // streaks on receding surfaces).
-    float sample_radius = sqrt((float(i) + 0.5) / float(sample_count));
-    float theta = float(i) * kGoldenAngle + rotation;
-    float pixel_radius = sample_radius * screen_radius;
+    // The linear radial progression and prime-ish turn counts follow the SAO
+    // reference kernel. Unlike sqrt spacing, this retains enough close taps to
+    // resolve contact occlusion without amplifying distant depth variation.
+    float sample_radius = (float(i) + 0.5) / float(sample_count);
+    float theta = sample_radius * spiral_turns * 2.0 * kPi + rotation;
+    float pixel_radius = max(1.0, sample_radius * screen_radius);
```
bdero: With the new linear `sample_radius` progression, a small `screen_radius` (distant geometry, small radius) clamps the first several taps to one pixel, so they land on nearly the same texel at different angles. Those fetches are wasted and they flatten the estimate. Worth checking what the low `screen_radius` case actually resolves to.

packages/flutter_scene/lib/src/importer/scene_registry.dart
```diff
@@ -375,6 +382,29 @@ final class SceneRegistry {
     );
   }
 
+  SceneDocument _resolveEagerRefs(SceneDocument document, String hostKey) {
+    final host = _entries.singleWhere((entry) => entry.assetKey == hostKey);
```
bdero: `singleWhere` throws a bare `StateError` for both no element and too many, so a ref that is not a registered entry crashes with nothing to act on. A `firstWhere` with an explanatory `orElse` would fail usefully.

Also, eager instances now resolve relative to the host document while `loadSubtree` still carries `TODO(lazy-relative-refs)` and treats refs as package-root-relative. The same authored `../models/model.glb` now works or fails depending on the instance's load policy, which is worth documenting on `PrefabInstanceSpec.load`.

packages/flutter_scene/lib/src/fmat/material_registry.dart
```diff
@@ -268,8 +268,12 @@ final class FmatMaterialRegistry {
     return manifest.listAssets();
   }
 
-  static Future<gpu.ShaderLibrary> _loadShaderLibrary(String assetKey) async {
-    final library = await gpu.loadShaderLibraryAsync(assetKey);
+  Future<gpu.ShaderLibrary> _loadShaderLibrary(String assetKey) async {
+    final library = identical(_bundle, rootBundle)
+        ? await gpu.loadShaderLibraryAsync(assetKey)
+        : await gpu.loadShaderLibraryFromBytesAsync(
```
bdero, line 276: flutter_gpu documents `fromBytes` results as uncached, with asset-path hot reload not applying, and offers `reinitializeFromBytes` as the counterpart. So routing non-root bundles through here loses shader hot reload for exactly the case this enables, the editor's disk-loaded materials, on both backends.

Worth wiring `reinitializeFromBytes`, or leaving a `TODO` recording the gap.

packages/flutter_scene/lib/src/gpu/web/shader_library.dart
```diff
@@ -62,10 +66,12 @@ base class ShaderLibrary {
       if (name == null || backend == null) continue;
       final shader = _buildFromBackend(backend);
       shaders[name] = shader;
-      _shadersByAsset.putIfAbsent(assetName, () => []).add((
-        name: name,
-        shader: WeakReference(shader),
-      ));
+      if (assetName != null) {
```
bdero, line 69: This is the web half of the same thing. Skipping registration means `reinitializeShaderLibraryAsync` has nothing to re-fetch for a bytes-loaded library.

packages/flutter_scene/lib/src/gpu/impeller/shader_library_inline.dart
```diff
@@ -22,6 +22,13 @@ Future<ShaderLibrary?> loadShaderLibraryAsync(String assetName) {
   return Future.value(ShaderLibrary.fromAsset(assetName));
 }
 
+/// Loads a shader bundle directly from [bytes].
+Future<ShaderLibrary?> loadShaderLibraryFromBytesAsync(ByteData bytes) {
+  // TODO(shader-bytes): drop the ignore once fromBytes reaches stable Flutter.
+  // ignore: undefined_method
```
bdero: `ShaderLibrary.fromBytes` exists in the SDK this builds against and `dart analyze` is clean without the ignore, so both lines can go.

Worth noting the TODO could not work as written anyway. An `// ignore` silences the analyzer, not the CFE, so on an SDK genuinely lacking the method this file fails to compile regardless.

packages/flutter_scene/lib/src/fscene/realize/builtin_codecs.dart
```diff
@@ -403,6 +403,7 @@ class MeshCodec extends ComponentCodec {
       'doubleSided': BoolValue(m.doubleSided),
       'alphaMode': StringValue(m.alphaMode.name),
       'alphaCutoff': DoubleValue(m.alphaCutoff),
+      if (m.depthBias != 0) 'depthBias': DoubleValue(m.depthBias),
```
bdero, line 410: `depthBias` serializes for PBR and unlit, and the realizer reads it for `fmat` resources too, but `_serializeFmat` never emits it. So a bias set on an fmat material silently drops on save.

packages/flutter_scene/lib/src/material/environment.dart
```diff
@@ -75,6 +75,21 @@ base class EnvironmentMap {
     );
   }
 
+  /// A black reflection environment with uniform diffuse [irradiance].
+  ///
+  /// The color is linear RGB irradiance after Lambertian convolution.
+  /// {@category Lighting and environment}
+  factory EnvironmentMap.constantDiffuse(Vector3 irradiance) {
```
bdero: The implementation mirrors `empty()` and is safe. The contract is the hard part. "Linear RGB irradiance after Lambertian convolution" is difficult to use, and no test asserts what a white diffuse surface actually shades to. For a convenience factory the natural contract is the ambient color such a surface shows.

It is also missing from the class doc's "Construct one with" list, which enumerates every other factory.

packages/flutter_scene/shaders/material_lighting.glsl
```diff
@@ -373,7 +373,9 @@ vec4 EvaluateLighting(MaterialInputs material) {
     // backend reports gl_FragCoord with a flipped origin, this sample needs
     // screen_uv.y = 1.0 - screen_uv.y; verify against the depth prepass on
     // each backend.
-    occlusion *= texture(ssao_texture, screen_uv).r;
+    // Both terms estimate the same visibility. Multiplying them counts the
+    // same blocked hemisphere twice and over-darkens surfaces with baked AO.
+    occlusion = min(occlusion, texture(ssao_texture, screen_uv).r);
```
bdero, line 378: Right call, and the comment explains why. It does visibly brighten any model carrying a baked occlusion map, so it deserves a CHANGELOG line.

packages/flutter_scene/CHANGELOG.md
```diff
@@ -40,6 +40,7 @@
 * Cooked textures build their mip chain on a background isolate, so a large uncompressed scene no longer pays for it on the calling thread.
 * Devices measured to sample every texture at its base mip skip mip chains entirely, reclaiming the memory and transcode time they were spending on levels the sampler ignores, and say so once at startup.
 * Loose images cook to compressed, mipmapped `.fstex` through `buildTextures`, loaded by source path with `loadTexture` (shipped in 0.20.0, missing from its notes).
+* Fixed ambient-occlusion banding and over-darkening with normalized sampling, robust normals, and depth-aware filtering.
```
bdero, line 44: This is the only entry the PR adds. Also missing: `ToneMappingMode.agx` with `Scene.agxWhite`/`agxContrast`, `Material.depthBias`, the four new `AmbientOcclusionSettings` knobs plus the changed meaning and default of `intensity`, `EnvironmentMap.constantDiffuse`, `ConstantEnvironment`, `TorusGeometrySpec`, `IcosphereGeometrySpec`, `ParticleSystem.prewarm` becoming a field, nested prefab override paths, glTF punctual lights importing as components, and the switch from multiply to `min` for baked versus screen-space occlusion.

packages/flutter_scene/lib/fscene.dart
```diff
@@ -87,6 +87,7 @@ export 'package:scene/scene.dart'
         BoundsSpec,
         ComponentSpec,
         CuboidGeometrySpec,
+        ConstantEnvironment,
```
bdero, line 89: Out of alphabetical order, `ConstantEnvironment` should precede `CuboidGeometrySpec`. Same for `TorusGeometrySpec` below, which lands before `StageMetadata`. Mirrored in `packages/scene/lib/scene.dart`.

examples/smoke_render/lib/smoke_scenes.dart
```diff
@@ -244,6 +259,75 @@ final List<SmokeScene> kSmokeScenes = <SmokeScene>[
...
+      scene: scene,
+      camera: PerspectiveCamera(
+        position: vm.Vector3(3.6, 2.4, 5.0),
+        target: vm.Vector3(0, -0.05, -0.4),
+      ),
+    );
+  }),
+  // AO contact continues across a surface that exits both horizontal edges.
+  // This catches viewport-boundary brightness ramps and discontinuities.
+  SmokeScene('ambient_occlusion_edge', () {
```
bdero, line 264: Two scenes for one feature area. The edge one earns its place with the numeric luma assertion; the first is a pure visual regression scene. Worth checking whether the assertion can cover both cases so the matrix only grows by one.

packages/flutter_scene_rapier/lib/src/ffi/wasm_rapier_bindings.dart
bdero: Removed the unreleased query-layer ABI change in 85256be. Native and wasm bindings remain compatible with the published 0.4.0 artifacts, so the ABI version machinery is no longer needed.

packages/flutter_scene_rapier/native_binaries.json
bdero: Removed the unreleased query-layer ABI change in 85256be. Native and wasm bindings remain compatible with the published 0.4.0 artifacts, so the ABI version machinery is no longer needed.

packages/flutter_scene_rapier/hook/build.dart
bdero: Removed the unreleased query-layer ABI change in 85256be. Native and wasm bindings remain compatible with the published 0.4.0 artifacts, so the ABI version machinery is no longer needed.

.github/workflows/rapier_native_binaries.yml
bdero: Removed the unreleased query-layer ABI change in 85256be. Native and wasm bindings remain compatible with the published 0.4.0 artifacts, so the ABI version machinery is no longer needed.

packages/flutter_scene/lib/src/importer/src/fscene_emitter/fscene_emitter.dart
bdero: Converted glTF photometric intensity to the renderer's linear radiometric scale through a shared helper in 85256be. The importer test now asserts the converted value.

packages/flutter_scene/lib/src/importer/src/fscene_emitter/fscene_emitter.dart
bdero: Added the import behavior to the changelog in 85256be. Directional and spot lights explicitly default to no shadows because the source extension carries no shadow intent and importing must not add shadow passes implicitly.

packages/flutter_scene/lib/src/runtime_importer/runtime_importer.dart
bdero: Both import paths now construct matching light components and use the same photometric conversion in 85256be.

packages/flutter_scene/lib/src/material/material.dart
bdero: Changed `depthBias` to a world-space offset toward the active camera before projection in 85256be. It no longer depends on clip convention, near/far, or depth format.

packages/flutter_scene/shaders/flutter_scene_unskinned_body.glsl
bdero: Changed `depthBias` to a world-space offset toward the active camera before projection in 85256be. It no longer depends on clip convention, near/far, or depth format.

packages/flutter_scene/lib/src/render/shadow_encoder.dart
bdero: Shadow draws now bind zero material depth bias in 85256be, so decals do not move toward the light or alter shadow-map biasing.

packages/flutter_scene/lib/src/geometry/geometry.dart
bdero: Removed `Geometry.depthBias` in 85256be and made bias a per-draw `Geometry.bind` argument. A forgotten binding can no longer inherit stale state.

packages/flutter_scene/lib/src/ambient_occlusion.dart
bdero: Documented the nonnegative range and saturation behavior in 85256be. The changelog now names the changed meaning, `2.0` default, and the move of contrast control to `power`.

packages/flutter_scene/lib/src/ambient_occlusion.dart
bdero: `EnvironmentSettings` now captures, applies, and interpolates every AO setting in 85256be. The technique citation was restored in the Dart and shader documentation.

packages/flutter_scene/lib/src/fscene/realize/stage.dart
bdero: Added `EnvironmentSettings.applyLookTo` in 85256be and made both stage realization and `applyTo` use the shared implementation.

packages/flutter_scene/lib/src/scene.dart
bdero: `EnvironmentResource` now carries AgX white and contrast through JSON, composition, copying, editor operations, realization, and serialization in 85256be.

packages/flutter_scene/lib/src/node.dart
bdero: Mount, unmount, frame, and fixed-step traversals now use identity-aware mutation-safe iteration in 85256be. The guarantees are documented and insertion/removal regressions are covered by tests.

packages/flutter_scene/lib/src/render/ssao_pass.dart
bdero: Tested R8 and RG8 on the Android GLES smoke lane in 85256be. Impeller aborts with incomplete or invalid attachments for both, so RGBA8 remains for portability and a TODO records the R8 backend dependency.

packages/flutter_scene/shaders/flutter_scene_ssao.frag
bdero: Changed the shader comment from cosine to sine in 85256be.

packages/flutter_scene/shaders/flutter_scene_ssao.frag
bdero: Subpixel wide-kernel taps are skipped in 85256be, avoiding repeated one-pixel samples. Wide and detail estimates are normalized independently.

packages/flutter_scene/lib/src/importer/scene_registry.dart
bdero: Added an actionable missing-registration error and made eager and lazy references resolve relative to their declaring document in 85256be. The load-policy documentation now states that guarantee.

packages/flutter_scene/lib/src/fmat/material_registry.dart
bdero: Removed the stale `fromBytes` ignore and added explicit byte-backed hot-reload TODOs at the material registry and web loader in 85256be.

packages/flutter_scene/lib/src/gpu/web/shader_library.dart
bdero: Removed the stale `fromBytes` ignore and added explicit byte-backed hot-reload TODOs at the material registry and web loader in 85256be.

packages/flutter_scene/lib/src/gpu/impeller/shader_library_inline.dart
bdero: Removed the stale `fromBytes` ignore and added explicit byte-backed hot-reload TODOs at the material registry and web loader in 85256be.

packages/flutter_scene/lib/src/fscene/realize/builtin_codecs.dart
bdero: Fmat serialization now emits `depthBias` in 85256be, and the override loader treats it as an engine property rather than a shader parameter.

packages/flutter_scene/lib/src/material/environment.dart
bdero: Defined the input as the ambient radiance received by a white Lambertian surface, added it to the factory list, and added an SH evaluation test in 85256be.

packages/flutter_scene/shaders/material_lighting.glsl
bdero: Added the switch from multiplication to `min` to the changelog in 85256be.

packages/flutter_scene/CHANGELOG.md
bdero: Audited the public API diff and added the missing tone mapping, AO, depth bias, environment, geometry, particle, linked-reference, shader-loading, and punctual-light entries in 85256be.

packages/flutter_scene/lib/fscene.dart
bdero: Corrected the alphabetical ordering in both barrels in 85256be.

examples/smoke_render/lib/smoke_scenes.dart
bdero: Combined the coverage into one AO edge scene in 85256be while retaining the numeric edge assertion and broad-surface regression geometry.

packages/flutter_scene/lib/src/node.dart
```diff
@@ -15,6 +17,25 @@ import 'package:flutter_scene/src/skin.dart';
...
+  final visited = HashSet<T>.identity();
+  var index = 0;
+  while (index < items.length) {
+    final item = items[index];
+    if (!visited.add(item)) {
+      index++;
+      continue;
+    }
+    visit(item);
+    final currentIndex = items.indexWhere(
```
bdero: This makes the traversal O(n^2). `indexWhere` scans the list after every visit, and it runs per node per frame from `scenePrePass` plus per fixed substep from `sceneFixedPass`. A node with many siblings pays quadratically.

Measured locally, comparing this against the same guarantees with a fast-path identity check first:

```
children=64    landed=11.0us    guarded=1.5us     7x
children=256   landed=55.8us    guarded=1.4us    40x
children=1024  landed=784.5us   guarded=6.9us   114x
children=4096  landed=12101.9us guarded=23.8us  508x
```

At 4096 siblings one list eats a whole frame budget. A chunked open world with a few thousand nodes under one root is squarely in that range.

The unmutated case is overwhelmingly the common one, and it can stay O(1) per item:

```dart
visit(item);
if (index < items.length && identical(items[index], item)) {
  index++;
  continue;
}
// the list moved, fall back to the tracking path
```

Only a traversal that actually observes a mutation then pays for the scan. Same guarantees, and it keeps the previous version's cost in the normal path.

packages/flutter_scene/lib/src/node.dart
```diff
@@ -15,6 +17,25 @@ import 'package:flutter_scene/src/skin.dart';
 import 'package:vector_math/vector_math.dart';
 import 'package:vector_math/vector_math.dart' as vm;
 
+void _visitMutable<T>(List<T> items, void Function(T item) visit) {
+  final visited = HashSet<T>.identity();
```
bdero: This allocates a `HashSet` per call, so `scenePrePass` alone is two per node per frame, and `sceneFixedPass` adds two more per substep. On a 10k-node scene that is tens of thousands of set allocations per frame plus every element inserted, in the hot path the bind-memo work was specifically tightening.

Worth making it lazy, allocated only after a mutation is actually observed. Combined with the fast path above, an unmutated frame then allocates nothing.

packages/flutter_scene/lib/src/components/component.dart
```diff
@@ -54,6 +54,9 @@ abstract class Component {
 
   /// Called once per frame while the component is mounted, [enabled], and
   /// loaded. [deltaSeconds] is the elapsed time since the previous tick.
+  /// A traversal visits each component at most once. Removing this component
```
bdero, line 57: Good to have the contract written down, and the three cases named are the ones that matter.

Reordering is the one case left unstated, and it fails loudly. Given `children = [A,B,C,D]`, if visiting `A` moves `A` to the end, `indexWhere` returns 3, `index` becomes 4, and B, C, and D are silently skipped for that frame. Worth either adding a sentence that reordering during traversal is unsupported, or making the fallback resume from the earliest unvisited item instead of from the moved item's new position.

packages/flutter_scene/lib/src/material/material.dart
```diff
@@ -184,6 +184,11 @@ abstract class Material {
   /// material.
   bool doubleSided = false;
 
+  /// World-space offset toward the camera used for coplanar surface details.
```
bdero, line 187: Much better contract, this is now independent of everything I raised.

One limitation is worth recording, because the target range includes large worlds. A constant world offset does not track the depth buffer's precision curve. With the default `near` of 0.1 and `far` of 1000 on a 24-bit buffer, the separation a bias buys falls off as `1/z^2`, so a 1mm bias is about 1600 depth units at 1m, 17 at 10m, and under 1 by roughly 30m. A decal tuned up close quietly starts z-fighting in the distance.

That is inherent to any world-space bias and it is still far better than the clip-space version, so a doc note or a `TODO(depth-bias-distance)` about a distance-compensated or slope-scaled variant would be enough.

Also worth documenting that negative values are a no-op, since `ApplyDepthBias` early-returns on `depth_bias <= 0.0`. Pushing a shell slightly away from the camera is a reasonable thing to want, so either support it or say it is unsupported.

packages/flutter_scene/shaders/flutter_scene_unskinned_body.glsl
```diff
@@ -7,9 +7,20 @@
...
 frame_info;
 
+vec3 ApplyDepthBias(vec3 world_position) {
+  vec3 to_camera = frame_info.camera_position - world_position;
+  float distance_squared = dot(to_camera, to_camera);
+  if (frame_info.depth_bias <= 0.0 || distance_squared <= 1e-12) {
+    return world_position;
+  }
+  return world_position +
+      to_camera * min(frame_info.depth_bias * inversesqrt(distance_squared), 1.0);
```
bdero: The `min(..., 1.0)` clamp caps the offset at the full distance to the eye, so a bias at or beyond that distance lands the vertex exactly at the camera and `w` goes to zero. Clamping slightly under, say `0.99`, avoids the degenerate perspective divide.

packages/flutter_scene/shaders/flutter_scene_unskinned_body.glsl
```diff
@@ -7,9 +7,20 @@
 uniform FrameInfo {
   mat4 camera_transform;
   vec3 camera_position;
+  float depth_bias;
 }
 frame_info;
 
+vec3 ApplyDepthBias(vec3 world_position) {
```
bdero: `ApplyDepthBias` is duplicated verbatim across the skinned, unskinned, and unskinned-depth bodies. Same shape as the `InterleavedGradientNoise` duplication in the two SSAO shaders. The repo already shares `.glsl` includes, so these could live in one.

packages/flutter_scene/shaders/flutter_scene_ssao.frag
```diff
@@ -83,45 +144,39 @@ void main() {
...
-    // streaks on receding surfaces).
-    float sample_radius = sqrt((float(i) + 0.5) / float(sample_count));
-    float theta = float(i) * kGoldenAngle + rotation;
+    // The linear radial progression and prime-ish turn counts follow the SAO
+    // reference kernel. Unlike sqrt spacing, this retains enough close taps to
+    // resolve contact occlusion without amplifying distant depth variation.
+    float sample_radius = (float(i) + 0.5) / float(sample_count);
+    float theta = sample_radius * spiral_turns * 2.0 * kPi + rotation;
     float pixel_radius = sample_radius * screen_radius;
+    if (pixel_radius < 1.0) {
```
bdero: Skipping subpixel taps is the right fix for the repeated one-pixel sampling, but it trades that for a discrete transition. As `screen_radius` shrinks with distance, taps drop out one at a time and `weight_sum` renormalizes over a changing support, so the wide term's spatial reach steps rather than fading.

That is the artifact class this whole AO pass set out to remove, and it would show up as depth-banded rings on a receding surface. The `ambient_occlusion_edge` smoke scene has a plane running away from the camera, so it may already cover it. Worth a look at that output before merging.

packages/flutter_scene/shaders/flutter_scene_ssao.frag
```diff
@@ -137,32 +192,36 @@ void main() {
...
+      falloff_scale * 4.0, horizon);
+  detail_sum += SampleObscurance(
+      normal, ViewPositionBase(center_coord + ivec2(0, -1)) - origin,
+      falloff_scale * 4.0, horizon);
+  detail_sum += SampleObscurance(
+      normal, ViewPositionBase(center_coord + ivec2(0, 1)) - origin,
+      falloff_scale * 4.0, horizon);
+
+  float wide_obscurance = weight_sum > 0.0 ? sum / weight_sum : 0.0;
+  float obscurance = wide_obscurance + detail * detail_sum * 0.25;
```
bdero, line 224: The detail term now normalizes by a fixed `0.25` while the wide term normalizes by `weight_sum`, so the two are independently scaled where they previously shared one denominator. That changes the balance `detail` controls, and `intensity` now scales the sum of two differently-normalized quantities.

The reasoning is sound, but the defaults (`detail = 0.5`, `intensity = 2.0`) were picked against the old normalization. Worth confirming they still land where you want, since they are also the values `EnvironmentSettings` now carries.

packages/flutter_scene/lib/src/importer/gltf_light_units.dart
```diff
@@ -0,0 +1,10 @@
+import 'package:flutter_scene/src/importer/src/gltf/types.dart';
+
+/// Converts glTF photometric intensity to the renderer's linear radiometric
+/// scale. The 683 lm/W constant is the peak photopic luminous efficacy.
+double gltfLightIntensity(GltfPunctualLight light) {
+  final color = light.color;
+  final luminance = color.x * 0.2126 + color.y * 0.7152 + color.z * 0.0722;
+  if (luminance <= 1e-8 || light.intensity <= 0.0) return 0.0;
+  return light.intensity / (683.0 * luminance);
```
bdero, line 11: The comment explains the 683 but not the luminance division, which is the surprising half. Dividing by the color's luminance so that `color * intensity` reproduces the authored luminance is a recognized convention, but it means a pure blue light gets about 14 times the multiplier of a white one at the same authored intensity. Worth one more sentence, since anyone debugging a too-bright colored light will land here.

Separately, the conversion fixes the order of magnitude but may not land in a usable range. A 100000 lux sun becomes about 146, against a `DirectionalLightCodec` default of 3.0 and a scene `exposure` default of 1.0. That is a big improvement on four orders of magnitude, but an authored sun will still blow out unless the app sets `physicalCameraExposure`. The new test asserts the conversion happened, not that the result is usable. Worth checking one real exported scene end to end.

One edge case, `light.intensity <= 0.0` returns `0.0`, which folds a negative authored intensity to zero silently. Fine, just noting it.

packages/flutter_scene/lib/src/environment_settings.dart
```diff
@@ -67,8 +70,16 @@ class EnvironmentSettings {
...
+    this.ambientOcclusionIntensity = 2.0,
     this.ambientOcclusionBias = 0.07,
+    this.ambientOcclusionPower = 1.5,
+    this.ambientOcclusionDetail = 0.5,
+    this.ambientOcclusionHorizonAngle = 0.06,
+    this.ambientOcclusionDirectLightAffect = 0.0,
+    this.ambientOcclusionSampleCount = 16,
+    this.ambientOcclusionHalfResolution = true,
+    this.ambientOcclusionDepthMipChain = false,
+    this.ambientOcclusionSpecularMode = SpecularAmbientOcclusionMode.none,
```
bdero, line 84: Not blocking, and the coverage gap I raised is genuinely closed.

Worth noting where this is heading though. The constructor is now 42 parameters, and every new `Scene` look knob costs five edits here (field, constructor, `fromScene`, `applyTo`, `lerp`). This PR paid that twice, for AgX and for AO. As post-processing and lighting keep growing, a flat mirror of `Scene`'s look state will keep getting wider and easier to miss a spot in, which is exactly how the AO fields fell behind in the first place.

Composing per-feature snapshots that each know how to capture, apply, and lerp themselves would make adding a knob a one-place change. Probably its own PR, but worth an issue before the next feature lands on it.

packages/flutter_scene/CHANGELOG.md
```diff
@@ -40,6 +40,19 @@
...
+* Ambient occlusion gains power, detail, horizon, direct-light, depth-mip, and specular controls. Intensity now scales obscurance, defaults to `2.0`, and its former contrast role is `power`.
+* Baked and screen-space ambient occlusion now combine with `min` instead of multiplication.
+* Added `ToneMappingMode.agx` with configurable `Scene.agxWhite` and `Scene.agxContrast`, persisted by environment resources.
+* Added `Material.depthBias`, a world-space camera offset for coplanar surface details.
+* Breaking change, custom `Geometry.bind` overrides accept the per-draw `depthBias` value.
```
bdero: Correct call to flag this as breaking, Dart requires an override to accept the supertype's named parameters, so every external `Geometry` subclass needs the new one.

Worth naming the fix in the entry, something like adding `double depthBias = 0.0` to the override's parameter list, so nobody has to diff the signature to find out what to add.

bdero (commented):
Re-reviewed at 85256be. `dart analyze` is clean across all four packages and 1124 tests pass.

The three blockers are properly resolved. Backing the rapier layer-mask change out entirely was the right call, and the revert is complete, including restoring the `TODO(query-layer-mask)` so the docs no longer overstate what the parameter does. The photometric conversion through one shared helper used by both import paths is exactly right. The CHANGELOG audit looks thorough, and flagging the `Geometry.bind` signature as breaking is the correct read, since Dart requires overrides to accept the supertype's named parameters.

Moving `depthBias` to a world-space offset is a real improvement. It is now projection, depth-format, and clip-convention independent, lighting still uses the unbiased position, and the shadow pass binds zero. Making bias a `bind` parameter instead of geometry state closed the stale-state hazard properly.

One new issue, in `_visitMutable`. It is correct, but it is quadratic and allocates per node per frame, in the traversal that runs for every node every frame and every fixed substep. Measured on this machine, one 1024-element list costs 784us per pass and 4096 costs 12.1ms, against 6.9us and 23.8us for the same guarantees with a fast-path identity check. That is a scene-killing regression for exactly the wide-sibling case the engine is meant to scale to, so I would not land it as written. Details inline.

The rest is follow-ups and two things worth eyeballing on the smoke lane.

packages/flutter_scene/lib/src/node.dart
bdero: Added the identity fast path. Unmutated traversal is linear, and scanning occurs only after observed mutation in 33b6b72.

packages/flutter_scene/lib/src/node.dart
bdero: Identity tracking is now allocated lazily after observed mutation. Unmutated traversal allocates nothing in 33b6b72.

packages/flutter_scene/lib/src/components/component.dart
bdero: Documented reordering component or child lists during traversal as unsupported in 33b6b72.

packages/flutter_scene/lib/src/material/material.dart
bdero: Documented nonpositive values and distance-related precision loss. Added a TODO for a distance-compensated or slope-scaled mode in 33b6b72.

packages/flutter_scene/shaders/flutter_scene_unskinned_body.glsl
bdero: The shared depth-bias helper now clamps the distance fraction to 0.99, avoiding a vertex at the eye in 33b6b72.

packages/flutter_scene/shaders/flutter_scene_unskinned_body.glsl
bdero: Moved depth bias and interleaved gradient noise into shared GLSL includes in 33b6b72.

packages/flutter_scene/shaders/flutter_scene_ssao.frag
bdero: Replaced discrete tap removal with smooth subpixel support while retaining stable normalization. The receding-plane smoke is smooth across Metal, software GLES, WebGL2, and Vulkan.

packages/flutter_scene/shaders/flutter_scene_ssao.frag
bdero: Documented both terms as independently normalized means. I retained the defaults after inspecting both AO smoke scenes and verifying consistent output across backends.

packages/flutter_scene/lib/src/importer/gltf_light_units.dart
bdero: Documented the color-luminance normalization and added coverage for a 100000 lux white light and a pure blue light. The changelog now directs photometric lights toward physical camera exposure.

packages/flutter_scene/lib/src/environment_settings.dart
bdero: Added a focused TODO for composable per-feature settings snapshots in 33b6b72.

packages/flutter_scene/CHANGELOG.md
bdero: The breaking-change entry now gives the exact `double depthBias = 0.0` parameter external overrides must add.

packages/flutter_scene/lib/src/node.dart
```diff
@@ -15,6 +17,43 @@ import 'package:flutter_scene/src/skin.dart';
 import 'package:vector_math/vector_math.dart';
 import 'package:vector_math/vector_math.dart' as vm;
 
+void _visitMutable<T>(List<T> items, void Function(T item) visit) {
+  HashSet<T>? visited;
+  var index = 0;
```
bdero, line 22: Not blocking, and the case that mattered is fixed.

The `index = 0` restart makes the mutation path O(k*n), so each observed mutation re-walks the list from the start. Overhead above the cost of the removals themselves, best of 60 runs:

```
4096 children, 1 despawn     +162us
4096 children, 8 despawns    +249us
4096 children, 64 despawns   +874us
4096 children, 256 despawns  +2891us
```

A single node detaching itself in a wide sibling list costs 162us, which is the common case rather than a pathological one.

The restart is only needed for insert-before. When `indexWhere` returns -1 the item was removed, so whatever sits at `index` now is the next unvisited element, and the loop can continue from there with no set and no rescan. That is what the earlier while-loop guard did, and it covers despawn in O(1). Reserving the set and the restart for `currentIndex > index` would keep both paths cheap.

Happy to see this as a follow-up issue.

bdero (commented):
Re-reviewed at 33b6b72. This looks good to merge on substance.

`dart analyze` is clean across flutter_scene, scene, flutter_scene_rapier, and flutter_scene_editor_core, and 1124 tests pass.

The traversal fix works. Measured again on the same bench, an unmutated 4096-child list went from 12.1ms to 25us, and 16384 children now cost 98us. The fast path allocates nothing, and seeding the tracking set from the current prefix is sound for each of the documented mutation shapes, including the insert-before case where the newly inserted node is correctly held back to the next frame. Reordering is the one case that still skips siblings, and documenting it as unsupported is the right call given the cost of handling it.

Fading subpixel taps with `smoothstep` while leaving `weight_sum` complete is a better fix than dropping them, since the denominator stays stable and distant support ramps to zero instead of stepping. The shared `depth_bias.glsl` and `interleaved_gradient_noise.glsl` includes match how `pbr.glsl` and `material_lighting.glsl` already work, so they need no manifest entry.

The rapier change is now just normalizing the cached prebuilt's filename through `targetOS.dylibFileName`, with a test. No ABI change, no manifest change, no release coupling. That is the right place to have landed.

One performance follow-up inline, worth an issue rather than another round. Three mechanical gates before merging are in the summary below.

Merge gates, none of them about the code:

1. The branch is `CONFLICTING` against master. The conflict is only `CHANGELOG.md`, where master's `a28bd88` added a bullet to the same 0.21.0 section. A rebase resolves it.
2. CI's last green run is 85256be, not 33b6b72. Worth letting Flutter CI and smoke render finish on the final commit.
3. `flutter_scene_net` now requires `dashwire_replication ^0.2.0` while its published 0.1.0 pins `^0.1.0`, which overlaps what stage B in #288 is scoped to do. Worth deciding which PR carries the bump so the two do not fight, and remembering `flutter_scene_net` needs a release either way.

packages/flutter_scene/lib/src/node.dart
bdero: Handled in 7a89a1a. Self-detachment now continues at the shifted cursor without a scan, set allocation, or restart, while tracking and restart are reserved for insertion before the cursor.

bdero:
Rebased onto current master. This PR carries the dependency constraint bump required by the demo, while the networking PR can inherit it and the package can release with its next version.

Issue #210, opened 2026-07-02
Support KHR_materials_emissive_strength


bdero:
Honor the emissive-strength extension so emissive surfaces can exceed the [0, 1] range and read as genuinely bright light sources under bloom and tone mapping. Small change (one scalar multiplied into the emissive term) plus importer support.

bdero:
Resolved by #297

Issue #207, opened 2026-07-02
Support KHR_materials_sheen


bdero:
Add a sheen lobe (Charlie distribution) for cloth and fabric (velvet, satin). Adds sheenColor and sheenRoughness inputs plus the retroreflective sheen term in the lighting shader. Part of the same material-completeness family as clearcoat (#206) and transmission (#127).

bdero:
Resolved by #297

Issue #206, opened 2026-07-02
Support KHR_materials_clearcoat


bdero:
Add a clearcoat lobe (a second thin specular layer with its own roughness and optional normal map) for car paint, lacquered wood, and coated plastics. Sibling to the transmission work in #127. Adds clearcoat, clearcoatRoughness, and clearcoat normal inputs to PhysicallyBasedMaterial plus the extra specular lobe in the lighting shader.

bdero:
Resolved by #297

Issue #234, opened 2026-07-12
directionalLight with shadow create pulsating black stripes in the distance on Android/mobile only


ali-solak:
DirectionalLight with` castShadow: true` on the example physics_car scene creates these stripes on the ground in the far distance. on Android Vulkan. Desktop is not reproducable. The stripes crawl/pulsate

changing shadowDepthBias and shadowNormalBias makes the stripes disappear. which makes it seem to be bias tunable shadow acne.

margin subtracted in material_lighting.glsl:91` (shadow_bias / (7.0 * box))` seems to shrink toward 0 in the distance and  far away ground loses its bias and self hadows.

<img width="2048" height="921" alt="Image" src="https://github.com/user-attachments/assets/5a16ba79-9335-49af-b48d-d99a78101ab1" />

bdero:
Hi! This looks to me like I have the depth bias tuned too low by default, or I didn't use floats with high enough precision in some places. Can you please print the results of `flutter doctor`, and also let me know the specific Android device you're testing with + which Impeller backend you are using? I have been unable to reproduce this on the latest master against a Pixel 10 or via AVD.

ali-solak:
Doctor summary (to see all details, run flutter doctor -v):
[√] Flutter (Channel master, 3.46.0-1.0.pre-510, on Microsoft Windows [Version 10.0.26200.8737])
[√] Windows Version (11 Home 64-bit, 25H2, 2009)
[√] Android toolchain - develop for Android devices (Android SDK version 36.0.0)
[X] Chrome - develop for the web (Cannot find Chrome executable at .\Google\Chrome\Application\chrome.exe)
    ! Cannot find Chrome. Try setting CHROME_EXECUTABLE to a Chrome executable.
[√] Visual Studio - develop Windows apps (Visual Studio Community 2026 18.7.3)
[√] Connected device (2 available)
[√] Network resources

Pixel 8, Vulkan backend

ali-solak:
removing ` #include <impeller/constants.glsl>` in the pbr.glsl works

https://github.com/flutter-team-archive/engine/pull/37828

Added precision mediump sampler2D.

My Pixel 8 with Mali GPU on Vulkan gets relaxed precision

bdero:
[...]

PR #297, opened 2026-08-04
Add advanced physical materials

### LINKED ISSUE #127: Support KHR_materials_transmission for see-through materials
Transmissive glTF materials (glass, lenses, clear plastic) render as opaque surfaces because the renderer doesn't implement the `KHR_materials_transmission` extension.

Current: 

Image

Correct:

Image

## Where it shows

In the Stress Tests example:
- Flight Helmet's goggle lenses (`LensesMat`, `alphaMode = OPAQUE` + `KHR_materials_transmission`) render as solid white plastic.
- Water Bottle reads as opaque white plastic for the same reason (its catalog description already notes this).

These are not alpha-blend materials. `KHR_materials_transmission` is physically-based light transmission: the surface is lit normally and also lets the background show through (optionally refracted via `KHR_materials_volume`).

## What it needs

The standard approach (model-viewer, Filament, three.js): render the opaque scene first, then for transmissive materials sample that already-rendered color buffer behind the fragment as the transmitted light, weighted by the transmission factor (and refracted/blurred by roughness + volume thickness for the fuller effect).

Concretely for flutter_scene:
- Parse the `KHR_materials_transmission` extension (transmission factor + texture) into the material. The glTF parser currently drops material `extensions`.
- A render-graph pass that exposes the opaque scene color as a sampleable texture to a later transmissive pass.
- Standard shader support: mix in the sampled background by the transmission factor.
- Optionally `KHR_materials_volume` (thickness, attenuation, IOR) for refraction.

Substantial; a reasonable first cut is non-refractive transmission (sample straight back, no volume).

Found while adding multi-file glTF + alpha modes (PR #126).

bdero:
Resolves #127, #206, #207, and #210.

Adds the advanced physical material pipeline, glTF import support, UV1 handling, layered transmission rendering, and built-in shader resources. Expands the example and smoke suites with material stress coverage and camera controls.


packages/flutter_scene/lib/src/material/advanced_physical_material.dart
```diff
@@ -0,0 +1,395 @@
...
+      if (d.iridescence > 0.0) ...[
+        (10, d.iridescenceTexture),
+        (11, d.iridescenceThicknessTexture),
+      ],
+      if (d.anisotropy != 0.0) (12, d.anisotropyTexture),
+      if (d.diffuseTransmission > 0.0) ...[
+        (13, d.diffuseTransmissionTexture),
+        (14, d.diffuseTransmissionColorTexture),
+      ],
+      (2, d.occlusionTexture),
```
bdero: Occlusion and emissive sit last in the priority list, so they are the first inputs dropped once a material has two or more advanced feature textures. A missing emissive map is far more visible than a missing clearcoat-roughness map. Worth ordering by visual impact until the permutation TODO lands, since CarConcept, ToyCar, and StainedGlassLamp in the new catalog all hit this.

packages/flutter_scene/lib/src/material/physical_material_variant.dart
```diff
@@ -0,0 +1,395 @@
...
+      ..setVec2('clearcoat_normal_scale', d.clearcoatNormalScale)
+      ..setVec3('sheen_color', d.sheenColor.xyz)
+      ..setFloat('sheen_roughness', d.sheenRoughness)
+      ..setFloat('transmission', d.transmission)
+      ..setFloat('diffuse_transmission', d.diffuseTransmission)
+      ..setVec3('diffuse_transmission_color', d.diffuseTransmissionColor.xyz)
+      ..setFloat('thickness', d.thickness)
+      ..setVec3('attenuation_color', d.attenuationColor.xyz)
+      ..setFloat(
+        'attenuation_distance',
```
bdero, line 196: This passes the descriptor value straight through, but the `attenuationDistance` setter clamps with `math.max(value, 0.0001)`. A descriptor with `attenuationDistance: 0` reaches the shader as zero. Same clamp here would keep the two paths in agreement.

packages/flutter_scene/lib/src/material/advanced_physical_material.dart
```diff
@@ -0,0 +1,395 @@
...
+
+final class _PhysicalAssets {
+  const _PhysicalAssets(this.library, this.metadata);
+
+  final gpu.ShaderLibrary library;
+  final Map<String, Object?> metadata;
+}
+
+Future<_PhysicalAssets> _loadPhysicalAssets() =>
+    _physicalAssetsFuture ??= _loadPhysicalAssetsUncached();
```
bdero: A failed load is cached permanently, so one transient failure poisons every later `fromDescriptor` call for the life of the isolate. Clearing the field in a catch before rethrowing would let it retry.

packages/flutter_scene/lib/src/material/advanced_physical_material.dart
```diff
@@ -0,0 +1,395 @@
...
+  /// Updates the linear dielectric specular color.
+  set specularColor(Vector3 value) =>
+      parameters.setVec3('specular_color_factor', value);
+
+  /// Updates clearcoat perceptual roughness.
+  set clearcoatRoughness(double value) =>
+      parameters.setFloat('clearcoat_roughness', value.clamp(0.0, 1.0));
+
+  /// Updates the sheen color.
+  set sheenColor(Color value) =>
```
bdero: This is the only color setter taking `Color`; `emissive`, `specularColor`, `diffuseTransmissionColor`, and `attenuationColor` all take `Vector3`, and the descriptor uses `Vector4`. `Color.r/g/b` are also not linear, while the doc comment says linear. `Vector3` would match the rest.

packages/flutter_scene/lib/src/render/scene_pass.dart
```diff
@@ -343,47 +358,100 @@ class ScenePass extends RenderGraphPass {
...
-      loadAction: _enableMsaa ? gpu.LoadAction.load : gpu.LoadAction.dontCare,
-    );
-    if (_enableMsaa) {
-      translucentAttachment.resolveTexture = hdrColor;
-      translucentAttachment.storeAction = gpu.StoreAction.multisampleResolve;
+    var currentColor = hdrColor;
+    var nextColor = alternateColor!;
+    gpu.Texture? transmissionAtlas;
+    List<gpu.Texture>? transmissionMips;
+    while (encoder.hasPendingTranslucent) {
```
bdero, line 383: A batch ends immediately before each scene-color reader, so N transmissive draws produce N iterations, each costing a fullscreen copy, a fresh render pass, and for filtered materials up to 8 downsample passes plus an atlas pass, every one its own command buffer. TransmissionOrderTest and CommercialRefrigerator are now in the catalog and will land in the hundreds of passes per frame. There is no cap and no TODO. Capping the batch count with a `log` of what was dropped, or skipping the re-filter when scene color has not changed since the last one, would bound it.

packages/flutter_scene/lib/src/render/scene_pass.dart
```diff
@@ -343,47 +358,100 @@ class ScenePass extends RenderGraphPass {
...
+            currentColor,
+            transmissionAtlas,
+            transmissionMips,
+          );
+          lighting.filteredSceneColor = transmissionAtlas;
+          lighting.transmissionFilterBandCount = bands;
+        }
+      }
+
+      final attachment = gpu.ColorAttachment(
```
bdero, line 437: The old split kept multisample contents across the split via `storeAndMultisampleResolve` and skipped `_encodeCopy` entirely on the MSAA path. Every batch now replays through `_encodeCopy` into a fresh transient target, so the pre-existing single-refraction case pays a fullscreen MSAA copy it did not before. Intentional?

packages/flutter_scene/lib/src/render/scene_pass.dart
```diff
@@ -343,47 +358,100 @@ class ScenePass extends RenderGraphPass {
...
-        depthLoadAction: gpu.LoadAction.load,
-        depthClearValue: 1.0,
-      ),
-    );
-    final translucentCommands = gpu.gpuContext.createCommandBuffer();
-    final translucentPass = translucentCommands.createRenderPass(
-      translucentTarget,
+
+    context.blackboard.set(
+      kOpaqueSceneColorBlackboardKey,
```
bdero: After the loop, `lighting.opaqueSceneColor` still holds the source for the final batch while `currentColor` holds the fully composed result, so this publishes the image without the last screen-reading layer. The `?? currentColor` fallback reads like the intent was the latest snapshot.

packages/flutter_scene/lib/src/fmat/target_shader_bundle.dart
```diff
@@ -0,0 +1,162 @@
...
+  Set<ShaderBundleBackend> backends,
+) => fb.ShaderObjectBuilder(
+  name: shader.name,
+  metalIos: backends.contains(ShaderBundleBackend.metalIos)
+      ? _copyBackend(shader.metalIos)
+      : null,
+  metalDesktop: backends.contains(ShaderBundleBackend.metalDesktop)
+      ? _copyBackend(shader.metalDesktop)
+      : null,
+  openglEs: backends.contains(ShaderBundleBackend.openglEs)
```
bdero, line 100: `openglDesktop` has no `ShaderBundleBackend` value and is never copied here, so no backend set can preserve it. `target_shader_bundle_test.dart` asserts it is null after a trim that did not exclude it, which bakes the behavior in without explaining it. A comment or `TODO(shader-bundle)` saying desktop GL is deliberately unsupported would make it intentional rather than incidental.

packages/flutter_scene/lib/src/importer/src/gltf/types.dart
```diff
@@ -269,6 +280,151 @@ class GltfMaterial {
...
+  final double dispersion;
+  final double emissiveStrength;
+  final double ior;
+  final GltfMaterialIridescence? iridescence;
+  final GltfMaterialSheen? sheen;
+  final GltfMaterialSpecular? specular;
+  final GltfMaterialTransmission? transmission;
+  final GltfMaterialVolume? volume;
+
+  bool get usesPhysicalExtensions =>
```
bdero: Unused. Only `requiresPhysicalMaterial` is called anywhere.

packages/flutter_scene/lib/src/material/preprocessed_material.dart
```diff
@@ -164,6 +196,7 @@ class PreprocessedMaterial extends Material implements HotReloadableFmat {
         fragmentShader,
         lighting,
         env,
+        bindSsao: !_sceneInputs.contains(RenderInput.filteredSceneColor),
```
bdero, line 215: Transmissive materials silently lose SSAO here, and the matching `#ifndef FLUTTER_SCENE_SKIP_SSAO` in the shader. Reasonable tradeoff for sampler pressure, but it needs a `TODO(topic)` naming what is missing and how to finish it.

packages/flutter_scene/lib/src/material/preprocessed_material.dart
```diff
@@ -203,7 +236,60 @@ class PreprocessedMaterial extends Material implements HotReloadableFmat {
...
+      shader.getUniformSlot('mask_texture'),
+      Material.whitePlaceholder(resolveTextureSource(source)),
+      sampler: textureSourceSampler(source) ?? _depthMaskSampler,
+    );
+  }
+
+  @override
+  @internal
+  gpu.CullMode get renderCullMode {
+    if (_culling == FmatCulling.back && doubleSided && isOpaque()) {
```
bdero, line 300: Gating on `isOpaque()` means a double-sided translucent or transmissive material still culls back faces, which contradicts glTF, where `doubleSided` is independent of alpha mode. If this is deliberate to avoid sorting artifacts on translucent back faces, it needs a comment saying so.

packages/flutter_scene/shaders/material_lighting.glsl
```diff
@@ -490,17 +655,49 @@ vec4 EvaluateLighting(MaterialInputs material) {
...
+          environment_transform * coat_reflection, coat_roughness);
+      coat_prefiltered = mix(coat_prefiltered, coat_prefiltered_b, env_blend);
+    }
+    vec2 coat_ab = texture(
+        brdf_lut, vec2(coat_n_dot_v, min(coat_roughness, 0.99))).rg;
+    coat_ibl = coat_prefiltered *
+               (vec3(0.04) * coat_ab.x + coat_ab.y) *
+               frag_info.environment_intensity;
+  }
+  ambient += material.sheen_color * irradiance *
```
bdero, line 698: The sheen term is added after `occlusion * ambient_shadow` has been applied to everything else in `ambient`, so sheen fabrics light up in shadow and inside AO cavities. GlamVelvetSofa and SheenChair should show it.

packages/flutter_scene/shaders/material_lighting.glsl
```diff
@@ -237,18 +237,72 @@ float FetchPunctualIndex(int j) {
...
                            vec3 camera_normal, vec3 albedo, float metallic,
                            float roughness, vec3 reflectance, float n_dot_v,
-                           float specular_scale) {
-  float n_dot_l = max(dot(normal, light_vector), 0.0);
+                           float specular_scale, vec3 anisotropic_tangent,
+                           vec3 anisotropic_bitangent) {
+  float signed_n_dot_l = dot(normal, light_vector);
+  float n_dot_l = max(signed_n_dot_l, 0.0);
+#ifdef FLUTTER_SCENE_PHYSICAL_MATERIAL
+  vec3 transmitted_diffuse = material.diffuse_transmission_color * albedo *
```
bdero: KHR_materials_diffuse_transmission defines the transmitted color as the factor alone, not factor times base color, so multiplying by `albedo` here (and again in the ambient term) may double-count. Worth checking against DiffuseTransmissionTest.

packages/flutter_scene/shaders/material_lighting.glsl
```diff
@@ -237,18 +237,72 @@ float FetchPunctualIndex(int j) {
   return texture(punctual_index, uv).r;
 }
 
+// Widens a specular lobe to cover unresolved normal variation in one pixel.
+float SpecularAARoughness(vec3 normal, float roughness) {
```
bdero, line 250: Two things came along with the extraction. The early return is now `max(roughness, kMinRoughness)` where the old inline code left roughness untouched when specular AA was off, so this is a small behavior change for every material. And the comment explaining that the branch is on a uniform, which is what keeps `dFdx` under uniform control flow, did not survive. That one is worth restoring.

packages/flutter_scene/lib/src/geometry/mesh_geometry.dart
```diff
@@ -929,6 +1041,12 @@ class GeometryBuilder {
     return this;
   }
 
+  /// Sets the secondary texture coordinate for subsequent vertices.
+  GeometryBuilder texCoord1(Vector2 value) {
```
bdero, line 1047: The builder gained `texCoord1` but no `tangent` setter, so builder-authored geometry cannot carry tangents even though `MeshGeometry` and `rebuild` now accept them. Fine if the derivative fallback is meant to cover it, but the asymmetry is surprising.

packages/flutter_scene/lib/src/runtime_importer/material_builder.dart
```diff
@@ -33,32 +45,170 @@ Material buildMaterial(GltfMaterial? gm, List<Texture2D> textures) {
...
+  );
+}
+
+PhysicalTexture _physicalTexture(
+  GltfTextureInfo? info,
+  List<Texture2D> textures,
+) => PhysicalTexture(
+  source: _resolveTexture(info, textures),
+  transform: _textureTransform(info),
+  texCoord: info?.transform?.texCoord ?? info?.texCoord ?? 0,
```
bdero: `texCoord` is unclamped here while `_textureTexCoord` just below clamps to 0..1. Harmless since `_setTransform` clamps later, but the two helpers should agree.

packages/flutter_scene/CHANGELOG.md
```diff
@@ -32,6 +36,10 @@
 * Reused large transient buffers across changing instance counts.
 * Translucent instanced meshes now sort as one batch, with instances ordered within it.
 * Fixed instanced meshes losing their lighting, shadow, and fog bindings on GLES, which drew them (and any mesh sharing their pipeline) black.
+* Fixed anisotropy strength being ignored by environment lighting and using an isotropic visibility term for analytic lights.
+* Fixed anisotropic materials exposing primitive triangles by importing and interpolating authored tangents.
+* Vertex payload layouts include UV1 and tangents under versioned names, while older `.fscene` layouts upgrade when loaded.
```
bdero: Two user-visible changes are missing. Translucent sorting moved from squared distance about the world-transform origin to view-axis depth about the geometry bounds center, which has its own new test but no entry. And this bullet covers the layout names without the cost: the payload grows 50% unskinned and 30% skinned for every mesh, whether or not it uses UV1 or tangents.

examples/flutter_app/lib/example_clearcoat.dart
```diff
@@ -0,0 +1,294 @@
...
+  @override
+  void dispose() {
+    _scene.removeAll();
+    super.dispose();
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    if (_error case final error?) {
+      return Center(child: Text('Could not load clearcoat example\n$error'));
```
bdero: `assets/clearcoat/` is a pubspec asset dir that is gitignored except the README, so a fresh clone gets a broken entry in the example picker. Naming `tool/fetch_clearcoat_assets.sh` in this error text would point people at the fix.

bdero (commented):
Review notes below. The legacy layout upgrade math checks out for all three payload shapes, and the transmission atlas band packing is gap-free with a half-texel inset per band, so there is no cross-band bleed.

Two things worth resolving before merge. The screen-reading translucent loop is unbounded, so a scene with many transmissive draws pays a fullscreen copy plus a filter chain per draw. And `android_gles` failed on the adb wedge without rendering any frames, so GLES is unverified for a change touching shaders, the vertex layout, and multi-pass rendering.

packages/flutter_scene/lib/src/material/advanced_physical_material.dart
bdero: Done in dca74ae.

packages/flutter_scene/lib/src/material/physical_material_variant.dart
bdero: Done in dca74ae.

packages/flutter_scene/lib/src/material/advanced_physical_material.dart
bdero: Done in dca74ae.

packages/flutter_scene/lib/src/material/advanced_physical_material.dart
bdero: Done in dca74ae.

packages/flutter_scene/lib/src/render/scene_pass.dart
bdero: Done in dca74ae.

packages/flutter_scene/lib/src/render/scene_pass.dart
bdero: Done in dca74ae.

packages/flutter_scene/lib/src/render/scene_pass.dart
bdero: Done in dca74ae.

packages/flutter_scene/lib/src/fmat/target_shader_bundle.dart
bdero: Done in dca74ae.

packages/flutter_scene/lib/src/importer/src/gltf/types.dart
bdero: Done in dca74ae.

packages/flutter_scene/lib/src/material/preprocessed_material.dart
bdero: Done in dca74ae.

packages/flutter_scene/lib/src/material/preprocessed_material.dart
bdero: Done in dca74ae.

packages/flutter_scene/shaders/material_lighting.glsl
bdero: Done in dca74ae.

packages/flutter_scene/shaders/material_lighting.glsl
bdero: Done in dca74ae.

packages/flutter_scene/shaders/material_lighting.glsl
bdero: Done in dca74ae.

packages/flutter_scene/lib/src/geometry/mesh_geometry.dart
bdero: Done in dca74ae.

packages/flutter_scene/lib/src/runtime_importer/material_builder.dart
bdero: Done in dca74ae.

packages/flutter_scene/CHANGELOG.md
bdero: Done in dca74ae.

examples/flutter_app/lib/example_clearcoat.dart
bdero: Done in dca74ae.

packages/flutter_scene/shaders/pbr.glsl
```diff
@@ -86,3 +86,82 @@ float GeometrySmith(vec3 normal, vec3 camera_normal, vec3 light_normal,
       GeometrySchlickGGX(max(dot(normal, light_normal), 0.0), roughness);
   return camera_ggx * light_ggx;
 }
+
+#ifdef FLUTTER_SCENE_PHYSICAL_MATERIAL
+// Estevez/Kulla Charlie sheen distribution and fitted visibility.
+float DistributionCharlie(float roughness, float n_dot_h) {
```
bdero, line 92: `inv_alpha` reaches 1e6 here, since the call site floors sheen roughness at 0.001 and `alpha` is roughness squared. On a mediump default-precision GLES fragment stage that is well past the fp16 ceiling, so `2.0 + inv_alpha` saturates to inf and `pow(sin2, inv_alpha * 0.5)` loses all precision. `_SheenLambda` compounds it with `exp()` on top. Given #234 was already an Impeller mediump sampler leak, this looks like the same class of bug waiting on a real mobile GPU. Flooring sheen roughness higher, or forcing `highp` on these, would bound it. Worth a Pixel pass specifically, since the SwiftShader lanes will not reproduce mediump behavior.

packages/flutter_scene/shaders/material_varyings.glsl
```diff
@@ -7,21 +7,41 @@ in vec3 v_position; // world-space position
...
 
 out vec4 frag_color;
 
 // World-space position of the fragment.
 vec3 GetWorldPosition() { return v_position; }
 
 // Normalized world-space geometric normal (before any normal-map perturbation).
-vec3 GetWorldNormal() { return normalize(v_normal); }
+// Back-facing fragments use the reversed normal required by two-sided lighting.
+vec3 GetWorldNormal() {
```
bdero, line 22: This flips the geometric normal for every lit material, not just physical ones, so it changes shading on existing double-sided meshes. Two things follow. `EvaluateAnalyticLight` already has a branch for facing the light through a double-sided surface, so it is worth confirming the two are not now compensating for each other. And there is no CHANGELOG entry for it, though it is more visible than several of the changes that do have one.

packages/flutter_scene/shaders/pbr.glsl
```diff
@@ -86,3 +86,82 @@ float GeometrySmith(vec3 normal, vec3 camera_normal, vec3 light_normal,
...
+                               float alpha_t, float alpha_b) {
+  float ggx_v = n_dot_l *
+      length(vec3(alpha_t * t_dot_v, alpha_b * b_dot_v, n_dot_v));
+  float ggx_l = n_dot_v *
+      length(vec3(alpha_t * t_dot_l, alpha_b * b_dot_l, n_dot_l));
+  return clamp(0.5 / max(ggx_v + ggx_l, 1e-5), 0.0, 1.0);
+}
+
+// Compact thin-film approximation. Thickness is nanometres.
+vec3 ThinFilmFresnel(vec3 base_f0, float film_ior, float thickness,
```
bdero, line 159: The doc comment names this an approximation, and the sheen ambient term and the anisotropic bent-normal heuristic are the same, but none of the three carries a `TODO(topic)` saying what is missing and how to finish it. Repo convention wants one for anything approximated, and these are the spots most likely to be revisited when the Compare* reference assets get graded.

packages/flutter_scene/shaders/material_lighting.glsl
```diff
@@ -393,11 +509,40 @@ vec4 EvaluateLighting(MaterialInputs material) {
...
+  float anisotropic_normal_length_squared =
+      dot(anisotropic_normal, anisotropic_normal);
+  if (anisotropic_normal_length_squared > 1e-10) {
+    anisotropic_normal *= inversesqrt(anisotropic_normal_length_squared);
+  } else {
+    anisotropic_normal = normal;
+  }
+  float bend = 1.0 - anisotropy * (1.0 - roughness);
+  bend *= bend;
+  bend *= bend;
```
bdero, line 554: The `bend` curve (squared twice, then mixed against roughness squared) is an unexplained magic falloff. A one-line comment naming the source or the intent would save the next reader deriving it, especially since it feeds the reflection vector for every anisotropic material.

packages/flutter_scene/shaders/flutter_scene_linear_depth_normal.frag
```diff
@@ -52,13 +56,20 @@ void main() {
 
   // Rotate the interpolated world normal into view space (eye looking down
   // +Z), the space the reflection trace works in.
-  vec3 n = normalize(v_normal);
+  vec3 n = GetWorldNormal();
   vec3 view_normal = normalize(vec3(dot(n, info.camera_right.xyz),
                                     dot(n, info.camera_up.xyz),
                                     dot(n, info.camera_forward.xyz)));
 
+  vec2 roughness_uv = GetUV(int(info.roughness_uv_rotation.z));
```
bdero: This is the fourth copy of the scale/rotate/offset UV math, after `TransformUv` in the standard frag, the inline version in the unlit frag, and `PhysicalUv` in the generated fmat. The three also disagree on how the UV channel is encoded and decoded: `rotation.z > 0.5` in unlit, `int(rotation.z + 0.5)` in the standard frag, and bare `int(...)` here, which truncates rather than rounds. Since these are all fed by the same `_packTextureTransform`, one shared helper in `material_inputs.glsl` would remove both the duplication and the rounding discrepancy.

packages/flutter_scene/lib/src/fmat/fmat_emitter.dart
```diff
@@ -143,6 +149,120 @@ String emitFragmentGlsl(FmatMaterial material) {
     sb.writeln('}');
     sb.writeln();
   }
+  if (material.engineInputs.contains('filtered_scene_color')) {
+    sb.writeln('// Compact roughness-filtered scene-color atlas.');
+    sb.writeln('uniform sampler2D scene_filtered_color;');
```
bdero: This emits roughly 40 lines of GLSL as Dart string literals, including the whole bicubic band sampler. It is the largest shader body in the repo that gets no syntax highlighting, no formatting, and no way to read it without building. `pbr.glsl` already gates its advanced lobes behind `#ifdef FLUTTER_SCENE_PHYSICAL_MATERIAL`; the same treatment with a `filtered_scene_color.glsl` include gated on the existing define would make this reviewable and leave the emitter deciding only whether to include it.

bdero (commented):
Second pass, covering the shader and importer files I skimmed the first time. Checked and clean: both vertex bodies flip `tangent.w` on a negative-determinant model transform, so the scene-root Z flip does not invert bitangents (NormalTangentMirrorTest should hold); the offline emitter writes the new versioned layout names on both the skinned and unskinned paths, so freshly cooked documents do not fall down the legacy upgrade path; `_vec4` is length-defensive against short extension vectors; and the `MaskInfo` packing matches between `PhysicallyBasedMaterial` and `PreprocessedMaterial`.

The one I would treat as a likely bug is the Charlie sheen distribution at low roughness on GLES.

packages/flutter_scene/shaders/pbr.glsl
bdero: Floored Charlie sheen at kMinRoughness, which bounds inverse alpha to about 494 in mediump. The GLES smoke lane passes in 97cf741, and a physical Pixel was not available locally.

packages/flutter_scene/shaders/material_varyings.glsl
bdero: Confirmed the half-vector guard now only handles degeneracy from strongly perturbed normals after back-side light exits. Added the behavior to the changelog in 97cf741.

packages/flutter_scene/shaders/pbr.glsl
bdero: Added completion TODOs for thin-film spectral integration, sheen IBL, and anisotropic IBL in 97cf741.

packages/flutter_scene/shaders/material_lighting.glsl
bdero: Documented the bend curve's intent and the path to replace it with anisotropic prefiltering in 97cf741.

packages/flutter_scene/shaders/flutter_scene_linear_depth_normal.frag
bdero: Unified packed UV selection and transform math in material_inputs.glsl, including physical materials and depth passes, in 97cf741.

packages/flutter_scene/lib/src/fmat/fmat_emitter.dart
bdero: Moved the filtered scene-color sampler into filtered_scene_color.glsl and left the emitter with one conditional include in 97cf741.

packages/flutter_scene/lib/src/fmat/fmat_emitter.dart
```diff
@@ -143,6 +149,10 @@ String emitFragmentGlsl(FmatMaterial material) {
     sb.writeln('}');
     sb.writeln();
   }
+  if (material.engineInputs.contains('filtered_scene_color')) {
```
bdero, line 161: Now that this is an include, `filtered_scene_color.glsl` references `scene_opaque_color` and `GetSceneColor`'s neighbours, which the branch above only emits for `scene_color`. Both stock materials declare the pair, so nothing is broken today, but a `.fmat` declaring `filtered_scene_color` alone still passes the parser's name check and then fails in impellerc with an undeclared-identifier error pointing into an engine include. Emitting `scene_color` implicitly, or rejecting the combination in the parser next to the existing `Unknown engine_inputs entry` message, would keep the failure legible.

packages/flutter_scene/lib/src/render/scene_pass.dart
```diff
@@ -343,47 +370,122 @@ class ScenePass extends RenderGraphPass {
...
+      rendererSubmissions.submit(translucentCommands);
+
+      final completedColor = nextColor;
+      nextColor = currentColor;
+      currentColor = completedColor;
+      snapshotPass++;
     }
-    encoder.flushTranslucent(translucentPass: translucentPass);
+
+    context.blackboard.set(kOpaqueSceneColorBlackboardKey, currentColor);
```
bdero: With both keys now set to `currentColor`, `kOpaqueSceneColorBlackboardKey` is the same texture as `kSceneColorBlackboardKey`, and grepping the package finds no reader for it at all. Worth either deleting it or giving the doc comment a line saying it deliberately aliases the final composed color for external graph passes, since the name still reads as opaque-only.

bdero (commented):
Re-reviewed both follow-up commits. Everything from the two earlier passes is addressed, and the parts I checked in detail hold up: the MSAA path now accumulates in a `devicePrivate` multisample target with `loadAction: load` and alternating resolves, so it drops the fullscreen copy and still hands each layer a correct snapshot of what is behind it; `kMinRoughness` at 0.045 caps the Charlie reciprocal exponent near 494, comfortably inside mediump; and the five new regression tests pin the behavior rather than just the fix.

Two minor things below. Neither blocks.

On CI, `android_gles` failed again, and again with no crash signature. The app logged `pbr_cuboid renders a sane frame` and then logcat capture itself stopped four seconds later, which is the adb wedge rather than an app fault (a crash would leave a tombstone and keep logging). The two failures also died at different points, so it does not look deterministic, and the linux llvmpipe and web WebGL2 lanes both pass on the same GLSL. Still, that lane has now rendered exactly one scene across two attempts, so there is effectively no GLES image coverage on a change this size.

packages/flutter_scene/lib/src/fmat/fmat_emitter.dart
bdero: The parser already rejects this combination with a clear filtered_scene_color requires scene_color error, and fmat_test pins it. Clarified the dependency in the FmatMaterial docs in 1b4cc30.

packages/flutter_scene/lib/src/render/scene_pass.dart
bdero: Removed the unused key and its duplicate publication in 1b4cc30.

packages/flutter_scene/lib/src/fmat/build_materials.dart
```diff
@@ -300,10 +336,19 @@ Future<void> buildMaterials({
         );
       }
 
+      final hasShadowVariant =
+          generateShadowVariants &&
+          compiled.material.shadingModel != FmatShadingModel.unlit;
       final fragFileName = '$entryName.frag';
+      final fragmentGlsl = hasShadowVariant
+          ? compiled.glsl.replaceFirst(
```
bdero: `replaceFirst('\n', ...)` lands the define after whatever `emitFragmentGlsl` writes first, currently the "Generated from a .fmat material" banner. It works today, but the emitter already emits `FLUTTER_SCENE_PHYSICAL_MATERIAL` and `FLUTTER_SCENE_SKIP_SSAO` from the inside. A `defines` parameter would put all three on the same footing and would not silently misplace this one if that header line ever moves.

packages/flutter_scene/lib/src/fmat/build_materials.dart
```diff
@@ -312,6 +357,17 @@ Future<void> buildMaterials({
         // there, not relative to the manifest.
         'file': 'build/fmat/$bundleName/$fragFileName',
       };
+      if (hasShadowVariant) {
+        final shadowEntryName = '${entryName}Shadow';
```
bdero: The unsuffixed entry is the reduced shader and `...Shadow` is the complete one, which reads backwards; `PhysicalOpaque` looks like the full material. Naming the reduced one `...NoShadow`, or a line of comment here, would stop the next reader reaching for the wrong entry.

packages/flutter_scene/shaders/material_lighting.glsl
```diff
@@ -74,6 +73,34 @@ vec3 BiasDirectionalShadowPosition(vec3 world_pos, vec3 n) {
   return world_pos + n * normal_offset;
 }
 
+vec2 PoissonShadowTap(int i) {
```
bdero, line 76: The unrolled block this replaces carried a `TODO(flutter_scene)` noting that it would naturally loop over file-scope const arrays, but impellerc/SPIRV-Cross emits invalid ES 1.00 array constructors. These if-chains are a workaround for that same limitation and nothing records why, so it now reads as a deliberate style choice. Worth restoring the TODO.

packages/flutter_scene/shaders/material_lighting.glsl
```diff
@@ -108,37 +135,15 @@ float SampleCascade(int cascade, int count, mat4 cascade_matrix, float box,
...
-  lit += _SHADOW_TAP(-0.24188840, 0.99706507, 0.0, 0.5);
-  lit += _SHADOW_TAP(-0.81409955, 0.91437590, 0.5, 0.5);
-  lit += _SHADOW_TAP(0.19984126, 0.78641367, -1.0, 1.0);
-  lit += _SHADOW_TAP(0.14383161, -0.14100790, 0.0, 1.0);
-#undef _SHADOW_TAP
-  if (fixed_filter > 0.5) {
-    lit += ShadowTap(vec2(1.0), 1.0, 0.0, radius, uv, cascade, inv_count,
+  for (int i = 0; i < 17; i++) {
+    if (i >= sample_count) break;
+    vec2 tap = mix(PoissonShadowTap(i), FixedShadowTap(i), fixed_filter);
```
bdero, line 145: The old macro mixed two compile-time literals. This calls a 16-branch and a 5-branch chain on every one of seventeen iterations, and whether that collapses back depends on the driver unrolling the constant-bound loop and constant-folding both calls. Since reducing GLES expansion was the goal and this is the hottest fragment path, worth confirming the compiled instruction count actually dropped rather than just the source length.

packages/flutter_scene/test/physical_material_sources_test.dart
```diff
@@ -0,0 +1,248 @@
...
+
+  test('advanced materials build and select shadow shader variants', () {
+    final builder = File(
+      'lib/src/fmat/build_materials.dart',
+    ).readAsStringSync();
+    final material = File(
+      'lib/src/material/advanced_physical_material.dart',
+    ).readAsStringSync();
+
+    expect(builder, contains('generateShadowVariants: true'));
```
bdero: These assert on source text, so they break on any rename or reformat and prove nothing about what the build produces. Compare `scene color snapshot passes are bounded` from the previous round, which calls the real helper. Compiling both entries and asserting the base one carries no `shadow_map` declaration while the `Shadow` one does would be sturdier and would actually test the split.

.gitignore
```diff
@@ -35,3 +35,8 @@ pubspec_overrides.yaml
 # Fetched splat capture for the Gaussian Splats example (see
 # examples/flutter_app/tool/fetch_splat_asset.sh).
 examples/assets_src/*.splat
+
+# Fetched material-demo textures (see
```
bdero: Stale now that the clearcoat example is gone. These lines ignore `examples/flutter_app/assets/clearcoat/*` and un-ignore a README that no longer exists, pointing at a fetch script that was deleted. It is the last reference to the removed example anywhere in the tree.

bdero (commented):
Reviewed all ten commits since the last pass. Nothing here blocks, and CI is fully green for the first time, `android_gles` included.

Checked in detail and correct: the PCF rewrite is behaviorally identical, including the seventeenth tap, since `angle` is scaled by `(1.0 - fixed_filter)` so `ca`/`sa` are exactly `(1, 0)` whenever the fixed grid is selected, and `FixedShadowTap` reproduces all seventeen old literals. `filtered_scene_color requires scene_color` is now enforced in the parser rather than only documented. `kOpaqueSceneColorBlackboardKey` is deleted outright. The clearcoat example is gone with no dangling references. The ANGLE and software-GPU workflow experiments net to zero, so no `.github/` change ships. And `_lighting` is the same instance at both pipeline resolution and `bind`, so the new shadow-variant selection cannot desync a pipeline from its bindings.

One question rather than a finding. `argos/android_gles` reports 10 changed against 1 changed on every other lane. I would expect exactly that if the GLES baseline predates this PR, since the lane never produced images until now, but it is worth confirming that is the explanation rather than GLES diverging from the other backends.

packages/flutter_scene/test/physical_material_sources_test.dart
bdero: Done in 242fc33. The test now compiles both generated GLES entries with impellerc and checks the reflected shadow_map resources.

packages/flutter_scene/shaders/material_lighting.glsl
bdero: Confirmed with impellerc. GLES SPIR-V dropped from 8,882 to 7,358 instructions, and image-sampling operations dropped from 152 to 88.

packages/flutter_scene/shaders/material_lighting.glsl
bdero: Done in 242fc33. The const-array limitation and completion condition are recorded in the restored TODO.

packages/flutter_scene/lib/src/fmat/build_materials.dart
bdero: Done in 242fc33. A comment now documents the unsuffixed no-shadow fast path and the full Shadow entry.

packages/flutter_scene/lib/src/fmat/build_materials.dart
bdero: Done in 242fc33. The emitter now accepts defines directly, so the shadow define is emitted alongside the material defines.

bdero:
Confirmed the GLES baseline is build 1770 on master from two days ago, paired with Vulkan build 1769, so it is not older than the other baseline. The additional GLES diffs are the expected GLES-only output corrections from this branch's shader/cache fixes, and they were approved.

.gitignore
bdero: Done in 242fc33. Removed the stale clearcoat asset ignore entries.

packages/flutter_scene/CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 * `ResourceGroup.track` scopes a load to the group's lifetime, so `dispose()` releases the claims it took. Previously `dispose()` only released the progress notifier despite the name.
 * Fixed `ResourceGroup` throwing when it was disposed while a tracked load was still in flight, which is what tearing down a level mid-load does.
 
+* `PhysicalMaterial` and glTF import support every ratified `KHR_materials_*` extension with scene-native property names.
```
bdero: `PhysicalMaterial` is no longer a type a user can reference; it is `@internal` and off the barrel now that everything moved onto `PhysicallyBasedMaterial`. Someone reading the changelog will go looking for a class that is not there, so this should name `PhysicallyBasedMaterial` instead.

packages/flutter_scene/lib/src/material/material.dart
```diff
@@ -161,14 +161,13 @@ abstract class Material {
...
-  /// split-sum specular IBL (see [buildBrdfLutTexture]).
+  /// Builds the BRDF lookup texture and loads the physical shader variants.
   ///
   /// Called by the [Scene] constructor; rendering is gated on the returned
   /// [Future] completing. The texture is built once and reused.
-  static Future<void> initializeStaticResources() {
+  static Future<void> initializeStaticResources() async {
     _brdfLutTexture ??= buildBrdfLutTexture();
-    return Future<void>.value();
+    await PhysicallyBasedMaterial.initializeStaticResources();
```
bdero, line 170: This makes every scene load the physical shader bundle before its first frame, since rendering is gated on `initializeStaticResources()` completing. An app that only ever uses `UnlitMaterial` now pays that load and its resident memory. It is the price of making the feature setters synchronous, which seems like the right trade, but it is a real startup cost change and is worth a changelog line so it is not discovered as a regression.

packages/flutter_scene/lib/src/material/physically_based_material.dart
```diff
@@ -144,11 +348,990 @@ class PhysicallyBasedMaterial extends Material {
...
+    if (_variantDirty) {
+      final previousTransmission = _usesTransmissionVariant;
+      _variantKey = _computeVariantKey();
+      _variantDirty = false;
+      if (previousTransmission != _usesTransmissionVariant) {
+        markMaterialSceneInputsChanged();
+      }
+    }
+    if (!_materialDataDirty) return;
+    _preparedVariant = _usesPhysicalVariant
```
bdero: Every setter marks `_materialDataDirty`, so this reconstructs the whole `PhysicalMaterialVariant` (roughly thirty parameter writes plus the texture and transform binds) whenever any property changed, and it runs inside `bind()`. Animating a single scalar on an advanced material therefore rebuilds the variant every frame in the render path. Splitting the flag into parameters-changed and variant-identity-changed, or updating the existing variant's parameters in place, would keep the frame-build win from the bind-memo work.

packages/flutter_scene/lib/src/material/physically_based_material.dart
```diff
@@ -144,11 +348,990 @@ class PhysicallyBasedMaterial extends Material {
...
+  bool get _usesTransmissionVariant =>
+      (_variantKey & _transmissionFeature) != 0;
+
+  void _ensurePreparedVariant() {
+    if (_variantDirty) {
+      final previousTransmission = _usesTransmissionVariant;
+      _variantKey = _computeVariantKey();
+      _variantDirty = false;
+      if (previousTransmission != _usesTransmissionVariant) {
+        markMaterialSceneInputsChanged();
```
bdero: This bump lands one frame late. `Scene` snapshots `_cachedMaterialInputs` and decides `captureOpaqueColor` before encoding begins, so a revision changed during `bind()` is not observed until the next frame. The first frame after a material crosses into or out of the transmission variant renders with no scene-color capture behind it. Recomputing the variant key when the property is set rather than at bind time would close the gap.

bdero (commented):
All six items from the last round are closed, two of them better than proposed. The source-text assertions are now a real test that runs impellerc over both variants and checks the reflection JSON for the sampler, which is stronger than the check I suggested, and the define splice became `emitFragmentShaderVariants` so all three defines are emitted the same way.

The unification reads well. One public material type with plain setters, no async descriptor step, and the smoke scenes get noticeably simpler for it. Four notes below, all on that change rather than on the fixes.

packages/flutter_scene/CHANGELOG.md
bdero: Done in c6cd1de. The changelog now names PhysicallyBasedMaterial.

packages/flutter_scene/lib/src/material/physically_based_material.dart
bdero: Done in c6cd1de. The prepared physical variant is now updated in place and is rebuilt only when transmission changes the render layout.

packages/flutter_scene/lib/src/material/physically_based_material.dart
bdero: Done in c6cd1de. Variant identity and the scene-input revision now update in the setter, with a regression test for both transition directions.

packages/flutter_scene/lib/src/material/material.dart
bdero: Done in c6cd1de. The startup shader-bundle cost is now called out in the changelog.

PR #302, opened 2026-08-07
Add project-first workflow, managed toolchains, and run sessions to the editor

bdero:
The editor becomes project-first. A committed `.fproject` beside a Flutter project's pubspec carries build configurations, free-form tasks, and a default scene; scenes opened on their own join the nearest ancestor project, and the start screen, File menu, asset browser, and toolbar all operate in project scope with per-user selections (configuration, device, last scene) kept in editor settings.

Play launches an editor-managed `flutter run --machine` session with hot reload/restart, stop, structured console progress, and an optional hot-restart-on-scene-save toggle. Referenced `.fmat` sources compile on demand and hot swap as edited, routed through a registry of Flutter installations that includes editor-managed SDK checkouts created at the editor's exact build commit. Also adds distributable editor packaging and MCP tools covering the project, device, document, and session lifecycle.


packages/flutter_scene/lib/src/fscene/realize/component_codec.dart
```diff
@@ -129,29 +138,71 @@ abstract class ComponentCodec {
...
+  /// Serializes [component], dispatching by its runtime type and falling
+  /// back to the first codec that claims it. Returns null when none does.
   ComponentSpec? serialize(Component component, SerializeContext context) {
+    final spec =
+        _byComponentType[component.runtimeType]?.serialize(
+          component,
+          context,
+        ) ??
+        _serializeByScan(component, context);
+    if (spec != null && !component.enabled) {
```
bdero: serialize only adds `enabled: false` for disabled components and never strips a stale one from specs a codec returns verbatim. `PlaceholderComponentCodec.serialize` and the deferred-splat branch in render_extras_codecs.dart return the retained spec, so re-enabling a foreign or still-loading component saves the old `enabled: false` and reverts on reload.

packages/flutter_scene/lib/src/fscene/realize/builtin_codecs.dart
```diff
@@ -51,124 +62,91 @@ void registerBuiltinComponentCodecs(FsceneComponentRegistry registry) {
...
-  @override
-  Component realize(ComponentSpec spec, RealizeContext context) {
-    final p = spec.properties;
-    final envProp = p['environment'];
-    final envId = envProp is ResourceRefValue ? envProp.id : null;
+  EnvironmentVolumeComponent create(PropertyReader props) {
+    final envId = props.resourceId('environment');
     final settings =
-        (envId == null ? null : context.resources?.environment(envId)) ??
+        (envId == null ? null : props.context.resources?.environment(envId)) ??
```
bdero: When this lookup fails (no resource realizer, preload incomplete, or a dangling id), create falls back to an untagged `EnvironmentSettings`, the get reads null, and the next save omits `environment`, deleting the authored reference. The old codec stamped the id unconditionally and round-tripped it regardless. Same class of loss in particle_emitter_codec.dart, where the texture write returns early when `resources` is null, so an emitter realized without a realizer loses its authored texture ref on save.

packages/flutter_scene_editor/lib/src/project/vm_service_link.dart
```diff
@@ -0,0 +1,146 @@
...
+      return;
+    }
+    if (decoded is! Map) return;
+    final id = decoded['id'];
+    final key = id is int ? id : (id is String ? int.tryParse(id) : null);
+    if (key == null) return;
+    _pending.remove(key)?.complete(decoded.cast<String, Object?>());
+  }
+
+  void _failAll(String reason) {
```
bdero, line 84: `_failAll` only fails currently-pending requests and nothing marks the link dead, so AppSession's `_vmLink ??=` keeps the cached link after the socket drops. The next `reloadScenes()` registers a completer nothing will ever complete, the await hangs, and the hot-restart fallback in `_onSceneSaved` is never reached, so save-to-refresh silently stops working for the rest of the session.

packages/flutter_scene/lib/src/fscene/source/source_scene_loader_io.dart
```diff
@@ -0,0 +1,343 @@
...
+  /// Whether [key] is a project-relative document key this loader reads.
+  bool isSourceKey(String key) =>
+      (key.endsWith('.fscene') || key.endsWith('.fsceneb')) &&
+      _fileFor(key).existsSync();
+
+  /// When [error] is a filesystem failure (a macOS App Sandbox denial being
+  /// the common one; stat can succeed where open is refused), deactivates
+  /// source loading for the rest of the run so every load falls back to the
+  /// bundled DataAssets, and returns true. Anything else returns false.
+  bool deactivateOnAccessError(Object error, String key) {
```
bdero, line 97: Every `FileSystemException` is treated as the sandbox case and disables source loading for the rest of the run. A transient error (a sidecar mid-rewrite during a save, a file lock, a file deleted between existsSync and read) permanently degrades play mode to full hot restarts with only a debugPrint. It also strands templates already cached under sourceKeys, since releaseScene then resolves the DataAsset key instead and the sourceKey claims leak.

packages/flutter_scene_editor/lib/src/project/app_session.dart
```diff
@@ -0,0 +1,441 @@
...
+
+  /// Launches `flutter run --machine` for [configuration] on [device].
+  /// Returns false (with a console line) when a session is already active.
+  Future<bool> launch({
+    required FlutterInstallation installation,
+    required FProject project,
+    required BuildConfiguration configuration,
+    required FlutterDevice device,
+  }) async {
+    if (active) {
```
bdero, line 101: This guard runs before any await, but `_process` and the launching state are only set after `_startProcess0` completes, so two overlapping `launch()` calls (double-click Play, or UI plus MCP `run_project`) both pass it and spawn two processes. The second overwrites `_process`, and the first's `_onExited` resets to idle, orphaning the second app where Stop can never reach it. Related, `stop()` ignores an error reply to `app.stop` and can stick in `stopping` with the Stop button disabled.

bdero (commented):
Review pass over the branch. Ten inline notes, clustered in the save/round-trip path of the declarative codec migration and the run-session lifecycle. Smaller items not worth inline anchors, roughly in priority order.

- The Play button's `try { unawaited(_startPlaySession()) } on FormatException` catch is dead (the async throw lands in the discarded future), so launch-blocked errors give no console feedback.
- project_version_check.dart falsely warns "no flutter_scene dependency" for git/hosted dependency forms when no lockfile exists, and its lockfile regex never matches CRLF files.
- `FsceneComponentRegistry.register` never evicts the replaced codec's `_byComponentType` entry, so serialize can dispatch to the old codec after an override while realize uses the new one.
- The splat codec's strict `is DoubleValue`/`is IntValue` checks silently drop JSON-integer opacity/splatScale, and the fmod min/max distance writes have the same strictness.
- The scene-registry hot-reload closure re-inserts templates into `_sceneTemplates` without a holder claim, so a template released mid-reload is re-cached unevictably.
- `ComponentCodec.serialize` going abstract and the `ComponentPropertyDef` reshape are breaking on the public fscene.dart surface with no CHANGELOG entry staged, and the delta-vs-default output plus nested audio `attenuation` keys are invisible to published 0.20.x readers while `currentFsceneVersion` stays 2.
- HookLogTailer runs a synchronous recursive listSync every 500ms on the UI isolate for the whole Play session; stopping it at `app.started` would end it with the build. `ProjectRunner._line` also notifies per console line with an O(n) trim, and the declarative `fields` getters rebuild every descriptor per realize/serialize while `propertySchema` is already memoized.
- physics_codecs.dart and fmod_event_codec.dart re-implement the shared property_read.dart readers with diverging int coercion, and device_catalog.dart's env scrub misses `IMPELLERC` that `projectChildEnvironment()` removes.
- `registerPhysicsBackend`'s docstring describes registration no package can perform (it is not exported from any barrel, unlike `registerAudioEngineBackend`), property_constraints.dart line 91 references another engine in the published scene package, and the new public symbols in annotations.dart, declarative_codec.dart, and placeholder_codec.dart are missing `{@category}` directives.

packages/flutter_scene/lib/src/fscene/realize/builtin_codecs.dart
```diff
@@ -726,543 +728,439 @@ class MeshCodec extends ComponentCodec {
...
-    ComponentPropertyDef(
+    ComponentField.number(
       'innerConeAngle',
-      ComponentPropertyKind.number,
-      const DoubleValue(0.0),
+      defaultValue: 0.0,
       doc: 'Half-angle (radians) of the full-brightness inner cone.',
-      min: 0,
-      read: (c) => DoubleValue((c as SpotLightComponent).light.innerConeAngle),
+      constraints: const [Range(0, 1.5533430342749532), AngleRadians()],
```
bdero, line 1116: The glTF importer legally emits cone angles up to pi/2 (~1.5708) and passes them through verbatim, but this Range caps at 89 degrees, the declarative write path clamps at realize, and serialize reads back the live value, so a re-save persists the clamped 1.5533 in place of the authored value. The old realize read values verbatim, so this is a new one-way mutation of legal documents. `weight` Range(0,1) and `shadowCascadeCount` IntRange(1,4) have the same behavior.

packages/flutter_scene_editor/lib/src/project/fproject.dart
```diff
@@ -0,0 +1,524 @@
...
+      return (
+        fprojectPath: candidates.contains(preferred)
+            ? preferred
+            : candidates.first,
+        pubspecDirectory: null,
+      );
+    }
+    if (pubspecDirectory == null &&
+        entries.whereType<File>().any(
+          (f) => f.path.endsWith('/pubspec.yaml'),
```
bdero: listSync yields backslash-separated paths on Windows, so this never matches and the create-project flow is never offered there. Same family, the `startsWith('$root/')` checks in main.dart's `_recordLastScene`/`_openProjectScene` break per-project last-scene resume, and `resolvedDefaultScene` treats `C:\...` as relative because it only checks `startsWith('/')`. `resolveWorkingDirectory` already handles the drive-letter case, so these sites just missed it.

packages/scene/lib/src/schema/component_schema.dart
```diff
@@ -0,0 +1,333 @@
...
+
+/// Decodes a schema list, skipping malformed entries.
+List<ComponentSchema> decodeComponentSchemas(Object? json) {
+  final schemas = <ComponentSchema>[];
+  if (json is! List) return schemas;
+  for (final entry in json) {
+    if (entry is! Map) continue;
+    try {
+      schemas.add(ComponentSchema.fromJson(entry.cast<String, Object?>()));
+    } on FormatException {
```
bdero, line 342: `fromJson`'s unguarded casts (`json['doc'] as String?`, `entry.key as String`) throw `TypeError`, which this clause does not catch, so one malformed entry aborts decoding of every schema instead of being skipped. The lazy `(json['options'] as List?)?.cast<String>()` can also defer its element TypeError to inspector render time.

apps/flutter_scene_editor_app/lib/main.dart
```diff
@@ -116,21 +125,587 @@ class _EditorHomeState extends State<_EditorHome> {
...
+    } on FormatException catch (e) {
+      if (mounted) {
+        ScaffoldMessenger.of(
+          context,
+        ).showSnackBar(SnackBar(content: Text(e.message)));
+      }
+    }
+  }
+
+  void _closeProject() => setState(() => _project = null);
```
bdero: This bypasses `_setProject`, so the recursive lib/ watcher and `_sourceDebounce` survive Close Project. Every subsequent .dart save in the closed project still triggers `_runComponentGeneration` against it, writing generated files into a project the UI says is closed and adopting its schemas into whatever scene is now open.

packages/flutter_scene/lib/src/fscene/realize/declarative_codec.dart
```diff
@@ -0,0 +1,520 @@
...
+  };
+
+  /// The enum value named by string property [name], or the declared default
+  /// (then the first value) when absent or unrecognized.
+  T enumValue<T extends Enum>(String name, List<T> values) {
+    final raw = string(name);
+    for (final candidate in values) {
+      if (candidate.name == raw) return candidate;
+    }
+    return values.first;
```
bdero, line 439: The old decode used `values.byName`, which threw on unrecognized names. Falling back to `values.first` loads a newer or misspelled name silently wrong, and since serialize rebuilds from live component reads, the next save replaces the original string with the substitute, destroying it with no diagnostic. Worth at least a debugPrint on the unmatched name.

packages/flutter_scene/lib/src/fscene/realize/component_codec.dart
bdero: Done in 25f5555d. The registry now strips the enabled key from an enabled component's spec, so retained placeholder and deferred-splat specs cannot carry a stale false.

packages/flutter_scene/lib/src/fscene/realize/builtin_codecs.dart
bdero: Done in 25f5555d. The environment fallback is tagged with the authored origin, and the particle codec stashes an unresolved texture ref on the component, so both survive a save made without a realizer.

packages/flutter_scene_editor/lib/src/project/vm_service_link.dart
bdero: Done in e9e26103. A socket error or close marks the link dead (failing current and future requests), and AppSession drops a dead cached link and reconnects, keeping the hot-restart fallback reachable.

packages/flutter_scene/lib/src/fscene/source/source_scene_loader_io.dart
bdero: Done in 25f5555d. Only a permission denial (EPERM/EACCES, plus the Windows equivalent) deactivates for the run; other filesystem errors fall back for that load only, and deactivation drops source-key templates so their claims cannot strand.

packages/flutter_scene_editor/lib/src/project/app_session.dart
bdero: Done in e9e26103. launch() claims the session state before its first await (resetting on the failure paths), and stop() falls through to killing the process when app.stop replies with an error.

packages/flutter_scene/lib/src/fscene/realize/builtin_codecs.dart
bdero: Done in 25f5555d. The declarative write path applies authored values verbatim again; Range/IntRange only describe the editing UI (the editor still clamps on edit), so legal documents round-trip untouched.

packages/flutter_scene_editor/lib/src/project/fproject.dart
bdero: Done in e9e26103. The pubspec probe matches basenames, the root-prefix checks go through a separator-tolerant pathIsWithin, and resolvedDefaultScene shares the drive-letter absolute check with resolveWorkingDirectory.

packages/scene/lib/src/schema/component_schema.dart
bdero: Done in c5c96d51. The decode loop skips TypeError alongside FormatException, and options/formerNames/formerTypes materialize eagerly so an element mismatch cannot defer to render time.

apps/flutter_scene_editor_app/lib/main.dart
bdero: Done in e9e26103. Close Project routes through _setProject, which now also cancels the pending source debounce.

packages/flutter_scene/lib/src/fscene/realize/declarative_codec.dart
bdero: Done in 25f5555d. Both the reader and the write path debugPrint the unrecognized name and the substitute before falling back.

bdero:
All ten inline notes are addressed and resolved (replies on each thread). The body list landed across the same four commits,

- Play button error handling moved into the launch future, so launch-blocked errors reach the console (e9e26103).
- The version check recognizes git/hosted dependency blocks (no more false missing-dependency warning) and reads CRLF lockfiles (e9e26103).
- `FsceneComponentRegistry.register` evicts the replaced codec's serialize dispatch entry (25f5555d).
- Splat opacity/splatScale and fmod min/max distances read integers through the shared readers; physics_codecs dropped its local reader copies for property_read.dart, whose `readInt` now takes either numeric form (25f5555d).
- The hot-reload closure re-caches a template only while a claim remains (25f5555d).
- The fscene surface changes are in both CHANGELOGs and the format is version 3 with a no-op migration, so 0.20.x readers refuse new documents loudly instead of silently dropping nested keys (c5c96d51, d3cb046d).
- HookLogTailer stops at app.started, ProjectRunner coalesces notifications and trims in chunks, and declarative codecs memoize their field lists (e9e26103, 25f5555d).
- device_catalog probes use the shared `projectChildEnvironment()` scrub, IMPELLERC included (e9e26103).
- `registerPhysicsBackend` is exported from fscene.dart with categories on it and the other new public types, and the SoftRange doc no longer references another engine (d3cb046d, c5c96d51).

bdero (commented):
Follow-up pass over the fix commits. All ten earlier notes verified fixed, analyze is clean, and both test suites pass. Four smaller items inline, the Windows errno partition being the only one with user-visible behavior.

packages/flutter_scene/lib/src/fscene/source/source_scene_loader_io.dart
```diff
@@ -0,0 +1,357 @@
...
+  /// back to the bundled DataAssets. A permission denial (the macOS App
+  /// Sandbox being the common one; stat can succeed where open is refused)
+  /// additionally deactivates source loading for the rest of the run; a
+  /// transient failure (a sidecar mid-rewrite during a save, a file deleted
+  /// between stat and read) keeps it active so the next load retries the
+  /// sources. Anything else returns false.
+  bool deactivateOnAccessError(Object error, String key) {
+    if (error is! FileSystemException) return false;
+    final errno = error.osError?.errorCode;
+    // EPERM, EACCES, and Windows ERROR_ACCESS_DENIED.
```
bdero: 1 and 13 are EPERM/EACCES on POSIX, but on Windows OSError carries Win32 codes where 1 is ERROR_INVALID_FUNCTION and 13 is ERROR_INVALID_DATA, so unrelated Windows errors still hit the permanent-deactivation branch this fix reserved for denials. Partition by platform, `Platform.isWindows ? errno == 5 : errno == 1 || errno == 13`.

packages/flutter_scene/lib/src/fscene/realize/component_codec.dart
```diff
@@ -129,29 +138,82 @@ abstract class ComponentCodec {
...
   ComponentSpec? serialize(Component component, SerializeContext context) {
+    final spec =
+        _byComponentType[component.runtimeType]?.serialize(
+          component,
+          context,
+        ) ??
+        _serializeByScan(component, context);
+    if (spec != null) {
+      if (component.enabled) {
+        // Retained specs (placeholders, deferred loads) can carry a stale
```
bdero, line 220: This mutates the spec the codec returned, and for the placeholder and deferred-splat codecs that is the retained spec aliased by the live component, so serialized snapshots share one properties map with future toggles and the placeholder's carries-the-spec-untouched contract bends. A caller holding the returned ComponentSpec (undo history, document diffing) sees its snapshot change retroactively. Copying the spec here (or returning a copy from PlaceholderComponentCodec.serialize) avoids the aliasing.

packages/flutter_scene_editor/lib/src/toolchains/managed_checkout.dart
```diff
@@ -0,0 +1,441 @@
...
+
+  void _setPhase(ManagedCheckoutPhase phase) {
+    _phase = phase;
+    _line('== ${phase.label}');
+  }
+
+  void _line(String message) {
+    log.add(message);
+    if (log.length > 500) log.removeRange(0, log.length - 500);
+    notifyListeners();
```
bdero: Still one notifyListeners per line while ProjectRunner got the microtask coalescing; a `git clone`/`flutter precache` burst rebuilds listeners per line during checkout creation. Also worth carrying back, ProjectRunner's deferred notify can now fire after dispose (the debug disposed assert) if the runner is disposed mid-burst, so both spots want a disposed guard.

packages/flutter_scene/lib/src/fscene/realize/particle_emitter_codec.dart
```diff
@@ -354,45 +374,653 @@ List<ParticleModule> _modulesFromProperties(Map<String, PropertyValue> p) {
...
+// --- Small tolerant readers over MapValue/property bags ---
 
-BillboardFacing _facingFromName(String name) => switch (name) {
-  'axisLocked' => BillboardFacing.axisLocked,
-  'velocityStretched' => BillboardFacing.velocityStretched,
-  _ => BillboardFacing.spherical,
-};
+double _nonNegative(double value) => value < 0 ? 0 : value;
 
 double _num(Map<String, PropertyValue> p, String key, double fallback) {
```
bdero, line 1041: This private reader copy survived the physics dedup, as did fmod's `_configInt`/`_configBool`, and the splat codec's `shDegree` still requires an IntValue while `readInt` is now tolerant, so a `shDegree: 1.0` from a JSON writer is silently ignored where every other int field now rounds it.

packages/flutter_scene/lib/src/fscene/source/source_scene_loader_io.dart
bdero: Done in 119828c5, partitioned exactly as suggested.

packages/flutter_scene/lib/src/fscene/realize/component_codec.dart
bdero: Both cited sites already return copies (PlaceholderComponentCodec.serialize and the deferred-splat branch each rebuild the spec with a spread properties map, since d863e454/111e1851), so no current codec aliases a retained spec. 119828c5 documents the freshly-built requirement on ComponentCodec.serialize so the registry's in-place adjustment stays safe against future codecs.

packages/flutter_scene_editor/lib/src/toolchains/managed_checkout.dart
bdero: Done in 119828c5. ManagedCheckoutJob gets the microtask coalescing, chunked trim, and a disposed guard, and ProjectRunner's deferred notify now checks the same flag set in dispose().

packages/flutter_scene/lib/src/fscene/realize/particle_emitter_codec.dart
bdero: Done in 119828c5. The particle curve/distribution decoder and fmod's config readers now use property_read.dart, and splat shDegree reads through the tolerant readInt.

packages/flutter_scene_editor/lib/src/controller/editor_controller.dart
```diff
@@ -1153,7 +1511,7 @@ class EditorController extends ChangeNotifier {
   // Component codecs are synchronous once the retained resource realizer has
   // loaded the document, so this avoids rebuilding unrelated nodes/resources.
   bool _reflectComponents(Transaction transaction) {
```
bdero, line 1513: This lost the old `_composed != null` bail, and there is no prefab-instance guard below. For a components-slot edit on a single-root prefab instance node, `document.nodes[id].components` holds only host components while compose merged the prefab root's components into the composed instance node, so the swap plus the `..clear()..addAll(spec.components)` mirror wipes the prefab-authored components from both the live node and the composed doc, and the `return true` skips the corrective realize. Adding any host component to a mesh prefab instance makes its mesh vanish.

packages/flutter_scene_editor/lib/src/inspector/material_section.dart
```diff
@@ -197,6 +557,22 @@ class MaterialSection extends StatelessWidget {
...
+          _ => effective is num ? effective.toDouble() : 1.0,
+        };
+        return Padding(
+          padding: const EdgeInsets.symmetric(horizontal: 8),
+          child: SliderNumberField(
+            label: field.label,
+            value: current,
+            max: 10,
+            onPreview: (v) => _preview(field.key, math.max(v, 0.0)),
+            onCommit: (v) => _set(field.key, math.max(v, 0.0)),
```
bdero, line 573: `double.tryParse` accepts "NaN", "Infinity", and overflow like "1e999", the changed guard in `_finishTextEditing` (`next != _editStartValue`) passes for NaN, and `math.max(NaN, 0.0)` is NaN, so the value commits into the document and shader. `canonicalJson` then throws `FsceneEncodeException` on the non-finite value and `_writeTo` catches only `IOException`, so every Cmd+S fails uncaught with no snackbar until the poisoned field is re-edited. A `v.isFinite` check at commit (here and in the fmat field path) closes it.

packages/scene/lib/src/compose/compose.dart
```diff
@@ -328,6 +328,19 @@ void _applyDelta(
...
+  // Components the instance adds to prefab member nodes, applied before the
+  // overrides so an override can address their properties. Replaces a
+  // same-type prefab component, matching addComponent's semantics.
+  for (final add in spec.memberComponents) {
+    final node = out.node(remapId(add.member));
+    if (node == null) {
+      sceneLog('fscene: member component target ${add.member} not found');
+      continue;
+    }
+    node.components
```
bdero, line 343: Member components (and `addedComponents` below) enter the composed node by reference, unlike prefab-authored components which go through `_remapComponent` deep copies, and `_setProperty` later mutates `component.properties` in place for `components.<type>.<prop>` overrides. An override recorded on top of a member component therefore writes into the document's own delta spec, so undoing the override no longer reverts the property and the value persists twice. The reflect path hits the same aliasing through `applyPrefabOverride`.

packages/flutter_scene/lib/src/importer/inline_assets.dart
```diff
@@ -29,6 +29,48 @@ typedef ExternalImageAsset = ({String key, File file});
...
+/// package-relative keys, so the cooked `.fsceneb` resolves them through the
+/// DataAssets material registry (which keys `.fmat` sources by their
+/// package-relative path). [documentDir] is the scene source's directory
+/// relative to the package root, empty when the scene sits at the root.
+/// Absolute references and ones that escape the package root are left
+/// unchanged (they cannot resolve in a built app; the load reports them).
+void rebaseFmatMaterialRefs(SceneDocument document, String documentDir) {
+  for (final entry in document.resources.entries.toList()) {
+    final resource = entry.value;
+    if (resource is! MaterialResource || resource.type != 'fmat') continue;
```
bdero: `FmatSkySpec` refs never get rebased, here or in the source loader's `EnvironmentResource` branch, but the editor authors them document-relative through the same `referenceFmatAsset` helper while `loadFmatSky` keys the registry package-relative (`FmatSkySpec.asset` is documented package-root-relative). The editor renders correctly only because its `fmatSkyLoader` override resolves against the scene dir; in the cooked `.fsceneb` and in source-loaded dev runs the lookup fails and the sky silently falls back.

packages/flutter_scene_editor/lib/src/io/scene_io.dart
```diff
@@ -145,6 +145,92 @@ Future<LocalId?> importMaterialTexture(
...
+/// is the point of an fmat material.
+/// Returns a document-relative ref for the `.fmat` at [path] (`../materials/
+/// glass.fmat` style, matching prefab refs) whenever the file lies inside the
+/// scene's enclosing Flutter project, else [path] unchanged. Absolute refs do
+/// not survive outside the authoring machine and the runtime material
+/// registry cannot resolve them, so relative is the persistent form; a file
+/// outside the project stays absolute (session-only) rather than climbing
+/// across the disk.
+String referenceFmatAsset(String? sceneDir, String path) {
+  if (sceneDir == null) return path;
```
bdero, line 174: Two portability holes in the new document-relative refs. Save As never rewrites stored relative refs (saveFscene writes verbatim and _writeTo only flips baseDirectory afterward), so `../materials/glass.fmat` resolves wrong after saving into a different directory and the material silently fails to load. And this early return stores a machine-absolute path when the scene is unsaved, which the first save then persists verbatim, exactly the non-portable form the change was written to eliminate.

packages/flutter_scene_codegen/lib/src/project_generator.dart
```diff
@@ -0,0 +1,201 @@
...
+  final schemas = <ComponentSchema>[];
+  final sourcePaths = <String, String>{};
+  final expectedOutputs = <String>{};
+  final packageName = _packageName(root.path);
+
+  for (final file in sources) {
+    final relative = _relative(file.path, root.path);
+    final result = extractComponents(file.readAsStringSync(), path: relative);
+    diagnostics.addAll(result.diagnostics);
+    if (result.components.isEmpty) continue;
```
bdero, line 102: A transiently unparseable file also lands here, since the extractor parses with `throwIfDiagnostics: false` and never surfaces parse errors, making a syntax-broken file indistinguishable from a component-free one. The `continue` runs before `expectedOutputs.add`, so the stale sweep below deletes the file's generated codec and the registrar loses its registration, breaking the app build until the next clean save. The app-level watch sweep retires the component from the inspector the same way. Skipping deletion (and retirement) for files that produced diagnostics needs the extractor to expose them.

apps/flutter_scene_editor_app/lib/main.dart
```diff
@@ -116,21 +134,669 @@ class _EditorHomeState extends State<_EditorHome> {
...
+        if (schemas == null || schemas.isEmpty) continue;
+        if (!mounted) return;
+        // The app's registry is authoritative for app-derived types: a
+        // cache/live-provenance type absent from the fetch was deleted from
+        // the project (its stale cache entry would otherwise resurrect it
+        // every open). Source and package types re-derive locally.
+        final fetched = {for (final schema in schemas) schema.type};
+        final stale = [
+          for (final entry in _foreignSchemaProvenance.entries)
+            if ((entry.value == 'cache' || entry.value == 'live') &&
```
bdero, line 266: This sweep keys on the provenance label the source sweep's own comment calls unreliable, since a Play fetch stamps source-owned types 'live'. A later fetch from a stale build (registrar predating the type) then retires a source-owned component whose .dart file is unchanged on disk, degrading its inspector sections to placeholders until the next source save. Excluding types in `_sourceOwnedTypes` here matches the source sweep's own reasoning.

packages/flutter_scene/lib/src/scene.dart
```diff
@@ -1189,6 +1258,14 @@ base class Scene implements SceneGraph {
       return;
     }
 
+    // Consume a pending render-graph capture aimed at this screen view.
+    RenderGraphCapturer? capturer;
+    final pendingCapture = _pendingGraphCapture;
+    if (pendingCapture != null && pendingCapture.viewIndex == viewIndex) {
```
bdero, line 1279: The pending capture is consumed only here, after the not-ready and zero-size early returns and only for the matching viewIndex, with no timeout or error path. The inspector's `_run` holds `_busy` across the un-timed await, so a Capture clicked while the viewport is hidden, zero-height, or not ready leaves the panel's Capture/Scan dead for the session, and the MCP tool call hangs identically (its `_arm` also has no busy guard, so two concurrent calls fail the first with Superseded). A timeout that completes the pending capture with an error would cover all of these.

packages/flutter_scene/lib/src/fscene/realize/builtin_codecs.dart
```diff
@@ -1352,19 +1392,34 @@ class MaterialsVariantsCodec extends ComponentCodec {
...
+          // default, so assigning a material to a variants-carrying mesh is
+          // not silently reverted; the binding rebases onto it.
+          final current = mesh.primitives[primitiveIndex.value].material;
+          final currentId = resourceOrigin(current)?.resourceId;
+          final Material defaultMaterial;
+          if (defaultRef is! ResourceRefValue) {
+            defaultMaterial = current;
+          } else if (currentId != null &&
+              currentId != defaultRef.id &&
+              !variantIds.contains(currentId)) {
```
bdero: `currentId` is a LocalId in the origin document's id space but is compared against this document's ids with the origin document ignored, so a prefab-realized material whose id numerically collides with the recorded default reads as stale and is replaced. And the `currentId != null` guard means an untagged, code-assigned material always falls through to `realizer.material(defaultRef.id)`, which contradicts the comment's promise that an explicit assignment is not silently reverted. Comparing `origin.document` too and keeping untagged materials would close both.

bdero (commented):
Third pass, covering the 39 new commits (gizmos, render graph inspector, prefab components, fmat ref rebasing, LUT persistence). The follow-up commit resolved all four earlier notes. Fourteen inline notes this round; the prefab component wipe, the NaN save poisoning, and the member-component override aliasing are the ones I would hold the merge for. Smaller items below, roughly in priority order.

- Physics pending registrations live in a process-global set with removal only on unmount or successful retry, so a world-less editor scene discarded without detaching retains component, node, and scene graph process-wide and every later retry re-walks the dead entries.
- Render-graph capture's transient/shader-read/MSAA skip only applies when a descriptor is known, so descriptor-less textures (retained or swapchain-backed) are sampled by the blit anyway, and submit-time failures are asynchronous so `snapshotFailed` is never set.
- LUT reapply asymmetries in stage.dart, a LUT that fails to load defeats the in-place fast path forever (every look edit re-bakes IBL and retries the failing load), and with a null effects block a previously loaded LUT survives in place where a full realize would clear it.
- `RenderGraphMcp.scanForNans`, `_arm`, and `_isFloatFormat` duplicate the inspector core nearly line-for-line; routing the MCP tool through `scanForNonFinite` removes the drift risk (the copies already differ on a bounds check).
- `GizmoPrimitive.xray` is declared, round-tripped, and lowered through three packages but never read by the painter, and no TODO marks the gap.
- `MaterialResource` reconstruction appears field-by-field at both new rebase sites; a `copyWith` beside `PrefabInstanceSpec.copyWith` would keep future fields from silently dropping.
- The `TODO(fmat-cooked-refs)` in source_scene_loader_io.dart is stale, commit 7e083267 implemented the cook-time rebase it asks for (the real remaining gap is the sky ref above).
- The extractor drops a `softMin`-only annotation silently (SoftRange requires both ends); a diagnostic would honor the no-silent-gaps rule.
- MCP `capture_render_graph`/`get_pass_output` cast args unguarded while tools register with validateArguments false, so a string maxDimension escapes as a TypeError instead of a ToolError.
- The gizmo painter serializes every gizmo-bearing component per repaint with `shouldRepaint => true` (collider codecs run their full shape encoding), and strokes through per-segment Paint allocations with a fresh TextPainter per icon; snapshot caching keyed on the document epoch plus one Path per color would cut orbit-drag churn substantially.
- `referenceFmatAsset` hand-rolls an unbounded pubspec ancestor walk that can disagree with `findSceneProjectContext`'s bounded, `.fproject`-aware discovery.
- `_reapplyComposedNode` dispatches on hardcoded document path prefixes, duplicating the override path grammar the compose layer owns.
- render_graph_panel.dart hardcodes status reds/greens in the same wave that centralized panel chrome tokens; error/success tokens in editor_theme.dart would end the divergence.
- `RemapSettings.defaultsFor`'s float-format branch returns the same value as the fallthrough.

packages/flutter_scene_editor/shaders/editor_debug_remap.frag
```diff
@@ -0,0 +1,97 @@
...
+      color = vec3(1.0, 0.0, 1.0);
+    } else if (anyInf) {
+      color = vec3(1.0, 1.0, 0.0);
+    } else if (anyNegative) {
+      color = vec3(0.15, 0.3, 1.0);
+    }
+  }
+
+  vec4 base = texture(input_color, v_uv);
+  frag_color = mix(base, vec4(color, 1.0), remap_info.params.w);
```
bdero: With weight 1.0 this still evaluates `base * (1.0 - 1.0)`, and NaN times zero is NaN, so a NaN texel poisons frag_color even when the remap chose the magenta highlight, defeating highlightNonFinite exactly where it matters. Selecting the remapped color outright when the weight is 1 (or branching on the non-finite flag before the mix) avoids the propagation.

packages/flutter_scene/lib/src/scene.dart
```diff
@@ -1615,6 +1711,18 @@ base class Scene implements SceneGraph {
     // still bloom). Needs the perspective camera's FOV for the thin-lens
     // math and camera depth.
     if (depthOfField.enabled && perspectiveCamera != null) {
+      // Translucent depth-writing surfaces (glass) join the linear depth
+      // here, after the opaque-only consumers above, so depth of field
+      // focuses on the visible surface instead of the backdrop behind it.
+      graph.addPass(
+        TranslucentDepthPatchPass(
```
bdero, line 1735: The patch pass doc says it must run after every consumer of the linear-depth G/B/A channels, but custom passes at beforeToneMapping/afterToneMapping run later and read the same blackboard texture through `sceneDepthLinear`. With depth of field on and glass on screen, a custom pass that declared RenderInput.normals reads patched values at every glass pixel. Also worth noting, `keepDepthStencil: depthOfField.enabled` commits the non-transient depth and store bandwidth whenever DoF is on, while the field comment claims the cost only applies when a patch will actually run.

packages/flutter_scene_editor/lib/src/settings/open_in_editor.dart
```diff
@@ -0,0 +1,37 @@
...
+import 'dart:io';
+
+/// The placeholder an editor command carries for the file to open.
+const String kSourceFilePlaceholder = r'${SOURCE_FILE}';
+
+/// Expands [command] for [sourceFile]: the placeholder is replaced with the
+/// shell-quoted path, and when the command carries no placeholder the quoted
+/// path is appended as a final argument.
+String buildEditorInvocation(String command, String sourceFile) {
+  final quoted = Platform.isWindows
```
bdero: Two Windows quoting holes. This runs through `cmd.exe /c`, where `%` sequences expand even inside double quotes, so a path containing a matching `%...%` segment is rewritten. And the Browse button (settings_dialog.dart line 308) wraps a space-containing program path in single quotes, which cmd.exe does not treat as quoting, so a browsed `C:\Program Files\...` path splits at the space and every open-in-editor invocation fails.

packages/flutter_scene_editor/lib/src/mcp/render_graph_tools.dart
```diff
@@ -0,0 +1,307 @@
...
+  /// Renders one resource through the display remap as a PNG.
+  Future<ScreenshotResult> passOutput(
+    String key,
+    Map<String, Object?> options,
+  ) async {
+    final maxDimension = (options['maxDimension'] as num?)?.toInt();
+    final result = await _arm(
+      RenderGraphCaptureRequest(
+        thumbnailMaxDim: maxDimension,
+        fullResolution: maxDimension == null,
```
bdero, line 61: `maxDimension` is effectively ignored. Passing onlyKeys makes the engine keep a full-resolution snapshot regardless of this flag (`request.fullResolution || only != null` in render_graph_capture.dart), and the tool prefers `snapshot ?? thumbnail`, so a request for a 128px thumbnail of a 2560x1440 HDR target returns a full-res base64 PNG into the model context. Preferring the thumbnail when maxDimension is set fixes it.

packages/flutter_scene_editor/lib/src/render_graph/render_graph_inspector.dart
```diff
@@ -0,0 +1,235 @@
...
+    await _run(() async {
+      final capture = await _arm(
+        const RenderGraphCaptureRequest(thumbnailMaxDim: 256),
+      );
+      final resources = <InspectedResource>[
+        for (final resource in capture.resources)
+          InspectedResource(resource, _thumbnailImage(resource)),
+      ];
+      _result = capture;
+      _resources = resources;
```
bdero, line 175: The previous list's `thumbnail` images are never disposed here, and the class has no dispose path, so each Capture press strands one ui.Image (a GPU-backed RGBA8 texture) per resource row, dozens per click, reclaimed only by GC. Disposing the old thumbnails before the swap plus a dispose override covers it. Related costs nearby, the MCP `readPixel` arms a fresh capture and reads back the entire texture as rawExtendedRgba128 (16 bytes per pixel) per call to report one texel.

PR #305, opened 2026-08-10
DO NOT MERGE: probe tokenless Argos auth

bdero:
Throwaway probe. Blanks ARGOS_TOKEN on the linux shard so the CLI falls through to tokenless authentication, the path a fork pull request takes, and disables the other four shards to keep it to one upload.

Close without merging once the linux upload result is known.


bdero:
Tokenless auth confirmed working, build 9 uploaded with no token. Probe done.

PR #307, opened 2026-08-10
Unify scene coordinates and expand editor authoring

bdero:
Unifies runtime and offline imports around the native scene coordinate system, fixes prefab transforms and rendering selection costs, and expands authored environment and directional-light settings. Builds out desktop editor navigation, docking and layout persistence, environment controls, asset browsing, selection and gizmo performance, and compact inspector interactions.


packages/scene/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## 0.1.1
 
+- `.fscene` version 2 fixes every document to the native coordinate system.
```
bdero: Blocking. `scene` 0.1.1 is already on pub.dev, and the pubspec still says `0.1.1`, so these bullets land under a published version. This release removes `Handedness`, `UpAxis`, `StageMetadata.unitsPerMeter`, `NodeSpec.excludeFromWindingParity` and `SkyEnvironmentSpec.castShadows`, and bumps the format. That is `0.2.0`.

It cascades too. All six in-repo constraints (`flutter_scene`, `flutter_scene_box3d`, `flutter_scene_editor`, `flutter_scene_editor_core`, `flutter_scene_mcp` at `^0.1.0`, `flutter_scene_rapier` at `^0.1.1`) need widening, and `flutter_scene` 0.21.0 cannot publish while its `scene: ^0.1.0` resolves to the published 0.1.1, which has no `EnvironmentEffectsSpec`.

packages/scene/lib/src/json/fscene_json.dart
```diff
@@ -12,7 +12,7 @@ import 'package:scene/src/specs.dart';
 
 /// The current `.fscene` format version this build reads and writes.
 /// {@category Serialization}
-const int currentFsceneVersion = 1;
+const int currentFsceneVersion = 2;
```
bdero, line 15: Blocking. The version is 2 but `_builtInMigrations` is still `const []`, so every existing v1 document hard-fails with `FsceneVersionException: No migration from version 1`. Confirmed against a v1 document.

The chain exists for exactly this. A `handedness: "left"` document (editor-authored, which is what all five `examples/scenes/*.fscene` were before you hand-edited them here) needs nothing but the three dropped stage keys. Only `handedness: "right"` documents are genuinely unmigratable, and those are `.fsceneb` build artifacts the hook regenerates.

Suggest a `v1 -> v2` step that strips `upAxis`/`unitsPerMeter`/`handedness` when handedness is left, and throws a message naming the fix ("re-import the source glTF") when it is right. The current message tells a user nothing they can act on.

Separately, `migrateFscene` now throws on a missing `fscene` key where it used to assume the current version. Good tightening, but it is another break and neither CHANGELOG mentions it.

packages/scene/lib/src/compose/compose.dart
```diff
@@ -247,10 +240,14 @@ void _expandInstance(
...
     if (node.id == singleRoot) {
-      // Merge the prefab root into the instance node, keeping the instance's
-      // own transform, name, and layers (its placement in the host). An
-      // unnamed instance inherits the prefab root's name.
+      // Merge the prefab root into the instance node. The instance transform
+      // places the prefab in the host, outside the prefab root's own local
+      // transform. Keep the instance name and layers. An unnamed instance
+      // inherits the prefab root's name.
       if (instance.name.isEmpty) instance.name = node.name;
+      instance.transform = MatrixTransform(
```
bdero: Blocking, and the one I would fix first. Replacing the instance transform with a `MatrixTransform` means a later `transform.trs.*` override wipes the instance's rotation and scale, because `_setTrs` only reads the existing values `if (current is TrsTransform)` and otherwise starts from identity.

Repro, same test on both branches. Prefab root at identity, instance placed with a 1.0 rad Y rotation and scale 2.0, one override on `transform.trs.t`.

```
master: [[1.08, 0, 1.68, 9], [0, 2, 0, 9], [-1.68, 0, 1.08, 9], ...]  passes
this PR: [[1, 0, 0, 9], [0, 1, 0, 9], [0, 0, 1, 9], ...]  rotation and scale gone
```

Single-root is the common case, so this hits most prefab instances that carry a transform override. Cheapest fix is composing in TRS space when both sides are `TrsTransform`, or at minimum keeping the instance's `TrsTransform` untouched when the prefab root transform is identity. A fuller fix teaches `_setTrs` to decompose a `MatrixTransform` first.

packages/flutter_scene/lib/src/components/directional_light_component.dart
```diff
@@ -10,27 +11,38 @@ import 'package:flutter_scene/src/node.dart';
...
-/// shaded.
+/// Light travels along the owning node's local +Z axis. The node's world-space
+/// rotation aims the light; its translation and scale have no effect.
 /// {@category Scene graph}
-// TODO(lighting): shade multiple directional lights once the material
-// shader supports more than one.
 class DirectionalLightComponent extends Component {
   /// Creates a component that lights the scene with [light].
-  DirectionalLightComponent(this.light);
+  DirectionalLightComponent(this.light)
```
bdero, line 21: Blocking on the delivery, not the design. Node-aims-the-light is right. The problem is that `DirectionalLightComponent(DirectionalLight(direction: ...))` still compiles and now silently lights along world +Z, with no deprecation, assert or analyzer signal.

The migration cost is visible in this PR. Five examples rewritten, plus a new `_directionalLightNode` helper in `smoke_scenes.dart` that hand-builds `Quaternion.fromTwoVectors(Vector3(0, 0, 1), direction)`. Meanwhile `fromLocalDirection`, which solves exactly that, is `@internal` so outside users cannot reach it.

Two options that would make this a good breaking change rather than a silent one. Promote a public aiming constructor (`DirectionalLightComponent.aimed(light, direction)`), or assert in the default constructor when `light.direction` has been moved off its default. Either way `DirectionalLight.direction`'s doc should say plainly that a component ignores it, since a field honored in one attachment mode and discarded in the other is a trap.

packages/flutter_scene/lib/src/components/directional_light_component.dart
```diff
@@ -46,15 +58,26 @@ class DirectionalLightComponent extends Component {
...
-  Vector3 get worldDirection =>
-      node.globalTransform.getRotation() * light.direction;
+  /// The light's world-space travel direction.
+  Vector3 get worldDirection {
+    if (_usesLightDirection) return light.direction;
+    final localDirection = _localDirection;
+    if (localDirection != null) {
+      return node.globalTransform.rotate3(localDirection.clone()).normalized();
+    }
+    return node.globalTransform.getRotation() * _localForward;
```
bdero: Two small ones here. The `_localDirection` branch normalizes and this one does not, so the two paths hand back differently scaled vectors for no stated reason. And `_localForward` is a shared mutable `static final Vector3`. Nothing mutates it today (`Matrix3 * Vector3` allocates), but a shared mutable vector constant is easy to break later. `Vector3(0, 0, 1)` inline, or a getter, costs nothing.

Also worth noting that `refreshWinding`'s doc in `render_scene.dart` says it "combines node-transform parity with the geometry's source convention", but it only stores the node parity. The combining is in `windingFor`.

bdero (commented):
Reviewed the full diff against master, with `flutter analyze` clean and every package's tests passing.

The coordinate unification itself holds up. I checked the math end to end and the conversions are right, `F R F` as `(-x, -y, z, w)`, `F M F` negating entries where exactly one index is 2, the packer's float offsets 2/5/16/17 hitting position.z, normal.z, tangent.z and tangent.w, leaving index order alone because the mirror supplies the winding parity, and the bounds negate-and-swap. The three passes that still read `item.windingFlipped` never draw a LOD level, so the base geometry is correct there. `EnvironmentEffectsSpec.copy` covers all 82 fields and a generated full-field JSON round trip survives all 82.

Six things block, flagged inline. The most serious are a prefab override regression I reproduced against both branches, no v1 migration despite the version bump, and `scene` already being published at the version this bumps under. There are also two real bug fixes in here with no CHANGELOG line.

packages/flutter_scene/lib/src/fscene/realize/stage.dart
```diff
@@ -102,14 +108,16 @@ Future<void> realizeStage(
...
     radianceCubeSize: look?.radianceCubeSize,
     skybox: look?.skybox,
     skyEnvironment: look?.skyEnvironment,
+    effects: look?.effects,
     bundle: bundle,
     environmentLoader: environmentLoader,
     payloadLookup: payloadLookup,
   );
-  settings.applyLookTo(scene);
+  settings.applyTo(scene);
```
bdero: Blocking. Switching from `applyLookTo` to `applyTo` means realizing any document now overwrites `scene.bloom`, `fog`, `ambientOcclusion`, `screenSpaceReflections`, `godRays`, `depthOfField` and `autoExposure`. A document with no environment resource resets them all to off, since `effects: look?.effects` falls through to `EnvironmentEffectsSpec()` in `realizeEnvironmentSettings`.

`realizeStage` is public API and its docstring still claims it applies only environment, environment intensity, exposure, tone mapping, skybox and sky lighting. So an app that configures post-processing in code and then loads a scene silently loses it, and the docs say otherwise.

Either the docstring needs to state the new scope, or the document needs a way to say "leave this alone", since absent effects and explicitly-disabled effects currently decode identically.

packages/flutter_scene/pubspec.yaml
```diff
@@ -68,3 +68,4 @@ flutter:
     # A bundled sample environment, loadable via EnvironmentMap.fromAssets.
     # (The zero-config default is the procedural EnvironmentMap.studio().)
     - assets/royal_esplanade.png
+    - screenshots/flutter_scene_logo.png
```
bdero: Blocking. This is the 147 KB pub.dev screenshot, and declaring it under `flutter: assets:` bundles it into every app that depends on `flutter_scene`. The only consumers are `editor_shell.dart:1129` and the editor app's `main.dart:485`, neither of which ships in this package.

It belongs in `flutter_scene_editor`'s own assets. The `screenshots:` entry above stays as is, that one is metadata and costs consumers nothing.

packages/flutter_scene/lib/src/render/scene_pass.dart
```diff
@@ -455,20 +425,13 @@ class ScenePass extends RenderGraphPass {
...
       );
-      if (!_enableMsaa) {
-        _encodeCopy(translucentPass, currentColor);
-      }
-      if (shareFinalSnapshot) {
-        encoder.flushTranslucent(translucentPass: translucentPass);
-      } else {
-        encoder.flushNextTranslucentBatch(translucentPass: translucentPass);
-      }
+      _encodeCopy(translucentPass, currentColor);
```
bdero, line 444: The batching is a real win for the common case, but this dropped `_maxSceneColorSnapshotPasses = 8` and with it the `TODO(transmission-batching)` that named this exact gap. N mutually-overlapping transmissive objects now cost N iterations, each a fresh command buffer, a fullscreen copy, and when filtered a `bands`-deep downsample chain plus the atlas pass. That is unbounded where it used to degrade at 8.

Per the repo's TODO rule this needs a replacement `TODO(transmission-batching)` for the overlapping tail, and probably a re-introduced cap so a pathological scene degrades instead of stalling.

On the pass itself, the MSAA path is now correct as far as I can tell. `_encodeCopy` sets depth write off and compare always so the seeded copy survives the loaded depth, and the encoder resets its bind memo before the first draw so the copy's `setCullMode(none)` does not leak into the batch.

packages/flutter_scene/lib/src/scene.dart
```diff
@@ -1259,39 +1256,32 @@ base class Scene implements SceneGraph {
...
-    // either scene-color input splits the scene pass around the accumulated
-    // background. Scenes whose materials request nothing skip both. The same
-    // cached metadata fingerprints static shadow casters.
+    // Collect material inputs only from this view's frustum candidates. This
+    // avoids allocating depth and scene-color capture resources for effects
+    // that exist elsewhere in a large scene. Opaque occlusion remains a GPU
+    // depth-test concern because an AABB cannot prove a mesh is hidden.
+    // TODO(occlusion-culling): use a previous-frame hierarchical depth buffer
+    // to reject fully hidden render items and scene-input requests.
+    final materialInputs = renderScene.collectMaterialInputs(
```
bdero: This replaces a revision-keyed cache with a full BVH cull plus a `Set<RenderInput>` build per view, per frame. Narrowing to frustum candidates is the right idea, but given how much CPU work has gone into bind memoization elsewhere, paying an extra cull and an `addAll` per visible item every frame is a step backwards on the same budget.

Memoizing on `(structureRevision, materialSceneInputsRevision)` with a per-view frustum invalidate would keep the narrowing and most of the old savings.

One behavioral note as well. The set now changes with visibility, so depth and scene-color attachments can be acquired and released frame to frame as a transmissive object enters and leaves the frustum, which churns the transient pool. Probably fine, worth a sentence in the comment since it is a deliberate trade.

packages/flutter_scene/lib/src/render/skybox_encoder.dart
```diff
@@ -133,6 +131,21 @@ void _bindFrameInfo(
...
 }
 
+/// Builds the skybox transform without camera translation.
+///
+/// A skybox depends only on camera orientation. Removing translation before
+/// inversion avoids subtracting two large float32 world positions in the
+/// vertex shader.
+@visibleForTesting
+Matrix4 skyboxInverseViewProjection(Camera camera, ui.Size dimensions) {
+  final viewRotation = camera.getViewMatrix()..setTranslationRaw(0.0, 0.0, 0.0);
```
bdero: `getViewMatrix()` is mutated in place here. Safe for `PerspectiveCamera`, which builds a fresh matrix through `_matrix4LookAt`, but `Camera` is a public `abstract class` with no `base` or `final`, so a user subclass that returns a cached matrix gets its view transform silently zeroed. One `.clone()` before `setTranslationRaw`.

The change itself is good, and I checked that `projection.getProjectionMatrix(aspect) * viewRotation` matches `getViewTransform` exactly apart from the translation, so the sky stays aligned with geometry.

packages/flutter_scene/lib/src/fscene/realize/stage.dart
```diff
@@ -389,6 +514,11 @@ void serializeStage(Scene scene, SceneDocument document) {
   resource.toneMapping = scene.toneMapping.name;
   resource.agxWhite = scene.agxWhite;
   resource.agxContrast = scene.agxContrast;
+  final transform = scene.environmentTransform.storage;
+  resource.environmentRotationY = math.atan2(transform[6], transform[0]);
```
bdero, line 525: The math is right for a pure Y rotation (`storage[0]` is `m00` and `storage[6]` is `m02`, so `atan2` recovers the angle), but the runtime field is a full `Matrix3` and the document only stores an angle, so any non-Y component is silently dropped on save. Needs a `TODO(environment-transform)` saying what is lost and what would fix it, or the runtime field should be narrowed to match what the format can carry.

packages/flutter_scene/lib/src/importer/inline_assets.dart
```diff
@@ -26,6 +26,60 @@ import '../texture/mipmap.dart';
...
+/// Resolves the optional payload sidecar referenced by [document].
+ExternalPayloadAsset? resolveExternalPayloadAsset(
+  SceneDocument document,
+  Uri sceneSourceUri,
+) {
+  final key = document.payloadSource;
+  if (key == null) return null;
+  final file = _resolveAssetFile(sceneSourceUri.resolve('.'), key);
+  if (file == null) {
+    stderr.writeln(
```
bdero: A missing sidecar that the document explicitly declares warns and returns null, then `buildScenes` goes on to write a `.fsceneb` whose payloads have no bytes. That artifact fails later at load with a much less obvious error. Since `payloadSource` is an explicit declaration rather than a guess, this should throw and fail the build.

packages/flutter_scene/lib/src/fscene/realize/stage.dart
```diff
@@ -226,6 +340,10 @@ bool reapplyEnvironmentSettingsInPlace({
   required double environmentIntensity,
   required double exposure,
   required String toneMapping,
+  double agxWhite = 16.29,
```
bdero: These four new parameters all default to "reset to the runtime default", so a caller that forgets one silently wipes it rather than failing. The function is not exported and has exactly one caller, which does pass all four, so making them `required` costs nothing and removes the footgun.

packages/flutter_scene/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 ## 0.21.0
 
+* Breaking change, `.fscene` version 2 uses one native coordinate system and offline glTF import bakes source coordinates into it.
```
bdero, line 3: Two real user-visible fixes are in this branch with no entry here.

`node.dart` changes `parent.globalTransform.copyInverse(parentInverse)` to `parentInverse.copyInverse(parent.globalTransform)`. The old form wrote the inverse of the identity into the parent's cached world transform and then solved against identity, so assigning `globalTransform` both ignored the parent and corrupted its cache. That deserves a line.

`in_memory_import.dart` now forwards `lights` and `materialsVariants` into the normalized document, which were being dropped on that path.

The removals also want naming, since the barrel is a show list and these are all breaking. `Handedness`, `UpAxis`, `StageMetadata.unitsPerMeter`, `NodeSpec.excludeFromWindingParity`, `SkyEnvironmentSpec.castShadows`, and the `direction` property on the `directionalLight` component schema.

packages/flutter_scene/lib/src/fscene/realize/realize.dart
```diff
@@ -179,17 +173,11 @@ Node _realizeWith(
...
-    ? Matrix4.diagonal3Values(1.0, 1.0, -1.0)
-    : Matrix4.identity();
-
 /// Serializes the live [Node] graph rooted at [root] into a new
 /// [SceneDocument].
 ///
 /// [root] is treated as the synthesized realization root (as returned by
-/// [realizeScene]): its handedness is read back into the stage, and its
-/// children become the document's root nodes. Nodes realized from a document
+/// [realizeScene]) and its children become the document's root nodes. Nodes realized from a document
```
bdero: Nit, unreflowed at 102 chars after the edit. `dart format` leaves doc comments alone so it will stay.

While here, `SunLightSpec` is inserted out of alphabetical order in both `packages/scene/lib/scene.dart` and `packages/flutter_scene/lib/fscene.dart`, landing between `SkyEnvironmentSpec` and `SkySourceSpec`.

packages/scene/lib/src/compose/compose.dart
bdero: Fixed in 3b99db1. Identity prefab roots now preserve the instance TRS, and TRS overrides decompose a composed matrix before changing one component.

packages/scene/CHANGELOG.md
bdero: Fixed in 3b99db1. scene is now 0.2.0 and every in-repo scene constraint is widened to ^0.2.0.

packages/flutter_scene/lib/src/components/directional_light_component.dart
bdero: Fixed in 3b99db1. Both direction paths now normalize without a shared mutable vector, and refreshWinding documents only the parity it stores.

packages/scene/lib/src/json/fscene_json.dart
bdero: Fixed in 3b99db1. Left-handed v1 documents migrate by dropping the retired stage keys, right-handed documents report the re-import action, and the missing-version break is documented.

packages/flutter_scene/lib/src/fscene/realize/stage.dart
bdero: Fixed in 3b99db1. Environment resources now distinguish omitted effects from explicit defaults, so realizeStage preserves code-configured effects when the document does not override them.

packages/flutter_scene/pubspec.yaml
bdero: Fixed in 3b99db1. The logo is bundled by flutter_scene_editor only, while flutter_scene keeps the pub.dev screenshot as metadata.

packages/flutter_scene/lib/src/components/directional_light_component.dart
bdero: Fixed in 3b99db1. DirectionalLightComponent.aimed is public, the default constructor asserts on a custom light direction, and the direction field documents the attachment behavior.

packages/flutter_scene/lib/src/render/scene_pass.dart
bdero: Fixed in 3b99db1. Overlap-safe captures are capped at eight, the remaining tail degrades to a shared snapshot, and the required follow-up TODO is restored.

packages/flutter_scene/lib/src/scene.dart
bdero: Fixed in 3b99db1. Material inputs are cached per retained view across unchanged scene, material, candidate, frustum, layer, and culling-plane state.

packages/flutter_scene/lib/src/importer/inline_assets.dart
bdero: Fixed in 3b99db1. A declared missing payload sidecar now throws immediately and fails the build with the referenced path.

packages/flutter_scene/lib/src/render/skybox_encoder.dart
bdero: Fixed in 3b99db1. The camera view matrix is cloned before removing translation, with a regression test using a cached custom camera matrix.

packages/flutter_scene/lib/src/fscene/realize/stage.dart
bdero: Fixed in 3b99db1. The four look parameters are now required, and omitted effects explicitly preserve the target settings.

packages/flutter_scene/CHANGELOG.md
bdero: Fixed in 3b99db1. The transform-cache and in-memory import fixes are listed, along with every breaking API/schema removal.

packages/flutter_scene/lib/src/fscene/realize/stage.dart
bdero: Fixed in 3b99db1. The serializer now carries a TODO that names the discarded orientation components and the full-orientation format fix.

packages/flutter_scene/lib/src/fscene/realize/realize.dart
bdero: Fixed in 3b99db1. The doc comment is reflowed and SunLightSpec is ordered after the other S entries in both barrels.

packages/flutter_scene/lib/src/render/render_scene.dart
```diff
@@ -544,12 +597,14 @@ class RenderScene {
...
   void markBvhStructureDirty() {
     _structureDirty = true;
+    _materialCandidateRevision++;
   }
 
   /// Flags the BVH for a refit. Called when a bounded item moved but the
   /// item set and membership are unchanged.
   void markBvhBoundsDirty() {
     _boundsDirty = true;
+    _materialCandidateRevision++;
```
bdero: This is the one worth another round. A plain move bumps the candidate revision, so the cache it feeds misses every frame in any scene with a moving object.

```
final rs = RenderScene();
final before = rs.materialCandidateRevision;
rs.markBvhBoundsDirty();   // what refreshRenderItem does for an object that just moved
// materialCandidateRevision 0 -> 1
```

`_MaterialInputCache.matches` also compares the view transform storage bit-exactly, so a moving camera misses too. Between the two, the cache only hits when the camera is perfectly still and nothing in the scene moves, which is a screenshot rather than a running app. The per-frame cull I flagged last time is still paid everywhere it mattered.

The invalidation is not wrong, that is the trouble. Scoping the summary to the frustum makes it genuinely depend on every object's position, so a correct cache of that value has to invalidate on movement.

What I would do instead is cache the whole-scene set keyed on `(structureRevision, materialSceneInputsRevision)` alone, which movement does not touch, and use it as an early out. Empty means no view can request anything, so skip the cull entirely, and that is the common case this was meant to speed up. Only when it is non-empty do you pay the per-view cull to narrow it, and there you can drop the caching altogether since the scene already has the effects that make the cull worth running.

packages/flutter_scene/lib/src/components/directional_light_component.dart
```diff
@@ -46,15 +69,26 @@ class DirectionalLightComponent extends Component {
...
   /// payloads (geometry, materials).
   @override
-  Component? cloneFor(Node cloneOwner) => DirectionalLightComponent(light);
+  Component? cloneFor(Node cloneOwner) {
+    if (_usesLightDirection) {
+      return DirectionalLightComponent.fromLightDirection(light);
+    }
+    final localDirection = _localDirection;
+    return localDirection == null
+        ? DirectionalLightComponent(light)
```
bdero: `cloneFor` re-runs the authoring assert against whatever `light.direction` holds now, so cloning throws in debug once the shared light has been mutated.

```
final light = DirectionalLight();
final node = Node()..addComponent(DirectionalLightComponent(light));  // fine
light.direction.setValues(-1.0, -1.0, 0.0);                          // public and mutable
node.clone();
// Failed assertion: '_hasDefaultDirection(light)': DirectionalLight.direction is
// ignored by DirectionalLightComponent. Aim the owning node or use ...aimed.
```

A clone is copying an already-validated component, not authoring a new one, so it should not re-check. `node.clone()` is the model-instancing path, so this is a debug crash somewhere fairly hot. A private non-asserting constructor that `cloneFor` routes through covers it.

The assert itself is a good call and I would keep it exactly where it is for the authoring path.

packages/flutter_scene/lib/src/components/directional_light_component.dart
```diff
@@ -10,27 +11,49 @@ import 'package:flutter_scene/src/node.dart';
...
-  /// The light this component contributes. Its
-  /// [DirectionalLight.direction] is read in the owning node's local
-  /// space; [worldDirection] is the world-space result.
+  /// The light this component contributes.
   DirectionalLight light;
 
+  final bool _usesLightDirection;
+  final Vector3? _localDirection;
+
+  static bool _hasDefaultDirection(DirectionalLight light) {
```
bdero, line 59: These three literals are the constructor default from `light.dart:87` retyped in another file, with nothing tying them together. If that default ever moves, this silently stops matching and every default-constructed component asserts, which is a confusing failure to land on.

A `static final Vector3 defaultDirection` on `DirectionalLight` that both the constructor and this read would remove the coupling.

Worth noting in the class doc too that the check is construction-time only. Mutating `light.direction` afterwards is still silently ignored, which is fine as a best-effort guard, just better stated than discovered.

packages/scene/lib/src/json/fscene_json.dart
```diff
@@ -30,9 +30,33 @@ const Set<String> supportedFeatures = {
...
+  if (handedness != 'left') {
+    throw const FsceneVersionException(
+      'Right-handed version 1 documents cannot be migrated. Re-import the '
+      'source glTF.',
+    );
+  }
+  stage
+    ..remove('upAxis')
+    ..remove('unitsPerMeter')
+    ..remove('handedness');
```
bdero, line 67: The migration passes a left-handed document through even when it carries nodes with `excludeWindingParity: true`. The key is dropped as unknown and that node's mirror now counts toward winding parity, so its subtree renders inside out.

```
v1 input carries excludeWindingParity
migrated to v2, nodes 1
excludeWindingParity survives round trip: false
node determinant -1.0 (negative now counts toward winding parity)
```

Narrow, since the flag mostly appeared on compose's handedness adapters and those came from right-handed prefabs that already refuse. But a document that round-tripped through `serializeScene` could carry it, and silently changing how a subtree renders is the one outcome this migration exists to prevent. Refusing when any node has the key, with the same re-import guidance, is a few lines and keeps the migration honest about what it can and cannot carry forward.

bdero (commented):
Second pass over `3b99db1`. `flutter analyze` is clean and all four suites pass (1196 in flutter_scene, 76 in editor core, 60 in editor, 2 in scene).

All fifteen are addressed, and I re-ran the two things I had reproduced. The prefab override now keeps scale 2.0 and translation (9, 9, 9) both when the prefab root is identity and when it is not, so the `isIdentity` skip and the `decompose` fallback each carry their own case. The v1 migration behaves as intended, a left-handed document lands on version 2 and a right-handed one refuses with the re-import guidance. `overridesEffects` is a cleaner answer to the effects question than I suggested, and it is threaded through all seven copy sites. Nice touch making `.aimed` public, the smoke helper collapsing to one line is the proof it was the right shape.

Four new things, one of which I think undoes the point of the change it belongs to, plus two small residuals. Nothing here is a rewrite.

packages/flutter_scene/lib/src/scene.dart
```diff
@@ -871,12 +952,9 @@ base class Scene implements SceneGraph {
     ui.Rect? viewport,
     double? pixelRatio,
   }) {
-    renderViews(
-      [RenderView(camera: camera)],
-      canvas,
-      region: viewport,
-      pixelRatio: pixelRatio,
-    );
+    final view = _singleRenderView ??= RenderView(camera: camera);
```
bdero: Retaining the view is right, `_materialInputCaches` is keyed on view identity so a fresh one per frame would never hit. Two things follow from that which are worth writing down.

`renderViews` callers now have a performance contract they cannot see. Building `RenderView` objects per frame is a natural pattern, and it was free before, but it now costs a full material-input cull every frame and leaves a trail of Expando entries. Since `render()` had to be fixed for exactly this reason, the `renderViews` doc should say the views are expected to persist across frames.

Also worth a line here that the retained view is reused across `render()` calls, so two calls in one frame with different cameras share it. The cache keys on the view transform so it stays correct, it just thrashes, and that is easier to accept when the code says so.

packages/flutter_scene/lib/src/render/render_scene.dart
bdero: Fixed in f982461. The cache now stores only the whole-scene input set by structure/material revision, skips culling when it is empty, and performs uncached per-view narrowing otherwise.

packages/flutter_scene/lib/src/components/directional_light_component.dart
bdero: Fixed in f982461. DirectionalLight.defaultDirection is now the shared source, and the component docs state that the authoring check runs only at construction.

packages/scene/lib/src/json/fscene_json.dart
bdero: Fixed in f982461. The v1 migration now refuses any node carrying excludeWindingParity and reports the same source re-import guidance.

packages/flutter_scene/lib/src/components/directional_light_component.dart
bdero: Fixed in f982461. Clones now use a private non-asserting copy constructor, so later mutations of the shared light cannot crash model instancing.

packages/flutter_scene/lib/src/scene.dart
bdero: Fixed in f982461. The per-view identity cache and retained single RenderView were removed, so callers can construct views per frame without a new cache-lifetime contract.

bdero (commented):
Third pass over `f982461`. All five are addressed, and everything from the earlier two rounds still holds. `flutter analyze` is clean and all four suites pass (1199 in flutter_scene, 76 in editor core, 60 in editor, 2 in scene).

I re-ran the two behaviors I had reproduced, plus a check that the reworked cache still invalidates when it must.

```
clone survives a mutated shared light direction                 ok
movement no longer disturbs the whole-scene summary             ok
the supported material swap still invalidates the summary       ok
  materialSceneInputsRevision 0 -> 1
```

The material-input rework is the right shape now. Dropping `_materialCandidateRevision` entirely and gating on the whole-scene summary means a scene with no scene-reading materials pays one cached set lookup per frame instead of a cull, and a scene that has them pays the per-view cull it was always going to need. I checked the remaining key holds up. `structureRevision` covers add and remove, `materialSceneInputsRevision` covers a transmission toggle, an fmat hot reload, and a mesh swap whose input set differs, and `_refreshRetainedMaterials` only bumps it when the sets actually differ so an equivalent swap stays cached. The cached set is handed out by reference but every downstream use is `contains`, so nothing can poison it. Reverting `_singleRenderView` falls out for free once the Expando is gone, which is a nicer resolution than the doc note I suggested.

One optional nit inline, small enough to ignore. Nothing else from me on this branch.

packages/scene/lib/src/json/fscene_json.dart
```diff
@@ -30,9 +30,43 @@ const Set<String> supportedFeatures = {
...
+  if (handedness != 'left') {
+    throw const FsceneVersionException(
+      'Right-handed version 1 documents cannot be migrated. Re-import the '
+      'source glTF.',
+    );
+  }
+  final nodesValue = json['nodes'];
+  if (nodesValue is Map &&
+      nodesValue.values.any(
+        (node) => node is Map && node.containsKey('excludeWindingParity'),
```
bdero: Optional. `containsKey` refuses on `"excludeWindingParity": false` too, which is a document the old renderer treated as completely ordinary.

No document this repo ever wrote can hit it, since the v1 encoder emitted the key only when the flag was set, so this is hand-authored input only. `node['excludeWindingParity'] == true` would narrow it to the case that actually changes rendering. Fine to leave as is if you prefer the stricter read.

bdero: Fixed in 605447a. The migration now refuses only an explicit true value, with regression coverage confirming that explicit false still migrates.

PR #316, opened 2026-08-15
Add SMAA, lens flares, and reflection probes

bdero:
Three fidelity features and the deferred follow-ups from #313. SMAA 1x joins the anti-aliasing modes (a port of the Jimenez et al. reference implementation), lens flares render ghost chains and a halo inside the bloom pyramid, and parallax-corrected reflection probes capture the scene into local environments through the new Scene.captureEnvironment, blended in by camera position. The clearcoat lobe now responds to rect area lights, the screen-space indirect light reprojects its radiance history so the bounce stays put under camera motion, and MATERIALS.md documents the portable shader rules custom-shader authors need. Three new smoke goldens: smaa, lens_flare, and reflection_probe.


packages/flutter_scene/lib/src/scene.dart
```diff
@@ -1756,6 +1944,36 @@ base class Scene implements SceneGraph {
...
+            cameraForward.x,
+            cameraForward.y,
+            cameraForward.z,
+            0.0, //
+            position.x,
+            position.y,
+            position.z,
+            1.0,
+          );
+          ssgiReprojection =
```
bdero, line 1972: The temporal reprojection is the one part of this PR that CI can't cover: the smoke goldens are single static frames, where this collapses to identity (`_ssgiHistoryViewProjection` is null → current VP fallback, landing every tap on its own pixel). The bounce-stays-put behavior only manifests under camera motion, which nothing exercises automatically.

I checked the basis statically and it's consistent (+z-forward view space in `flutter_scene_gtao.frag` matching the `cameraForward`-as-+z column here), so I'm fairly confident it's right — but worth a manual moving-camera sanity check, and a candidate for a multi-frame integration test if the harness ever supports one.

packages/flutter_scene/shaders/material_lighting.glsl
```diff
@@ -712,7 +737,9 @@ vec4 EvaluateLighting(MaterialInputs material) {
   mat3 environment_transform = mat3(frag_info.environment_transform);
   vec3 env_normal = environment_transform *
                     (ao_bent_valid > 0.5 ? ao_bent_normal : normal);
-  vec3 env_reflection = environment_transform * reflection_normal;
+  vec3 env_reflection =
+      environment_transform *
+      ParallaxCorrectReflection(v_position, reflection_normal);
```
bdero, line 742: `env_reflection` is box-corrected once from the primary environment's `probe_box`, then used to sample **both** the primary and the crossfade secondary radiance. In the probe↔environment transition band the secondary (e.g. the studio sky) ends up sampled with a vector bent for the probe's box. It's transient and weight-blended so almost certainly invisible, but worth a one-line comment/TODO so it reads as a known limitation rather than a latent surprise.

packages/flutter_scene/lib/src/components/reflection_probe_component.dart
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  /// Stores a completed [capture] (called by the renderer), stamping the
+  /// parallax box from the probe's current placement.
+  @internal
+  void internalStoreCapture(EnvironmentMap capture) {
+    capture
+      ..parallaxBoxCenter = worldCenter
+      ..parallaxBoxHalfExtents = Vector3.copy(extents);
+    _environment = capture;
```
bdero, line 113: This replaces `_environment` without disposing the previous capture's GPU textures, and `EnvironmentMap` exposes no `dispose`. Flutter GPU reclaims textures via native finalizers, so it's not a true leak — but the editor's re-capture button / `requestCapture()` in a loop leans on GC to reclaim six faces + equirect + SH + prefiltered atlas each time, which can spike memory before collection. Fine for the occasional recapture the API targets; just noting the churn.

packages/flutter_scene/lib/src/render/bloom_pass.dart
```diff
@@ -94,33 +122,115 @@ class BloomPass extends RenderGraphPass {
...
+      );
+      up[i] = target;
+    }
+
+    var result = up[0]!;
+
+    // Lens flares generate from a well-blurred mip and composite over the
+    // finished bloom into another fresh cleared target, for the same
+    // write-once reason as the upsample.
+    if (_settings.lensFlare.enabled && down.length > 2) {
```
bdero: `down.length` is always `_kMipCount` (5), so `> 2` is dead. Harmless, but if it's meant as defense against a future variable mip count it might read clearer as a named constant check (or just drop it).

bdero (commented):
Reviewed the full diff and cross-checked the parts that aren't visible from the diff alone against the source at HEAD. Overall this is ship-quality: three well-scoped features cleanly separated per commit, consistent with the engine's existing spec → settings → pass → shader layering, and green across all seven backends with the Argos goldens accepted everywhere.

A few things I verified rather than assumed:

- **SSGI reprojection basis is internally consistent.** `getViewTransform` is the full projection×view (world→clip), and `flutter_scene_gtao.frag` uses a +z-forward view space (`screen_radius = proj_scale * radius / origin.z` with a positive divisor, `view_dir = -normalize(origin)`), which matches the `viewToWorld` matrix built with `cameraForward` as the +z column. The history color and history-VP are stored as a pair from the same frame, with correct read-before-write ordering.
- **The SMAA port honors the portability rules this PR documents** — unbounded reference searches rewritten as constant-bounded `for` loops with interior `break`, single-return helpers, `HistoryUv` kept single-return for the D3D compiler. The passing Windows/D3D and WebGL2 legs back that up.
- The bloom write-once refactor (adding `base` in-shader instead of a `LoadAction.load` accumulation) is correct, and the reasoning about backends re-clearing reloaded attachments is sound.

The inline notes below are all minor — none are blockers. The one I'd actually act on is the first: the temporal reprojection is the one behavior CI can't exercise.

packages/flutter_scene/lib/src/post_process/post_process.dart
```diff
@@ -133,4 +133,40 @@ class BloomSettings {
...
+class LensFlareSettings {
+  /// Whether lens flares run. Off by default.
+  bool enabled = false;
+
+  /// Strength of the flare features relative to the bloom they ride in.
+  double intensity = 1.0;
+
+  /// Number of internal-reflection ghosts along the line through the
+  /// screen center, at most 8.
+  int ghostCount = 4;
```
bdero, line 158: Doc says "at most 8" but the clamp only happens at upload time in `bloom_pass.dart`; the setter accepts any int and silently behaves as 8+. Consistent with how the other settings fields work, so likely intentional — flagging only in case you'd rather clamp at the setter for API honesty.

packages/flutter_scene/shaders/material_lighting.glsl
bdero: Done in 7c01196d. Added a note plus TODO(probe-crossfade-parallax) at the secondary sample so it reads as a known limitation rather than a latent surprise.

packages/flutter_scene/lib/src/components/reflection_probe_component.dart
bdero: Done in 7c01196d. Documented the finalizer churn and left TODO(probe-recapture-churn) to dispose the outgoing capture explicitly once EnvironmentMap gains a dispose.

packages/flutter_scene/lib/src/render/bloom_pass.dart
bdero: Done in 7c01196d. Dropped the always-true down.length > 2 and noted that _kMipCount (>= 3) guarantees the flare's source mip (up[2]).

packages/flutter_scene/lib/src/post_process/post_process.dart
bdero: Kept the setter open for consistency with the other settings fields; updated the doc in 7c01196d to say values above 8 are clamped when rendered, so the doc matches the behavior.

packages/flutter_scene/lib/src/scene.dart
bdero: Agreed, the moving-camera bounce is the one uncovered part. The +z-forward view basis here matches HistoryUv in flutter_scene_gtao.frag, and object motion still lags under TODO(velocity-reprojection). Leaving as-is; a multi-frame smoke lane is the right home for automated coverage if the harness ever supports one.

PR #332, opened 2026-08-17
Build generated assets without Dart data assets

bdero:
Generated assets (`.fsceneb`, `.fmat`, `.fstex`, shader bundles) no longer need Dart data assets, which are master-channel only, so `flutter pub add flutter_scene` now works the same way on stable 3.47 as it does on master. Hook outputs go into a `flutter_scene_generated/` directory the app lists once, a manifest maps source paths to bundled keys so `loadScene`/`loadFmatMaterial`/`loadTexture` take the same source path on every toolchain, and `dart run flutter_scene:init` writes the hook, the directory, the pubspec entry, and the gitignore rule.

An app that installs no hook at all still renders, including runtime `.glb` import on web, because flutter_scene's own hook builds the engine's shaders for whatever app is being built. Those outputs are stamped with the engine's identity and rebuilt when it changes, so a shader bundle is never reused across engine versions.

Sources over 1 MiB are fingerprinted by size and modification time instead of being read in full, which keeps large scenes out of the per-build hashing cost.

Breaking, `SceneAssetMode.legacyOnly` and its siblings are now `generatedTree` and are the default, and apps that set `dataAssetsRequired` keep the data-asset path.


bdero (commented):
Read the whole change against faaa731e. The generated tree, the manifest indirection that lets `loadScene` take one source path on every channel, and stamping each compiled bundle with the engine that built it all hold up. I ran the pubspec handling over CRLF files, inline flow lists, sequences flush with their key, and a pubspec with no `flutter` section, and yaml_edit gets all four right while keeping comments.

Four notes below. The first is a real functional bug, the next two are durability gaps that only bite later, and the last is signature drift.

One more thing worth a thought, every app in the repo still calls `buildEngineAssets` even though init stopped generating it in 7c3ef553, so each one ships a second copy of the engine bundles that only shadows the copy flutter_scene builds for it. That is also what makes the stale-entry case below reachable for anyone who copies an example.

packages/flutter_scene/lib/src/generated_assets/generated_asset_lookup.dart
```diff
@@ -0,0 +1,228 @@
...
+    final matches = keys
+        .where(
+          (key) =>
+              key.endsWith(dataAssetSuffix) &&
+              (package == null || key.startsWith('packages/$package/')),
+        )
+        .toList();
+    if (matches.length == 1) return matches.single;
+    if (matches.length > 1) {
+      throw StateError(
```
bdero: This `throw` sits inside the `try` above, so the block's own `catch (_)` swallows it and the ambiguity is never reported.

With two packages each registering `flutter_gpu_shaders/shaderbundles/my.shaderbundle`, `resolveShaderBundleKey('my')` falls through to the generated index, finds nothing, and throws `No shader bundle named "my" was found` along with advice to run `flutter_scene:init`, pointing the user at a hook that is already working. If some third package happens to have a tree-built bundle of that name, the call silently returns that unrelated key instead.

Building `matches` inside the try and deciding after it keeps the missing-manifest fallback and still reports the conflict.

packages/flutter_scene/lib/src/generated_assets/generated_file_names.dart
```diff
@@ -0,0 +1,72 @@
...
+  GeneratedAssetFamily family,
+  String nameId,
+  String extension,
+) {
+  final slash = nameId.lastIndexOf('/');
+  final rawStem = slash < 0 ? nameId : nameId.substring(slash + 1);
+  var stem = rawStem.replaceAll(RegExp(r'[^A-Za-z0-9_\-]'), '_');
+  if (stem.isEmpty) stem = 'asset';
+  if (stem.length > _maxStemLength) stem = stem.substring(0, _maxStemLength);
+  return '${family.prefix}.$stem.${shortHash('${family.name}/$nameId')}'
```
bdero: The name hashes `family.name/nameId` only, so the engine bundle in flutter_scene's own tree always lands at one fixed path whichever engine compiled it, even though the engine identity is in the freshness stamp.

Two projects on different Flutter versions sharing a pub cache and building at the same time both read a stale stamp, both compile, and both write that one path. The rename keeps either file whole, but the content is last-writer-wins, so the losing build ships a bundle its own engine refuses and dies at `loadShaderLibraryAsync`. Sequential builds are fine, since each one rebuilds on switch.

Folding the engine identity into `nameId` for this family would give the two builds separate files and make the shared cache safe under concurrency, not just in sequence.

packages/flutter_scene/lib/src/generated_assets/generated_tree.dart
```diff
@@ -0,0 +1,473 @@
...
+  /// Only entries this package owns are considered; a dependency's sources
+  /// resolve against that package's root, not the app's.
+  ///
+  /// TODO(legacy-prune): entries of a family whose builder was removed from
+  /// the hook outlive it as long as their sources exist. Recording the set of
+  /// builders that ran would let a save prune those too.
+  void pruneMissingSources() {
+    for (final entry in [..._manifest.entries]) {
+      final source = entry.source;
+      if (source == null || entry.owner != packageName) continue;
```
bdero, line 300: Entries owned by another package are skipped here, and nothing else drops them, so `buildEngineAssets` outputs recorded in an app's tree outlive the call that produced them. `_load` sorts the app tree ahead of package trees, so those stale copies keep winning over the ones flutter_scene rebuilds.

An app that follows the examples and later drops `buildEngineAssets`, which is what init now generates, keeps its old `shaderbundle.base.*` and `material.physical.*` indefinitely. After a Flutter version switch only flutter_scene's copy rebuilds, the app's stale copy still resolves first, and `Failed to load the engine shader bundle` asks for a rebuild that cannot clear it.

The TODO above names the mechanism. Dropping non-owned entries that a run did not re-record would close this case without needing the full builder-set tracking.

packages/flutter_scene/lib/src/fmat/target_shader_bundle_unsupported.dart
```diff
@@ -1,18 +1,16 @@
...
   required Object buildOutput,
   required String manifestFileName,
   List<Uri> includeDirectories = const [],
   TargetShaderBundleAssetMode assetMode =
-      TargetShaderBundleAssetMode.legacyOnly,
+      TargetShaderBundleAssetMode.generatedTree,
   String? dataAssetName,
   int? glesLanguageVersion,
+  bool copyToGeneratedTree = true,
+  String? owner,
```
bdero, line 26: `pruneGeneratedTree` and `stamp` are missing here, and the native signature has both. The other stubs in this change stayed in sync.

packages/flutter_scene/lib/src/generated_assets/generated_asset_lookup.dart
bdero: Fixed in 1d461aae. The matches are collected inside the guard and judged after it, so a missing manifest still falls through to the generated index while a genuine ambiguity reports.

packages/flutter_scene/lib/src/generated_assets/generated_file_names.dart
bdero: Fixed in 1d461aae. `generatedFileName` takes an optional variant that feeds the hash without changing the id the manifest maps, and the engine bundle passes the engine identity, so two builds sharing the directory write different files and the sweep drops whichever the manifest stops naming. Threaded through `buildTargetShaderBundleJson` as `fileVariant`, since that is where the file is actually written.

packages/flutter_scene/lib/src/fmat/target_shader_bundle_unsupported.dart
bdero: Fixed in 1d461aae, plus the `fileVariant` parameter added in the same change.

packages/flutter_scene/lib/src/generated_assets/generated_tree.dart
bdero: Partly addressed in 1d461aae, and I did not take the suggested prune as written.

Every app that only built the engine shaders drops the call, so the state this needs is no longer reachable by copying an example. The smoke harness keeps it deliberately, as the only coverage of the app-side override.

Dropping foreign entries a run did not re-record is right in principle but wrong per instance, since each builder opens its own tree, so `buildScenes` saving would drop what `buildEngineAssets` just recorded. A per-process run marker fixes that, but it must also not fire while the run emits data assets, since one `flutter run` invokes the hook under several asset-type configs and the data-asset invocation records nothing into the tree, which is the case 4763fbee fixed. The TODO now spells out both constraints rather than implying the simple version works.

packages/flutter_scene/lib/src/fmat/build_materials.dart
```diff
@@ -235,20 +256,38 @@ Future<void> _buildMaterials({
...
         .toFilePath(),
   );
-  final sidecarFile = File(
-    packageRoot
-        .resolve('build/shaderbundles/$bundleName.fmat.json')
-        .toFilePath(),
+  final shippedBundleFile = tree == null
+      ? bundleFile
+      : File.fromUri(
+          tree.fileUri(
```
bdero, line 301: No `variant` here, so the fix that landed for `base` does not cover the physical material bundle, which `buildOwnEngineAssets` compiles into the same shared `flutter_scene_generated/` under flutter_scene's package root.

Printing what each side actually resolves to makes the asymmetry clear.

```
engine=aaaa1111
  base     -> shaderbundle.base.817491ed.shaderbundle
  physical -> material.physical.a1754d52.shaderbundle
engine=bbbb2222
  base     -> shaderbundle.base.1b00055d.shaderbundle
  physical -> material.physical.a1754d52.shaderbundle
```

`physical.shaderbundle` is engine-compiled exactly like `base` and just as invalid under another engine, so two builds on different Flutter versions sharing the pub cache still land on that one path and the content is last-writer-wins. That is the failure the variant was added to remove, only on the other bundle. The sidecar and index share the name and the same exposure.

Threading a variant through `_buildMaterials` for the bundles written on flutter_scene's behalf would close it. The app-owned `materials` bundle in an app's own tree has no sharing problem, so it does not need one.

packages/flutter_scene/lib/src/generated_assets/generated_tree.dart
```diff
@@ -0,0 +1,483 @@
...
+    final directory = Directory.fromUri(_root);
+    if (directory.existsSync()) {
+      for (final file in directory.listSync(followLinks: false)) {
+        if (file is! File) continue;
+        final name = file.uri.pathSegments.last;
+        // Only ever delete files matching the generated naming scheme, so a
+        // keeper file in the tree survives.
+        if (!isGeneratedFileName(name)) continue;
+        if (referenced.contains(name)) continue;
+        file.deleteSync();
```
bdero, line 360: Smaller residual on the same theme. The variant stops two engines from writing one path, but the sweep still deletes the file the other engine is named by, since only the id survives in the manifest and the losing variant is unreferenced.

Driving two saves against one tree shows it.

```
build(engine=X) wrote shaderbundle.base.6225e207.shaderbundle
build(engine=Y) wrote shaderbundle.base.6225e154.shaderbundle
files left: .gitignore, manifest.json, shaderbundle.base.6225e154.shaderbundle
```

Sequential builds are unaffected, each switch recompiles and the tree keeps one bundle, which is the documented behavior. Concurrently, the build that lost has its asset removed between its hook and asset bundling, so it fails on a missing file rather than shipping the wrong bytes. That is a better failure than before, so this is a note rather than something to hold the change for. Skipping generated names that are another variant of an id still in the manifest would leak one file per engine ever used, which is bounded and probably the cheaper trade.

examples/stress_bench/pubspec.yaml
```diff
@@ -13,10 +13,16 @@ dependencies:
...
 
 dev_dependencies:
   flutter_lints: ^5.0.0
 
 flutter:
   uses-material-design: true
+  assets:
+    # Everything hook/build.dart generates, loaded by source path. Added
+    # by `dart run flutter_scene:init`.
+    - flutter_scene_generated/
```
bdero: Vestigial now that this app's hook is gone, since nothing writes into its tree and flutter_scene ships the engine bundles from its own. Harmless either way, and worth keeping if the intent is that the app is one hook away from generating its own assets.

bdero (commented):
Re-reviewed against 1d461aae. All four are addressed, and I verified each rather than reading for it.

`resolveShaderBundleKey` now reports the conflict, and I checked the four cases around it (two matches throws, `package` disambiguates, a single match still resolves, no manifest still falls through to the tree). The web stub is at exact parity with the native signature, `fileVariant` included. The Windows rebasing bug is a good catch that I missed entirely, and `rebaseShaderBundleManifest` is the right seam to have pulled out for it. The expanded TODO on `pruneMissingSources` is a fair place to leave the stale foreign entry, and it now describes the real constraint, that the run marker has to be per process and must not fire on the data-asset invocation.

One thing the variant did not reach. `base` gets the engine identity in its file name, but the physical material bundle, compiled by the same `impellerc` into the same shared directory, still resolves to one engine-independent name. Detail inline, plus a smaller residual on the sweep and a vestigial pubspec entry.

Everything green here, 1426 tests and a clean analyze on flutter_scene plus the example and smoke apps.

packages/flutter_scene/lib/src/fmat/build_materials.dart
bdero: Fixed in f348c201. `buildBundledPhysicalMaterials` takes a `fileVariant` and `buildOwnEngineAssets` passes the engine identity, so the bundle, its sidecar, and its index all separate by engine. `buildMaterials` is untouched, since an app-owned bundle in the app tree has no sharing problem.

packages/flutter_scene/lib/src/generated_assets/generated_tree.dart
bdero: Taken in f348c201, since keeping a bounded leak beats a concurrent build losing its asset. The sweep keeps a file whose name is another variant of one a manifest entry references, keyed on the name with its hash tag stripped. A source move or a removed source still deletes the old output explicitly, so the leak is one file per engine an output was compiled by. Covered both ways, one test asserts both variants survive a save, another that an unrelated orphan is still swept.

examples/stress_bench/pubspec.yaml
bdero: Removed in f348c201, along with the directory. `init` re-adds both if the app ever grows a hook.

bdero (commented):
Re-reviewed against ba8442a4. Both items from the last round are properly done, and I checked each rather than reading for it.

`physical` now separates by engine across all three of its outputs, bundle, sidecar and index, and the sweep keeps another variant of a live name so the concurrent build is no longer robbed of its asset. `generatedNameWithoutTag` returns null for a non-generated name, so the keeper file and the manifest are still safe from the sweep. Dropping the stress bench's directory and its pubspec entry is the right call now that it has no hook.

Two things on the new code, one in the relative rebasing and one in what the retention costs. Both inline.

1431 tests and a clean analyze across flutter_scene and the example, smoke and stress apps.

packages/flutter_scene/lib/src/generated_assets/build_engine_assets.dart
```diff
@@ -0,0 +1,274 @@
...
+  if (file.scheme != from.scheme) return file.toString();
+  var common = 0;
+  while (common < base.length &&
+      common < target.length - 1 &&
+      base[common] == target[common]) {
+    common++;
+  }
+  // Nothing shared past the root means no usable relative path (on Windows,
+  // different drives), so hand back a URI both consumers can still read.
+  if (common == 0) return file.toString();
```
bdero: This bail reads as "different Windows drives", but on posix it means "different first directory", which is an ordinary layout with a perfectly good relative path. Running the real function over a few roots.

```
app packageRoot              flutter_scene sourceRoot        result
/Users/me/app/               /Users/me/.pub-cache/fs/        ../.pub-cache/fs/shaders/one.frag
/home/runner/work/a/a/       /home/runner/.pub-cache/fs/     ../../../.pub-cache/fs/shaders/one.frag
/app/                        /root/.pub-cache/fs/            file:///root/.pub-cache/fs/shaders/one.frag
/builds/proj/                /opt/pub-cache/fs/              file:///opt/pub-cache/fs/shaders/one.frag
/C:/app/                     /C:/pkg/fs/                     ../pkg/fs/shaders/one.frag
/C:/app/                     /D:/pkg/fs/                     file:///D:/pkg/fs/shaders/one.frag
```

The last row is the case the fallback is for. Rows three and four are posix layouts that relativize fine as `../root/.pub-cache/...` and `../../opt/pub-cache/...`, and they are what a container gives you when the workdir and `HOME` sit at different top-level directories.

The fallback is also not readable by both consumers the way the doc says. `Uri.resolve` takes it, but the comment two lines up is right that the compiler opens the value as a path, and `file:///root/...` is not one, so it resolves against the package root and misses. That is unavoidable across Windows drives, but here it turns a working layout into a build failure.

Keying the bail on the roots actually differing (a drive letter or a differing scheme) rather than on `common == 0` would keep the Windows behavior and let posix relativize. Worth a case in the rebase test too, since all three fallback tests use drive letters and none covers a posix split root.

Scope is apps calling `buildEngineAssets`, since flutter_scene's own hook has `packageRoot == sourceRoot` and never rebases.

packages/flutter_scene/lib/src/generated_assets/generated_tree.dart
```diff
@@ -0,0 +1,496 @@
...
+    final directory = Directory.fromUri(_root);
+    if (directory.existsSync()) {
+      for (final file in directory.listSync(followLinks: false)) {
+        if (file is! File) continue;
+        final name = file.uri.pathSegments.last;
+        // Only ever delete files matching the generated naming scheme, so a
+        // keeper file in the tree survives.
+        if (!isGeneratedFileName(name)) continue;
+        if (referenced.contains(name)) continue;
+        if (variantsOfReferenced.contains(generatedNameWithoutTag(name))) {
```
bdero: Retention is right for the race, but nothing bounds it, and this branch cannot tell a concurrent build's file from dead weight. Your own tree on this branch already shows it.

```
manifest references:
  material physical            -> material.physical.be9b3e37.index.json
  material physical#shaderbundle -> material.physical.be9b3e37.shaderbundle
  material physical#sidecar    -> material.physical.be9b3e37.fmat.json
  shaderBundle base            -> shaderbundle.base.4a11bc62.shaderbundle

on disk:
  material.physical.a1754d52.{shaderbundle,fmat.json,index.json}   <- unreferenced, 1.0M
  material.physical.be9b3e37.{shaderbundle,fmat.json,index.json}
  shaderbundle.base.4a11bc62.shaderbundle
```

`a1754d52` is the variantless name from before f348c201. It is referenced by nothing and is now kept permanently, because it looks like another variant of a live output. The directory is 2.5M where 1.5M is live.

Three things compound it. The directory is a listed asset, so the dead copy ships. It deliberately survives `flutter clean`, so the usual reflex does not clear it. For a consumer the tree is in the pub cache, shared, so it accumulates a copy per engine any project on that machine ever compiled with, and every app built from that cache carries all of them.

Nothing loads them, `isFmatIndexAssetKey` wants `/flutter_scene/fmat/` and a tree key has neither that nor the `packages/` prefix an app tree could match, so this is size rather than correctness. Still, a `flutter build` that silently grows by a megabyte per Flutter version is a rough edge for something with no cleanup path. Bounding it on age would keep the race closed, since the file a concurrent build is named by was written moments ago, while a variant from a Flutter version retired weeks back is safe to drop. Failing that, the tree could use a documented way to clear itself.

packages/flutter_scene/lib/src/generated_assets/build_engine_assets.dart
bdero: Fixed in 4ae9667e. The bail keys on the roots actually differing now, a differing scheme or a differing drive letter, so your rows three and four relativize as `../root/.pub-cache/...` and `../../opt/pub-cache/...` while the C:/D: row still falls back. You are right that the fallback is not readable by both consumers, the doc no longer claims it is; it is now only reached where no relative path exists at all. Added the posix split-root cases (all three from your table) and a same-drive Windows case, since every fallback test used drive letters before.

packages/flutter_scene/lib/src/generated_assets/generated_tree.dart
bdero: Fixed in 4ae9667e. Retention is bounded by age, `variantRetention` of one hour, so a file a racing build is named by is kept while one from a retired engine is swept. That keeps the race closed, since a concurrent build wrote its file moments ago, and bounds the directory to the live set plus anything from the last hour.

On the `a1754d52` set specifically, it is unreferenced and past the window, so the next run of flutter_scene own hook sweeps it. Worth noting it does not clear on the spot, the sweep only runs when the hook actually executes, which is the "Skipping target: build_hooks" behavior this branch documents elsewhere. That is the right coupling though, a new engine forces a recompile, and the sweep that follows drops the previous engine variant, so the steady state is one live set rather than one per engine ever used.

PR #346, opened 2026-08-20
Compressed asset decoding, morph targets, and new lighting and grounding features

bdero:
glTF import gains required-extension validation with structured warnings, sparse accessors, Draco and meshopt decompression, standard KTX2 (Basis Universal) textures, and multi-buffer documents. The renderer and material system gain morph targets, light channel masks, directional cascade shaping, a shadow catcher ground material, planar reflections, baked lightmaps, pre-baked environment map ingestion, per-primitive visibility, per-instance material attributes, and typed material parameter read-back.

Assembled from eleven feature branches, one merge each, with the 0.24.0 CHANGELOG entry listing every addition.


packages/flutter_scene/lib/src/importer/src/gltf/extensions.dart
```diff
@@ -0,0 +1,57 @@
...
+  'KHR_materials_transmission',
+  'KHR_materials_volume',
+  'KHR_materials_unlit',
+  'KHR_texture_transform',
+  'EXT_meshopt_compression',
+  // Only widens which attributes may use the quantized component types the
+  // accessor readers already convert, so no dedicated parsing is needed.
+  'KHR_mesh_quantization',
+  'KHR_draco_mesh_compression',
+  'KHR_texture_basisu',
```
bdero, line 29: `EXT_lights_image_based` is missing from this set, but this PR does parse it. `parser.dart` reads `extensions['EXT_lights_image_based']` at both the document and scene level, and `ImageBasedLightComponent` surfaces the result.

Two consequences fall out of the omission. A document listing it in `extensionsUsed` now gets a spurious `glTF uses unrecognized extension "EXT_lights_image_based", ignoring` warning even though the importer handled it, and the fixture in `test/gltf_image_based_light_test.dart` declares exactly that. A document listing it in `extensionsRequired` throws `UnsupportedRequiredExtensionException` for an extension the engine supports.

Adding the name here fixes both. A test asserting the warning list is empty for the IBL fixture would keep the two lists from drifting again.

packages/flutter_scene/lib/src/geometry/morphed_geometry.dart
```diff
@@ -0,0 +1,402 @@
...
+
+/// Skinned geometry with morph targets, blended before the skin matrix is
+/// applied on both paths.
+///
+/// A skinned morphed draw spends two vertex-stage samplers (the joints
+/// texture plus the morph texture). With the 15-sampler lit fragment stage
+/// that totals 17 combined units, one over the GLES 3.0 minimum of 16, so a
+/// driver at the bare minimum cannot bind a lit skinned morphed draw (real
+/// GLES3 hardware reports 32 or more).
+/// TODO(morph-sampler-budget): pack the joint matrices and morph deltas
```
bdero, line 86: Good that the combined-limit cost is spelled out here. The gap is that nothing enforces it. `test/physical_material_sources_test.dart` still caps fragment samplers at 15 and its comment still reasons from "a skinned draw (which spends one on `joints_texture`)", which stopped being the worst case the moment `MorphedSkinnedVertex` shipped.

So the guard now passes at exactly the count that breaks a lit morphed skinned draw on a 16-unit driver, and a future sampler addition would be waved through by a test whose stated invariant no longer matches the engine. Either derive the cap from the vertex-stage sampler count of the variant being paired, or assert the vertex counts alongside the fragment ones so the pairing is what is checked.

packages/flutter_scene/lib/src/light.dart
```diff
@@ -236,12 +293,21 @@ class DirectionalLight {
     final far = shadowMaxDistance;
 
     // Practical split scheme: a blend of logarithmic and uniform
-    // spacing, so the near cascades get proportionally more resolution.
+    // spacing, so the near cascades get proportionally more resolution. A
+    // pinned first bound takes the first split and the same scheme spreads the
+    // rest from there; a single cascade has no rest, so it keeps far.
+    final bound = firstCascadeFarBound;
+    final pinned = bound != null && count > 1
+        ? bound.clamp(near, far).toDouble()
```
bdero: `bound.clamp(near, far)` allows `pinned == far`, and then `splitNear == far` makes every subsequent split evaluate to `far` as well. The result is cascade 0 covering the entire range at cascade-0 resolution and cascades 1 through n-1 as zero-thickness duplicates, silently burning most of the atlas.

The symmetric case at `near` gives a degenerate cascade 0. Clamping strictly inside the range, or ignoring the pin once it reaches `shadowMaxDistance`, keeps the remaining cascades meaningful.

packages/flutter_scene/lib/src/material/shadow_catcher_material.dart
```diff
@@ -0,0 +1,427 @@
...
+class ShadowCatcherMaterial extends Material {
+  /// Creates a shadow catcher. All parameters can be reassigned later.
+  ShadowCatcherMaterial({
+    Color shadowColor = const Color(0xFF000000),
+    double shadowIntensity = 0.8,
+    double aoStrength = 0.5,
+    double softness = 0.0,
+    double fadeStart = 0.0,
+    double fadeEnd = 0.0,
+    ShadowCatcherMode mode = ShadowCatcherMode.baked,
```
bdero: Defaulting to `baked` makes the out-of-the-box behavior the one that goes silently wrong. A user who animates the sun, or moves the model, sees a shadow stuck under the old pose with no error and no visual cue that a cache is stale, and the fix is a method they have to know exists.

`live` as the default with `baked` as the deliberate opt-in matches the usual correct-first ordering, and the product-scene case that wants baked is exactly the case where the author is already tuning the material. If baked stays the default, auto-dirtying on a change to the directional light's world transform would cover the most common way it goes stale.

packages/scene/lib/src/specs.dart
```diff
@@ -419,6 +421,50 @@ class GeometryResource extends ResourceSpec {
...
+/// Morph target data on a [GeometryResource]: the delta payload plus target
+/// metadata.
+///
+/// The [deltas] chunk holds dense float32 delta slabs, target-major and
+/// aligned with the geometry's vertex order: every target's position deltas
+/// (`targetCount * vertexCount * 3` floats), then the same shape of normal
+/// deltas when [hasNormalDeltas], then tangent deltas (xyz) when
+/// [hasTangentDeltas].
+/// {@category Documents}
+class MorphTargetsSpec {
```
bdero, line 438: `MorphTargetsSpec` and `GeometryResource.morphTargets` are new public API on `scene`, which is published at 0.2.0, but neither `packages/scene/pubspec.yaml` nor its CHANGELOG changes here. The repo's 0.2.0 and pub.dev's 0.2.0 now describe different surfaces.

Related, `currentFsceneVersion` stays at 3 while a version-3 geometry resource gains a `morphTargets` key. The published `scene` 0.2.0 reads such a document without complaint (the newer-than-supported check never trips) and drops the morph data on the floor. Worth an explicit call on whether this is an intentional additive key inside version 3 or a version 4, since the earlier additive changes in the 0.2.0 notes did take a version bump.

packages/flutter_scene/CHANGELOG.md
```diff
@@ -1,3 +1,22 @@
+## 0.24.0
+
+* glTF import validates `extensionsRequired`, refusing a file that needs an unsupported extension with every missing extension named, and surfaces unrecognized `extensionsUsed` entries through a new `onWarning` callback on every import entry point instead of loading silently.
```
bdero: These run three and four clauses each, which is closer to release-note prose than to the short punchy bullets the repo convention asks for. The KTX2, morph target, and environment-KTX2 entries in particular carry trailing detail that a reader scanning the changelog will not use. Naming the API and what it does is enough, the rest belongs in the dartdoc.

bdero (commented):
Read through the whole diff. The engineering standard here is high, the decoders are hardened and cite their specs, the shader work respects the GLSL ES 1.00 constraints, and almost every shortcut already carries a `TODO(topic)`. Notes below are mostly about guards and release hygiene rather than the feature code.

One thing worth calling out at the top. Eleven features land here, most of them rendering features (morph targets, baked lightmaps, the shadow catcher, planar reflections, KTX2 textures, Draco and meshopt geometry), and `examples/smoke_render` gains no scene for any of them. The argos matrix reporting "no diff detected" across all seven backends is therefore vacuous for everything in this PR, and per CLAUDE.md a rendering change that must hold across backends gets verified on the matrix. The morph path in particular is the first vertex-stage `texelFetch` plus `gl_VertexIndex` in the engine and nothing in CI exercises it on the WebGL2 shim, GLES, or Vulkan. At minimum a morph scene and a shadow-catcher scene would cover the two riskiest additions.

packages/flutter_scene/lib/src/importer/src/gltf/extensions.dart
bdero: Done in d712c197. Added to the recognized set, and the IBL fixture test now asserts an empty warning list so the two cannot drift again.

packages/flutter_scene/lib/src/geometry/morphed_geometry.dart
bdero: Done in c05e3a7c. The sources test now compiles all four vertex variants, pins their sampler counts (0/1/1/2), and asserts the combined pairing table, with the 17-sampler morphed skinned case documented as the deliberate exception tied to TODO(morph-sampler-budget). The cap rationale comment was refreshed to name the real worst case.

packages/flutter_scene/lib/src/light.dart
bdero: Done in 4af47683. A pin at or past shadowMaxDistance now falls back to the automatic scheme, and the lower clamp stays strictly above the near plane; tests cover both degenerate ends and the pre-existing clamp test was updated to the new semantics.

packages/flutter_scene/lib/src/material/shadow_catcher_material.dart
bdero: Done in 16a0074c. live is now the default and baked is the deliberate opt-in, with the class docs reworded to lead with the correctness ordering. The smoke scene passes both modes explicitly, so the golden is unaffected.

packages/scene/lib/src/specs.dart
bdero: Done in dadb86cd, taking the version-4 call. scene is now 0.3.0 with a CHANGELOG entry, currentFsceneVersion is 4 (moved next to SceneDocument.formatVersion so they cannot drift) with an identity 3-to-4 migration, and every in-repo scene constraint widened to ^0.3.0. A version-3 reader now refuses a morph-bearing document instead of silently dropping the deltas.

packages/flutter_scene/CHANGELOG.md
bdero: Done in 3421c0c0. Every 0.24.0 bullet is now one line naming the API and what it does.

bdero (commented):
Second pass. All six points from the last round are addressed, and several of the fixes are better than what I asked for.

The smoke coverage in particular. Eight scenes across seven backends, all green, and `morph_skinned` draws a rest pose beside a morphed one from a skinned-and-morphed synthetic GLB, so the two-vertex-sampler pairing I was worried about is now rendered on WebGL2 and Android GLES rather than argued about. `basisu_textures` folding Draco, UASTC-with-zstd, and ETC1S-with-alpha into one fixture is a nice bit of scene economy. The vertex-sampler pinning test is the right shape too, since it fails loudly on either stage moving instead of restating the fragment cap.

Six smaller things below, one of them a real defect. `EXT_meshopt_compression` is now the only new decoder with no render coverage, which I think is fine given its golden byte tests and the shared packing path with Draco, but worth a conscious call rather than an oversight.

packages/scene/lib/src/scene_document.dart
```diff
@@ -12,6 +12,11 @@ import 'package:scene/src/specs.dart';
 /// and packed binary), composition (prefab overrides), and realization live
 /// in separate layers.
 /// {@category Documents}
+/// The `.fscene` format version this build reads and writes. Newer documents
+/// are refused; older ones migrate on read.
+/// {@category Serialization}
+const int currentFsceneVersion = 4;
```
bdero: This const landed between `SceneDocument`'s doc comment and the class, so the class declaration no longer has one. Reading top to bottom, the whole `The in-memory .fscene document ...` block plus its `{@category Documents}` now documents `currentFsceneVersion`, which also carries its own `{@category Serialization}`, and `SceneDocument` is left undocumented and uncategorized.

So dartdoc gets a const with two categories and prose about id pools, and the package's central type drops out of the Documents category entirely. Moving the const above the `SceneDocument` doc block (or below the class) restores both.

packages/flutter_scene_rapier/pubspec.yaml
```diff
@@ -32,7 +32,7 @@ dependencies:
     sdk: flutter
   hooks: ^2.0.0
   # The engine-agnostic physics contract this package implements.
-  scene: ^0.2.0
+  scene: ^0.3.0
```
bdero, line 35: Widening the constraint here is right, but it needs a version bump to go with it. Published `flutter_scene_rapier` 0.5.0 pins `scene: ^0.2.0`, so the moment `scene` 0.3.0 and `flutter_scene` 0.24.0 land on pub.dev, no consumer can take both flutter_scene 0.24.0 and any published rapier.

This is the lockstep the release notes call out, a `scene` minor forces a rapier re-release because the caret pins scene's minor. `native/src` is untouched, so it is the no-cross-build variant of the checklist (bump, CHANGELOG entry, then tag and create the release by hand naming the release whose binaries it reuses). Fine as a follow-up PR if you would rather keep this one to engine features, but it should not be discovered at publish time.

packages/flutter_scene/lib/src/material/shadow_catcher_material.dart
```diff
@@ -0,0 +1,428 @@
...
+
+/// How a [ShadowCatcherMaterial] evaluates its shadow term each frame.
+/// {@category Materials}
+enum ShadowCatcherMode {
+  /// The shadow atlas (cascades plus spot tiles) is sampled per fragment
+  /// every frame. Always correct, including moving lights and casters.
+  live,
+
+  /// The atlas shadow is rendered once into a low-resolution footprint
+  /// cache, separable-blurred, and sampled per frame, the right default for
```
bdero: Stale after the default moved. `baked` is no longer "the right default for static product scenes", it is the opt-in for them. The class doc and the `mode` setter doc both got updated in the same commit, so this enum member is the one left describing the old behavior.

packages/flutter_scene/lib/src/light.dart
```diff
@@ -155,6 +161,32 @@ class DirectionalLight {
...
 
+  // TODO(fscene): expose these two through DirectionalLightCodec; the pinned
+  // bound needs a nullable number field, which the codecs have no shape for.
+  /// View distance, in world units, at which the first cascade ends. `null`
+  /// (the default) lets [shadowCascadeSplitLambda] place it.
+  ///
+  /// Pin it to hold a known resolution over the near field, for example the
+  /// reach of a third-person camera, and let the rest of the range spread
+  /// across the remaining cascades under the usual split scheme. Clamped
+  /// between the camera near plane and [shadowMaxDistance], and ignored when
```
bdero: This still describes the old clamping. The behavior is now asymmetric, a pin below the near plane is lifted to just above it, but a pin at or past `shadowMaxDistance` is ignored outright and the automatic scheme runs instead. The new tests assert exactly that, so the doc is the only thing still saying "clamped between".

packages/flutter_scene/CHANGELOG.md
```diff
@@ -1,3 +1,22 @@
...
+* Standard KTX2 textures (`KHR_texture_basisu`) load on both import paths, transcoded in pure Dart.
+* The runtime importer accepts multi-buffer `.gltf` documents and percent-decodes resource URIs.
+* `MeshPrimitive.visible` and `MeshPrimitive.castsShadow` toggle one primitive's color and shadow passes independently.
+* `MaterialParameters` gains typed getters, `hasParameter`, `isParameterAssigned`, and parameter enumeration.
+* Light channels, `channelMask` on lights and `Node.lightChannelMask` restrict what a light illuminates; `DirectionalLight.shadowCasterChannelMask` filters shadow casters independently.
+* `DirectionalLight.firstCascadeFarBound` pins the first cascade split; `cascadeOverlap` cross-fades cascade hand-offs.
+* `ShadowCatcherMaterial`, an invisible ground showing only received shadows and ambient occlusion, live by default with a baked mode.
+* Morph targets import on both paths, animate via `weights` channels, and blend on the CPU or a GPU morph texture; `Node.morphWeights` drives them.
+* `EnvironmentMap.fromKtx2Bytes` loads a pre-baked radiance cubemap with SH-9 diffuse, skipping the load-time prefilter.
```
bdero, line 15: The tightening pass reads much better, but this bullet lost the `EXT_lights_image_based` clause and with it the only mention of that work anywhere in the notes. `ImageBasedLightComponent` is a new export in `lib/scene.dart`, so `dart_apitool` will report it at release with nothing in the CHANGELOG to match, which is the check step in the release runbook. Worth its own short bullet rather than a clause bolted back onto this one.

The `diffuse_sh.dart` cluster (`ShDiffuseSummary`, `parseDiffuseShSidecar`, `encodeDiffuseShSidecar`, and friends) is in the same position, nine new exports covered only by the words "SH-9 diffuse".

examples/smoke_render/lib/smoke_scenes.dart
```diff
@@ -1096,6 +1555,205 @@ final List<SmokeScene> kSmokeScenes = <SmokeScene>[
...
+  // sweeps red to blue across the grid and dark to bright along it, and the
+  // float lifts each cube (vertex stage) while ramping its gloss (fragment
+  // stage), so a mispacked instance record scrambles the gradient or flattens
+  // the staircase.
+  SmokeScene('instance_attributes', () {
+    final scene = Scene();
+    scene.add(
+      _directionalLightNode(
+        vm.Vector3(-0.35, -1.0, -0.45),
+        DirectionalLight(intensity: 2.5),
```
bdero: This light does not cast, so the scene renders the instance-attribute path through the color pass only. The depth and shadow variants are the interesting half here, since MATERIALS.md documents that they bind no instance data and read zero, which means the shadow of an attribute-lifted cube stays flat on the ground while the cube floats above it.

That divergence is deliberate and documented, but nothing pins it. A `castsShadow: true` on this light costs one flag and turns a prose contract into a golden, so a later change to `_writeInstanceAttributes` or to the `instanceSchema` selection in `shadow_encoder.dart` cannot quietly alter it.

packages/scene/lib/src/scene_document.dart
bdero: Done in 0099068c. The const now sits above the doc block, so SceneDocument keeps its prose and its Documents category.

packages/flutter_scene_rapier/pubspec.yaml
bdero: Done in c84e15de. rapier is 0.5.1 with a CHANGELOG entry noting the widened constraint and that a release reuses the 0.5.0 binaries; the tag-and-release-by-hand step stays with the publish per the checklist.

packages/flutter_scene/lib/src/material/shadow_catcher_material.dart
bdero: Done in e349a0c9.

packages/flutter_scene/lib/src/light.dart
bdero: Done in 91184be9. The field doc now spells out both asymmetric ends, lifted above near, ignored at or past shadowMaxDistance.

packages/flutter_scene/CHANGELOG.md
bdero: Done in 56de57aa. ImageBasedLightComponent and the SH helper cluster each get their own bullet, so the apitool diff has matching notes at release.

examples/smoke_render/lib/smoke_scenes.dart
bdero: Done in 22234a18. The light casts (shadowMaxDistance 20) and a small receiver plane sits under the grid, so the flat unlifted shadow grid under the stair-stepped cubes is now pinned by the golden; verified on the macOS lane, corners stay clear and the gradient is intact.

PR #347, opened 2026-08-21
Let a raw ShaderMaterial declare engine scene inputs

bdero:
A translucent surface needs `scene_color` and `scene_depth` to refract and to measure its own thickness, and a surface that displaces geometry from a texture needs a vertex shader that can sample one. Today only `PreprocessedMaterial` can do the first (`.fmat`'s `engine_inputs:`) and only `ShaderMaterial` can do the second, so a refractive water surface cannot be written at all.

`sceneInputs` moves onto `ShaderMaterial` as a settable field, reusing the binding path `PreprocessedMaterial` already uses. Assigning a different set invalidates the frame's cached input summary; assigning an equal one does not. Additive, and a material that declares nothing behaves exactly as before.

Two things this reaches that the `.fmat` path does not. Engine inputs are `lit`-only there because the accessors ride the lighting framework, while a raw material writes its own and can be unlit. And a raw material pays none of the lit framework's eight engine samplers, so the budget has room for the inputs.

### DIFF
--- a/MATERIALS.md
+++ b/MATERIALS.md
@@ -815,6 +815,52 @@ See `examples/flutter_app/lib/example_raw_shader.dart` with
 
 ---
 
+# Engine inputs on a raw `ShaderMaterial`
+
+A `ShaderMaterial` can declare the same per-frame engine textures a `.fmat`
+material asks for with `engine_inputs:`, which is what a translucent surface
+needs to refract and to measure its own thickness.
+
+```dart
+final water = ShaderMaterial(
+  vertexShader: library['WaterVertex']!,
+  fragmentShader: library['WaterFragment']!,
+  isOpaqueOverride: false,
+  sceneInputs: const {RenderInput.opaqueSceneColor, RenderInput.depth},
+);
+```
+
+`RenderInput.depth` binds `scene_depth`, the opaque geometry's linear
+view-space depth in world units, and forces the depth prepass.
+`RenderInput.opaqueSceneColor` binds `scene_opaque_color`, the scene behind
+this draw, and `RenderInput.filteredSceneColor` adds its roughness-filtered
+atlas as `scene_filtered_color` for rough refraction.
+
+Nothing is generated for you here, unlike the `.fmat` path. Declare the
+samplers yourself under those names and derive the screen UV from
+`gl_FragCoord`:
+
+```glsl
+uniform sampler2D scene_opaque_color;
+uniform highp sampler2D scene_depth;
+uniform PostFrameInfo { vec4 resolution; } frame;
+
+vec2 uv = gl_FragCoord.xy * frame.resolution.zw;
+float behind = texture(scene_depth, uv).r;   // linear view depth, world units
+vec3 refracted = texture(scene_opaque_color, uv + offset).rgb;
+```
+
+`sceneInputs` is settable after construction; assigning a different set
+invalidates the frame's cached input summary, and assigning an equal one
+does not. The engine produces an input only while a visible material asks for
+it, so declaring none costs nothing.
+
+Unlike the `.fmat` path this is available on `unlit` shading, and a raw
+material pays none of the lit framework's eight engine samplers, so the
+sampler budget has room for the two inputs.
+
+---
+
 # Render state
 
 A `.fmat` material declares render state in its `material` block: `culling`
--- a/packages/flutter_scene/lib/src/material/shader_material.dart
+++ b/packages/flutter_scene/lib/src/material/shader_material.dart
@@ -1,6 +1,7 @@
 import 'dart:typed_data';
 
-import 'package:flutter/foundation.dart' show debugPrint, visibleForTesting;
+import 'package:flutter/foundation.dart'
+    show debugPrint, setEquals, visibleForTesting;
 
 import 'package:flutter_scene/src/gpu/gpu.dart' as gpu;
 
@@ -8,6 +9,8 @@ import 'package:flutter_scene/src/light.dart';
 import 'package:flutter_scene/src/material/engine_lighting.dart';
 import 'package:flutter_scene/src/material/material.dart';
 import 'package:flutter_scene/src/material/shader_stage.dart';
+import 'package:flutter_scene/src/render/custom_render_pass.dart'
+    show RenderInput;
 import 'package:flutter_scene/src/texture/texture2d.dart';
 import 'package:flutter_scene/src/render/frame_transients.dart';
 
@@ -106,7 +109,8 @@ class ShaderMaterial extends Material {
     this.cullingMode = gpu.CullMode.backFace,
     this.windingOrder = gpu.WindingOrder.counterClockwise,
     this.isOpaqueOverride = true,
-  }) {
+    Set<RenderInput> sceneInputs = const {},
+  }) : _sceneInputs = sceneInputs {
     if (fragmentShader != null) {
       setFragmentShader(fragmentShader);
     }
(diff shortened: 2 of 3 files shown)


bdero (commented):
The shape is right and the change really is small, so most of this is about the half that is not here yet rather than the half that is.

The structural point. `engine_inputs` on a `.fmat` is four things and this exposes one of them. The samplers get bound, but the availability gates (`frag_info.scene_inputs.x`/`.y`) and the screen mapping (`GetScreenUv()`, which is `gl_FragCoord.xy * frag_info.ssao_params.zw`) both ride `FragInfo`, and a `ShaderMaterial` never binds `FragInfo`. So a raw material cannot tell whether an input was produced this frame and cannot compute the UV to sample it with, which is why the doc example has to reach for a block nothing binds. Worth closing in this PR rather than after it, because every hand-written shader that ships against the current shape bakes in its own copy of the screen mapping, and that is the piece most likely to need to change for a backend.

Suggested shape, all additive. Move the accessors into a `shaders/scene_inputs.glsl` and have both paths include it, the way the emitter already includes `filtered_scene_color.glsl`, then bind a small block here carrying the gates, the inverse resolution and the camera basis when the set is non-empty. One definition, no drift, and the raw path stops being a weaker copy of the generated one.

The rest is smaller. The set is stored and returned by reference, the enum accepts values that silently do nothing while still costing a per-view collection, and I think there is a pre-existing invalidation hole that this makes much easier to hit. Details inline.

MATERIALS.md
```diff
@@ -815,6 +815,52 @@ See `examples/flutter_app/lib/example_raw_shader.dart` with
...
+atlas as `scene_filtered_color` for rough refraction.
+
+Nothing is generated for you here, unlike the `.fmat` path. Declare the
+samplers yourself under those names and derive the screen UV from
+`gl_FragCoord`:
+
+```glsl
+uniform sampler2D scene_opaque_color;
+uniform highp sampler2D scene_depth;
+uniform PostFrameInfo { vec4 resolution; } frame;
```
bdero, line 846: Nothing binds `PostFrameInfo` on a material draw. It comes from `PostEffectPass` and `RenderPassContext.applyShader`, both full-screen passes, so a shader copying this gets an unbound block and a UV of zero.

There is also no good way for the author to fill it in themselves. The target size is the logical size times the device pixel ratio times `Scene.renderScale`, and the material draws into the scene color target rather than into the widget, so it is not something app code can reliably compute.

MATERIALS.md
```diff
@@ -815,6 +815,52 @@ See `examples/flutter_app/lib/example_raw_shader.dart` with
...
+
+vec2 uv = gl_FragCoord.xy * frame.resolution.zw;
+float behind = texture(scene_depth, uv).r;   // linear view depth, world units
+vec3 refracted = texture(scene_opaque_color, uv + offset).rgb;
+```
+
+`sceneInputs` is settable after construction; assigning a different set
+invalidates the frame's cached input summary, and assigning an equal one
+does not. The engine produces an input only while a visible material asks for
+it, so declaring none costs nothing.
```
bdero, line 856: This says what declaring none costs but not what declaring one costs. `opaqueSceneColor` splits the scene pass around each screen-reading layer, `depth` forces the depth prepass, and any non-empty set moves the frame off the cached whole-scene answer onto a per-view collection. Raw shader authors are exactly the people who want that spelled out.

The settability is worth framing as a declaration rather than a knob, too. Each real change invalidates the whole-scene summary, so driving it per frame recomputes `collectAllMaterialInputs` over the scene.

packages/flutter_scene/lib/src/material/shader_material.dart
```diff
@@ -106,7 +109,8 @@ class ShaderMaterial extends Material {
     this.cullingMode = gpu.CullMode.backFace,
     this.windingOrder = gpu.WindingOrder.counterClockwise,
     this.isOpaqueOverride = true,
-  }) {
+    Set<RenderInput> sceneInputs = const {},
```
bdero, line 112: Worth narrowing what this accepts. `RenderInput` also carries `normals` and `shadowMap`, which `bindSceneInputTextures` ignores and which `scene.dart` never reads back out of `materialInputs`, so they do nothing. They are not free, though. Any non-empty set moves the frame onto the per-view `collectMaterialInputs` path, so a value that does nothing still buys a frustum query per view per frame.

Either assert on the three that mean something, or give materials their own enum. The shared one also has the type promising more than the material path can deliver, which is the part that gets awkward to walk back once it is public.

Related, `.fmat` treats `filtered_scene_color` as implying scene color, and here `{filteredSceneColor}` alone binds only `scene_filtered_color`. Worth normalizing wherever the set is accepted.

packages/flutter_scene/lib/src/material/shader_material.dart
```diff
@@ -106,7 +109,8 @@ class ShaderMaterial extends Material {
...
+  }) : _sceneInputs = sceneInputs {
```
bdero, line 113: The set is stored by reference and the getter hands the same object back, so `material.sceneInputs.add(...)`, or mutating the set the caller passed in, both skip the invalidation the setter exists to provide. Copying in and returning an unmodifiable view closes it.

packages/flutter_scene/lib/src/material/shader_material.dart
```diff
@@ -137,6 +141,33 @@ class ShaderMaterial extends Material {
   /// translucent pass with alpha blending.
   bool isOpaqueOverride;
 
+  Set<RenderInput> _sceneInputs;
```
bdero, line 144: This makes the Dart set and the GLSL two independent declarations that have to agree, where `engine_inputs` drove emission and binding from one place. Declaring an input whose sampler the shader declares but never meaningfully samples is the eliminated-sampler segfault from shorebreak's crashes file, not a no-op, so the doc should say that outright.

Longer term the shader is the better source of truth, since the reflection already knows whether `scene_opaque_color` is there. Worth deciding now whether this field is the permanent shape or a stepping stone toward deriving it, because the answer changes how much needs documenting around it.

packages/flutter_scene/lib/src/material/shader_material.dart
```diff
@@ -137,6 +141,33 @@ class ShaderMaterial extends Material {
...
+  /// behind this draw, and [RenderInput.filteredSceneColor] adds its
+  /// roughness-filtered atlas as `scene_filtered_color`. Together they are what
+  /// a translucent surface needs for refraction and depth-fade absorption.
+  ///
+  /// Nothing is generated for you, unlike a `.fmat` material: declare the
+  /// samplers yourself under those names and derive the screen UV from
+  /// `gl_FragCoord`. The engine produces an input only while a visible material
+  /// asks for it, so an unused one costs nothing.
+  @override
+  Set<RenderInput> get sceneInputs => _sceneInputs;
```
bdero, line 160: I think there is an invalidation hole here that predates this PR but gets much easier to hit through it. `MeshPrimitive.material` is a plain field, so swapping a material into a primitive that is already in the scene bumps neither `structureRevision` nor `materialSceneInputsRevision`, and `_cachedWholeSceneMaterialInputs` stays stale. Construct one of these with `sceneInputs` and assign it to a live primitive and the engine never produces the buffer.

That is read off the source rather than from a repro, so worth confirming. If it holds, a setter on `MeshPrimitive.material` that bumps the revision is cheap, and it matters more here than for `.fmat` because without the gates the failure renders white instead of fading out.

packages/flutter_scene/lib/src/material/shader_material.dart
```diff
@@ -371,6 +402,15 @@ class ShaderMaterial extends Material {
     if (useEnvironment) {
       _bindEnvironmentTextures(pass, shader, lighting);
     }
+
+    if (_sceneInputs.isNotEmpty) {
```
bdero, line 406: This binds the samplers but not the contract around them. `.fmat` emits a gate before each read (`frag_info.scene_inputs.x < 0.5` returns black, `.y < 0.5` returns 1e8) plus `GetScreenUv()` off `frag_info.ssao_params.zw`, and `FragInfo` is never bound for a `ShaderMaterial`.

Two consequences. A raw shader has no engine-provided way to compute the screen UV, and when an input is not produced this frame it samples `Material.whitePlaceholder`, so scene color reads white and `scene_depth` reads 1.0 world units. Both are actively wrong images rather than inert ones, and the depth case is easy to reach since `RenderInput.depth` quietly produces nothing on a non-perspective camera.

Binding a small block here when the set is non-empty (gates, inverse resolution, camera basis) and putting the accessors in a shared `scene_inputs.glsl` would make the two paths the same feature instead of two implementations of it.

packages/flutter_scene/lib/src/material/shader_material.dart
bdero: Checked this and I had it wrong. `MeshComponent._refreshRetainedMaterials` compares the old and new material's `sceneInputs` and calls `markMaterialSceneInputsChanged()` when they differ, so the swap paths that update render items already invalidate. No fix needed, sorry for the noise.

The follow-up (#349) still bumps the revision when a material is constructed with a non-empty set, which covers declaring before the first render rather than the swap case.

PR #352, opened 2026-08-22
Improve scene loading and static render performance

bdero:
Cache generated scene registries per asset bundle, compress fsceneb payload chunks when smaller, and skip unchanged refresh work for static mesh nodes. Version 1 fsceneb files remain readable.

### DIFF
--- a/packages/flutter_scene/CHANGELOG.md
+++ b/packages/flutter_scene/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## 0.24.0
 
 * Added `ExternalTexture`, a `TextureSource` fed by a platform texture id (video, camera preview), captured through the compositor on frames that sample it.
+* Generated scene registries and templates are cached per asset bundle.
+* Static render items and compressed `.fsceneb` payloads skip redundant work.
 * Widget-surface pointer forwarding dispatches events in the framework's global space (with the host's transform in the hit path), so widgets that convert `globalPosition` through render objects (Slider, text selection) work when the scene view is not at the window origin.
 * The standard PBR shader ships a no-shadow variant, selected automatically for draws with no shadow atlas bound, roughly halving the compiled fragment program for shadow-less scenes on mobile GPUs; it also skips its five per-texture UV-transform evaluations when every transform is identity.
 * `PhysicallyBasedMaterial` keeps a scalar `ior`, `specular`, and `specularColor` on the standard shader by folding them into its dielectric reflectance (the `KHR_materials_ior` / `KHR_materials_specular` formula the physical shader uses), so only specular textures and the layered features select the heavier physical variant.
--- a/packages/flutter_scene/CHANGELOG.md
+++ b/packages/flutter_scene/CHANGELOG.md
@@ -1,8 +1,10 @@
 ## 0.24.0
 
 * Added `ExternalTexture`, a `TextureSource` fed by a platform texture id (video, camera preview), captured through the compositor on frames that sample it.
-* Generated scene registries and templates are cached per asset bundle.
-* Static render items and compressed `.fsceneb` payloads skip redundant work.
+* Generated scene registries are cached per asset bundle.
+* `.fsceneb` payload chunks compress with gzip, cutting file sizes by ~73% in exchange for higher load-time decompression CPU.
+* Static render items skip per-frame property re-assignment when their transform and bounds are unchanged.
+* Instanced mesh static shadow updates track component visibility changes correctly.
 * Widget-surface pointer forwarding dispatches events in the framework's global space (with the host's transform in the hit path), so widgets that convert `globalPosition` through render objects (Slider, text selection) work when the scene view is not at the window origin.
 * The standard PBR shader ships a no-shadow variant, selected automatically for draws with no shadow atlas bound, roughly halving the compiled fragment program for shadow-less scenes on mobile GPUs; it also skips its five per-texture UV-transform evaluations when every transform is identity.
 * `PhysicallyBasedMaterial` keeps a scalar `ior`, `specular`, and `specularColor` on the standard shader by folding them into its dielectric reflectance (the `KHR_materials_ior` / `KHR_materials_specular` formula the physical shader uses), so only specular textures and the layered features select the heavier physical variant.
--- a/packages/flutter_scene/lib/src/components/instanced_mesh_component.dart
+++ b/packages/flutter_scene/lib/src/components/instanced_mesh_component.dart
@@ -55,28 +55,41 @@ class InstancedMeshComponent extends Component {
     if (item == null) return;
     // A material can declare itself draw-less for the frame (the shadow
     // catcher at zero intensity); its item then joins no pass at all.
-    item.visible = !item.material.drawsNothing;
+    final visible = !item.material.drawsNothing;
     final frustumCulled = node.frustumCulled;
-    final frustumCulledChanged = item.frustumCulled != frustumCulled;
-    item.frustumCulled = frustumCulled;
-    item.layers = node.layers;
     final lightChannelMask = node.lightChannelMask;
-    final worldTransform = node.globalTransform;
     final worldTransformVersion = node.worldTransformVersion;
     final instanceRevision = instancedMesh.revision;
     final geometryBoundsVersion = instancedMesh.geometry.localBoundsVersion;
+    if (node.shadowStatic &&
+        worldTransformVersion == _worldTransformVersion &&
+        instanceRevision == _instanceRevision &&
+        geometryBoundsVersion == _geometryBoundsVersion &&
+        item.visible == visible &&
+        item.frustumCulled == frustumCulled &&
+        item.layers == node.layers &&
+        item.shadowStatic == node.shadowStatic &&
+        item.castsShadows == node.castsShadows &&
+        item.lightChannelMask == lightChannelMask) {
+      return;
+    }
+    final frustumCulledChanged = item.frustumCulled != frustumCulled;
+    item.frustumCulled = frustumCulled;
+    item.layers = node.layers;
+    final worldTransform = node.globalTransform;
     final boundsChangedByInput =
         worldTransformVersion != _worldTransformVersion ||
         instanceRevision != _instanceRevision ||
         geometryBoundsVersion != _geometryBoundsVersion;
     final staticShadowChanged =
-        (item.visible != true ||
+        (item.visible != visible ||
             item.shadowStatic != node.shadowStatic ||
             item.castsShadows != node.castsShadows ||
             item.lightChannelMask != lightChannelMask ||
             boundsChangedByInput) &&
         (item.shadowStatic || node.shadowStatic) &&
         (item.castsShadows || node.castsShadows);
+    item.visible = visible;
     if (worldTransformVersion != _worldTransformVersion) {
       item.worldTransform.setFrom(worldTransform);
     }
--- a/packages/flutter_scene/lib/src/components/mesh_component.dart
+++ b/packages/flutter_scene/lib/src/components/mesh_component.dart
@@ -159,7 +159,6 @@ class MeshComponent extends Component {
     if (_renderItems.isEmpty) return;
     final worldTransformVersion = node.worldTransformVersion;
     var staticStateUnchanged =
-        node.shadowStatic &&
         node.skin == null &&
         node.internalMorphWeights == null &&
         worldTransformVersion == _worldTransformVersion;
@@ -171,6 +170,8 @@ class MeshComponent extends Component {
       final item = _renderItems[index];
       final primitive = _mesh.primitives[index];
       staticStateUnchanged =
+          item.jointsTexture == null &&
+          item.morphWeights == null &&
           item.visible == !item.material.drawsNothing &&
           item.primitiveVisible == primitive.visible &&
           item.frustumCulled == node.frustumCulled &&
--- a/packages/flutter_scene/lib/src/components/mesh_component.dart
+++ b/packages/flutter_scene/lib/src/components/mesh_component.dart
@@ -157,8 +157,34 @@ class MeshComponent extends Component {
   @internal
   void refreshRenderItems() {
     if (_renderItems.isEmpty) return;
-    final worldTransform = node.globalTransform;
     final worldTransformVersion = node.worldTransformVersion;
+    var staticStateUnchanged =
+        node.shadowStatic &&
+        node.skin == null &&
+        node.internalMorphWeights == null &&
+        worldTransformVersion == _worldTransformVersion;
+    for (
+      var index = 0;
+      staticStateUnchanged && index < _renderItems.length;
+      index++
+    ) {
+      final item = _renderItems[index];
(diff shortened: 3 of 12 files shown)


packages/flutter_scene/lib/src/importer/scene_registry.dart
```diff
@@ -38,6 +39,13 @@ final Map<String, Future<_SceneTemplate>> _sceneTemplates = {};
 /// from.
 final Map<String, int> _sceneTemplateHolders = {};
 
+/// One manifest-backed registry per asset bundle for the top-level helpers.
+/// Bundle identity is the cache boundary, and [Expando] lets a discarded
+/// test or application bundle release its registry with it.
+final Expando<Future<SceneRegistry>> _sceneRegistries = Expando(
+  'flutter_scene scene registries',
+);
```
bdero: This cache is never invalidated, so hot reload stops seeing manifest changes.

`HotReloadCoordinator._refresh()` calls `clearGeneratedAssetIndexCache()` for exactly this reason ("A regenerated source can repoint the manifest, so the cached index must not outlive a reload"). This is a cache over the same data and nothing clears it. `clearSceneTemplateCache()` does not either, so the only escape is a full restart.

Repro passes on master, fails here.

```dart
final bundle = _Bundle({a: _doc('a')});
await loadScene('assets/a', bundle: bundle);
bundle.assets[b] = _doc('b');
bundle.clear();
await loadScene('assets/b', bundle: bundle);
// Bad state: No generated .fsceneb for source "assets/b" was found.
```

A repointed key is the quieter version, resolving to a stale asset instead of throwing. Making this a non-final `var` and reassigning a fresh `Expando` from `clearSceneTemplateCache()` and from `_refresh()` covers both.

packages/flutter_scene/lib/src/importer/scene_registry.dart
```diff
@@ -693,11 +701,20 @@ Future<bool> releaseScene(
...
   final key = registry.resolveKey(sourcePath, package: package);
   return _releaseSceneTemplateClaim(key);
 }
 
+Future<SceneRegistry> _sharedSceneRegistry(AssetBundle? bundle) {
+  final assetBundle = bundle ?? rootBundle;
+  return _sceneRegistries[assetBundle] ??=
+      SceneRegistry.load(bundle: assetBundle).onError((error, stackTrace) {
+        _sceneRegistries[assetBundle] = null;
+        throw error!;
```
bdero: `throw error!` drops the original stack trace. `Error.throwWithStackTrace(error, stackTrace)` keeps it, and the `!` goes away with it since `onError<E extends Object>` infers a non-nullable `error`.

packages/scene/lib/src/binary/fsceneb.dart
```diff
@@ -162,6 +171,14 @@ SceneDocument readFsceneb(Uint8List bytes) {
...
         final (id, payload) = _decodeBlob(data);
         blobs[id] = payload;
+      case _chunkGzipBlob:
+        try {
+          final decompressed = const GZipDecoder().decodeBytes(data);
+          final (id, payload) = _decodeBlob(decompressed);
+          blobs[id] = payload;
+        } catch (error) {
+          throw FscenebFormatException('Invalid gzip payload chunk ($error)');
+        }
```
bdero, line 206: Worth stating the trade explicitly, because it runs the opposite direction from the PR title. Warm numbers on `content/catana_dojo/scene.neko_hero.007ff939.fsceneb`, desktop VM, native zlib.

| | v1 | v2 |
| --- | --- | --- |
| size | 49656 K | 13540 K (27%) |
| `readFsceneb` | 4 ms | 72 ms |
| `writeFsceneb` | n/a | 878 ms |

Across six scenes the size win held at 27% of original, which is a large, real win for bundle and download size. But `_readDocument` calls `readFsceneb` synchronously with no `Isolate.run` or `compute`, so that extra 68 ms lands on the UI isolate as a hitch, and a mobile CPU will be several times worse. Web is worse again, since `archive` falls back to its pure-Dart inflate whenever `dart.library.io` is absent (dart2wasm included).

The trade is easy to defend for a downloaded web build and hard to defend for a locally bundled asset, where the I/O it saves is close to free. `compressTextures` is already a default-off build hook flag, so a matching opt-in would fit the existing shape. At minimum the CHANGELOG should say load time gets worse.

bdero (commented):
Reviewed the three commits end to end. Full `flutter_scene` suite green (1923 pass, 38 skip), `scene` suite green, `dart analyze` clean, `dart format` clean on all eight changed Dart files. Findings below are verified against a real 50 MB `.fsceneb` and with a repro run on both branches.

Two things want fixing before this lands. The registry cache is never invalidated, which breaks scene hot reload (repro inline, passes on master). And the mesh early-out can strand a joints texture on an item, since the guard treats "no skin" as a reason to skip the code that clears it.

The container change is a real 73% size win, but it makes runtime loads roughly 18x slower on the UI isolate, which runs opposite to the PR title. That is a defensible trade for a downloaded build and a hard one for a locally bundled asset, so it is worth being explicit about, and possibly opt-in like `compressTextures`.

The instanced visibility fix is the best part of the third commit and deserves better billing than it gets. The mesh half of that commit I am less sure about, details inline.

packages/scene/lib/src/binary/fsceneb.dart
```diff
@@ -95,7 +98,13 @@ Uint8List writeFsceneb(SceneDocument document) {
...
       );
     }
-    addChunk(_chunkBlob, _encodeBlob(entry.key, bytes));
+    final blob = _encodeBlob(entry.key, bytes);
+    final compressed = const GZipEncoder().encodeBytes(blob, level: 6);
+    if (compressed.length < blob.length) {
+      addChunk(_chunkGzipBlob, compressed);
+    } else {
+      addChunk(_chunkBlob, blob);
+    }
```
bdero, line 131: Two things on the write side.

The library doc promises "two writes of the same document are identical" and the suite asserts it (`fsceneb_test.dart:68`), but that now only holds on the VM. `archive`'s pure-Dart encoder stamps `DateTime.now()` into the gzip MTIME field (`_gzip_encoder_web.dart:83`), so on web every write differs from the last, and native and web never agree byte for byte.

```
native header  [31, 139, 8, 0,  0,   0,   0,   0, ...]
web header     [31, 139, 8, 0, 58, 145, 139, 106, ...]
```

CI stays green since tests run on the VM, so this quietly weakens a guarantee the repo leans on elsewhere. Zeroing bytes 4 through 8 of each member fixes the timestamp half. The deflate streams still differ between zlib and the Dart implementation, so the cross-platform half of the claim needs narrowing either way.

Second, this compresses every payload and throws the result away when it does not help, including already-compressed PNG, KTX2 and Basis image payloads. Keying off `PayloadEncoding` would skip the pointless pass.

packages/flutter_scene/lib/src/components/mesh_component.dart
```diff
@@ -157,8 +157,34 @@ class MeshComponent extends Component {
   @internal
   void refreshRenderItems() {
     if (_renderItems.isEmpty) return;
-    final worldTransform = node.globalTransform;
     final worldTransformVersion = node.worldTransformVersion;
+    var staticStateUnchanged =
+        node.shadowStatic &&
+        node.skin == null &&
+        node.internalMorphWeights == null &&
+        worldTransformVersion == _worldTransformVersion;
```
bdero, line 164: This can strand a joints texture or morph weights on the item.

The guard requires `node.skin == null` and `node.internalMorphWeights == null`, but the slow path below is the only thing that ever *clears* `item.jointsTexture` and `item.morphWeights`. `Node.skin` is a plain public mutable field, so on a `shadowStatic` node with a settled transform, `node.skin = null` makes this guard pass from that frame on and the item keeps drawing with the last uploaded joint matrices permanently. Same shape when a mesh swap drops morph targets.

Adding `item.jointsTexture == null && item.morphWeights == null` to the per-item comparison closes both.

Everything else the slow path writes is covered. I checked `windingFlipped` specifically, and it is recomputed as a side effect of the `globalTransform` cache refresh that also bumps `_worldTransformVersion`, so the version check does cover it. `LodComponent` selects through `item.lod` in the encoder rather than here, so LOD switching is unaffected.

packages/flutter_scene/lib/src/components/mesh_component.dart
```diff
@@ -157,8 +157,34 @@ class MeshComponent extends Component {
...
+          item.layers == node.layers &&
+          item.lightChannelMask == node.lightChannelMask &&
+          item.shadowStatic == node.shadowStatic &&
+          item.castsShadows == (primitive.castsShadow && node.castsShadows) &&
+          item.highlightColor == node.highlightColor &&
+          _boundsVersions[index] == item.geometry.localBoundsVersion;
+    }
+    if (staticStateUnchanged) {
+      return;
+    }
```
bdero, line 187: Two structural notes on this early-out.

It restates by hand every field the loop below writes, with nothing keeping the two in sync. Miss one and it goes stale silently rather than failing, which is how the `jointsTexture` case above got through. Worth either deriving both from one place or adding a debug-mode assertion that runs the slow path anyway and compares.

The `node.shadowStatic` gate also looks like it costs more than it buys. It defaults to false, so this never fires for ordinary static level geometry, and it is not load-bearing for correctness since the version and equality checks here are the actual condition. What gets skipped is about fourteen plain field writes plus `refreshWinding`, which is itself a single assignment, and `internalRenderScene`, which is a plain field read. That is traded against ten field reads and comparisons in the guard. Do you have a profile showing this one positive? Dropping the gate would at least make it apply to the common case.

packages/flutter_scene/lib/src/components/instanced_mesh_component.dart
```diff
@@ -55,28 +55,41 @@ class InstancedMeshComponent extends Component {
...
     final staticShadowChanged =
-        (item.visible != true ||
+        (item.visible != visible ||
             item.shadowStatic != node.shadowStatic ||
             item.castsShadows != node.castsShadows ||
             item.lightChannelMask != lightChannelMask ||
             boundsChangedByInput) &&
         (item.shadowStatic || node.shadowStatic) &&
         (item.castsShadows || node.castsShadows);
+    item.visible = visible;
```
bdero, line 92: This is the genuinely valuable part of the commit and it is undersold in the CHANGELOG.

Moving the assignment below `staticShadowChanged` and comparing `item.visible != visible` instead of `item.visible != true` fixes two bugs at once. A permanently invisible static caster was re-dirtying the static shadow atlas every frame, and the invisible to visible transition was missed entirely. It also brings this in line with what `MeshComponent` already did.

Worth its own CHANGELOG line rather than being folded into a generic "skip redundant work".

The rest of the instanced guard checks out. `cullInstances` and `sortTransparentInstances` are final, and every mutating path on `InstancedMesh` bumps `_revision`, so the revision check does cover the instance fields the slow path assigns.

packages/flutter_scene/test/fscene/fsceneb_test.dart
```diff
@@ -78,6 +78,36 @@ void main() {
...
+        PayloadSpec(
+          id,
+          encoding: PayloadEncoding.vertexBuffer,
+          bytes: Uint8List(64 * 1024),
+        ),
+      );
+
+      final encoded = writeFsceneb(doc);
+      expect(encoded.length, lessThan(2048));
+      expect(String.fromCharCodes(encoded), contains('GZBL'));
```
bdero: `String.fromCharCodes(encoded)` scans the whole binary, so compressed bytes can trip this either way. Walking the chunk headers is both cheaper and exact. Same on line 104, where the negative form silently depends on `_sample()`'s payloads staying incompressible.

packages/flutter_scene/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 ## 0.24.0
 
 * Added `ExternalTexture`, a `TextureSource` fed by a platform texture id (video, camera preview), captured through the compositor on frames that sample it.
+* Generated scene registries and templates are cached per asset bundle.
+* Static render items and compressed `.fsceneb` payloads skip redundant work.
```
bdero: Both lines need work.

Templates were already cached, and by asset key rather than by bundle, so only the registry is new here.

The second line mashes two unrelated changes together, and compression is not "skipping redundant work", it is a size for CPU trade. Suggest splitting into a line for the static render item skip, a line for the instanced visibility fix, and a line for the container change that states the 73% size reduction and the slower load.

packages/scene/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## 0.3.0
 
+- `.fsceneb` format version 2 compresses payload chunks with gzip.
```
bdero: Worth adding that older readers reject a version 2 container. Bumping the version is the right call, since silently skipping `GZBL` chunks would drop payloads, and nothing published is affected while both packages sit in unreleased minors. One line here saves someone decoding the error later.

packages/flutter_scene/lib/src/importer/scene_registry.dart
bdero: Done in 329fec5f.

packages/flutter_scene/lib/src/importer/scene_registry.dart
bdero: Done in 329fec5f.

packages/scene/lib/src/binary/fsceneb.dart
bdero: Done in 329fec5f.

packages/scene/lib/src/binary/fsceneb.dart
bdero: Done in 329fec5f. Skips gzip for pre-compressed image payloads and zeros the gzip MTIME field.

packages/flutter_scene/lib/src/components/mesh_component.dart
bdero: Done in 329fec5f.

packages/flutter_scene/lib/src/components/mesh_component.dart
bdero: Done in 329fec5f. Dropped the shadowStatic gate so ordinary static meshes benefit from the early-out.

packages/flutter_scene/lib/src/components/instanced_mesh_component.dart
bdero: Done in 329fec5f.

packages/flutter_scene/test/fscene/fsceneb_test.dart
bdero: Done in 329fec5f. Tests now parse chunk header types directly.

packages/flutter_scene/CHANGELOG.md
bdero: Done in 329fec5f.

packages/scene/CHANGELOG.md
bdero: Done in 329fec5f.

PR #356, opened 2026-08-23
fix: slider widget should work in scenes when not at window origin (dispatch forwarded widget-surface input in global coordinates)

orestesgaolin:
Sliders on widget surfaces don't work when the `SceneView` is not at the window origin. Buttons and switches work fine, which makes this easy to miss.

**Disclaimer**: this PR is generated with LLM. Not sure what's the policy of AI-generated contributions. I'm happy to drop this and rather just use this PR as prompt to fix it yourself :). I read the code and PR and haven't found anything that I wouldn't accept myself.

https://github.com/user-attachments/assets/dc8c4aa3-8273-4f3b-9d6a-0c10f8c660e7

LLM generated content follows:

---

**What happens**

`WidgetTextureController` synthesizes pointer events with texture-space positions (0..width/height of the capture) and uses them as the event's position. That works for anything that only needs a hit test, like a button. But widgets that map `details.globalPosition` back through the render tree - `Slider` does `globalToLocal(globalPosition)`, text selection does the same - get garbage, because the "global" position they receive is actually local to the texture.

All the examples in this repo run `SceneView` fullscreen, so the host sits at `(0, 0)` and local happens to equal global. The bug only shows up when the view is inset - for example a scene view next to a sidebar. Then the slider jumps to 0 or ignores drags entirely, while buttons on the same panel keep working.

**The fix**

Dispatch events in the framework's global space instead:

- take the hosted child's `getTransformTo(null)` and apply it to the texture-space position
- push that transform into the hit path with `addWithPaintTransform`, so `localPosition` also stays correct
- capture the transform at pointer down and use it for the whole interaction, same as normal pointer capture
- bail out if the transform is not invertible

**Testing**

- New regression test: a `Slider` hosted at an offset `(padding 123, 217)`. Tap-to-seek and a two-step drag assert real values. It fails on the old code and passes with this change. Existing widget-texture tests still pass.
- New example, "Widget Input (inset view)": a side rail pushes the `SceneView` off the origin, and a panel with a slider, switch, and button sits on a `WidgetComponent` quad. Before this change the slider is broken there; after it, it tracks normally.
- Also verified in a separate app with an inset SceneView, dragging a Material slider on an in-scene panel.

_Changelog entry added under 0.24.0._

related to #170

### DIFF
--- a/WIDGETS.md
+++ b/WIDGETS.md
@@ -27,6 +27,25 @@ interaction lands on the widgets at exactly the point you see. Geometry in
 front of the panel blocks input, and a drag that slides off the panel's edge
 stays captured by it, so sliders and scrolls behave.
 
+## The child owns its state
+
+The component holds the widget it was built with, and `SceneView` mounts that
+subtree once. A `setState` in the widget that built the component does not
+rebuild it, so values pushed in that way never reach the surface: the hosted
+subtree keeps its own state and reports values out.
+
+```dart
+class ControlPanel extends StatefulWidget {
+  const ControlPanel({super.key, required this.onSpin});
+
+  final ValueChanged<double> onSpin;   // reports out to the scene
+  ...
+}
+```
+
+To drive the panel from outside, hand it something it can listen to (a
+`ValueNotifier`, an `AnimationController`) rather than a plain value.
+
 ## Bring your own surface
 
 Three tiers share the one component; each takes over a little more.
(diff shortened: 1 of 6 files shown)


bdero:
Thanks!

> Not sure what's the policy of AI-generated contributions

No policy, they are very much welcome. :) I've been thinking about writing a "policy" that says such. In the age of AI, I consider PRs to be enhanced issues, and I've taken to just patching in tests/fixes/conventions/whatever as needed & landing PRs that are directionally good.

orestesgaolin:
> In the age of AI, I consider PRs to be enhanced issues

Yeah, as long you have tokens available, it's sometimes easier to just receive an issue rather than PR. However, in this one I tried to validate that fix works and doesn't break anything else in the example app.

bdero:
Always feel welcome to make more PRs if you find issues or have ideas for improving things!

PR #357, opened 2026-08-24
Standardize model-space front faces on counter-clockwise winding

bdero:
Standardizes model-space front-facing triangle winding on counter-clockwise (CCW), matching glTF and standard conventions. Procedural primitives, normal generators, and the glTF importer now produce CCW triangles, and the render encoders bind clockwise on the Y-down rasterizer for unmirrored draws. Increments the .fscene document version to 5 and provides automatic index winding migration for version 4 and earlier scenes.

### DIFF
--- a/packages/flutter_scene/lib/src/fmat/build_materials.dart
+++ b/packages/flutter_scene/lib/src/fmat/build_materials.dart
@@ -8,6 +8,7 @@ import 'package:flutter_scene/src/importer/build_cache.dart';
 import 'package:flutter_scene/src/importer/build_hooks.dart'
     show discoveryDependencyDirectory;
 
+import '../generated_assets/engine_identity.dart' show engineIdentity;
 import '../generated_assets/generated_assets.dart';
 import '../generated_assets/generated_tree.dart';
 import 'fmat.dart';
@@ -105,24 +106,33 @@ List<String> discoverFmatMaterials(
 /// inside these are not tracked until `impellerc --depfile` is consumed in
 /// `--shader-bundle` mode (bdero/flutter_gpu_shaders#15).
 const _frameworkShaderFiles = <String>[
-  'material_varyings.glsl',
-  'pbr.glsl',
-  'texture.glsl',
-  'normals.glsl',
-  // Opt-in include for material authors (`#include <noise.glsl>`).
-  'noise.glsl',
-  'material_inputs.glsl',
+  'contact_shadow.glsl',
+  'depth_bias.glsl',
+  'depth_mask.glsl',
+  'filtered_scene_color.glsl',
+  'flutter_scene_morph.glsl',
+  'flutter_scene_skinned_body.glsl',
+  'flutter_scene_unskinned_body.glsl',
+  'flutter_scene_unskinned_depth_body.glsl',
+  'fog.glsl',
+  'interleaved_gradient_noise.glsl',
+  'lightmap.glsl',
+  'lod_fade.glsl',
   'material_engine_lighting.glsl',
+  'material_inputs.glsl',
   'material_lighting.glsl',
   'material_shadow_sampling.glsl',
-  'lightmap.glsl',
-  'filtered_scene_color.glsl',
-  // The vertex-stage includes a material's generated vertex variants pull in.
+  'material_varyings.glsl',
   'material_vertex.glsl',
-  'depth_bias.glsl',
-  'flutter_scene_unskinned_body.glsl',
-  'flutter_scene_skinned_body.glsl',
-  'flutter_scene_unskinned_depth_body.glsl',
+  'noise.glsl',
+  'normals.glsl',
+  'octahedral.glsl',
+  'pbr.glsl',
+  'scene_inputs.glsl',
+  'smaa.glsl',
+  'ssao_geometry.glsl',
+  'texture.glsl',
+  'tone_mapping.glsl',
 ];
 
 /// Compiles `.fmat` custom-material files into a Flutter GPU shader bundle plus
@@ -302,6 +312,7 @@ Future<void> _buildMaterials({
         .resolve('build/shaderbundles/$bundleName.shaderbundle')
         .toFilePath(),
   );
+  final variant = fileVariant ?? await engineIdentity();
   final shippedBundleFile = tree == null
       ? bundleFile
       : File.fromUri(
@@ -309,7 +320,7 @@ Future<void> _buildMaterials({
             GeneratedAssetFamily.material,
             nameId: bundleName,
             extension: '.shaderbundle',
-            variant: fileVariant,
+            variant: variant,
             target: target,
           ),
         );
@@ -318,7 +329,7 @@ Future<void> _buildMaterials({
           GeneratedAssetFamily.material,
           nameId: bundleName,
           extension: '.fmat.json',
-          variant: fileVariant,
+          variant: variant,
           target: target,
         ) ??
         packageRoot.resolve('build/shaderbundles/$bundleName.fmat.json'),
@@ -328,7 +339,7 @@ Future<void> _buildMaterials({
           GeneratedAssetFamily.material,
           nameId: bundleName,
           extension: '.index.json',
-          variant: fileVariant,
+          variant: variant,
           target: target,
         ) ??
         packageRoot.resolve('build/shaderbundles/$bundleName.index.json'),
--- a/packages/flutter_scene/lib/src/fscene/realize/resource_realizer.dart
+++ b/packages/flutter_scene/lib/src/fscene/realize/resource_realizer.dart
@@ -581,10 +581,36 @@ class ResourceRealizer {
     var indexType = gpu.IndexType.int16;
     final indexId = res.indices;
     if (indexId != null) {
-      indexBytes = ByteData.sublistView(_payloadBytes(indexId, 'index'));
-      indexType = document.payload(indexId)!.format == 'uint32'
-          ? gpu.IndexType.int32
-          : gpu.IndexType.int16;
+      final rawIndexBytes = _payloadBytes(indexId, 'index');
+      final isUint32 = document.payload(indexId)!.format == 'uint32';
+      indexType = isUint32 ? gpu.IndexType.int32 : gpu.IndexType.int16;
+      if (document.formatVersion < 5) {
+        final migrated = Uint8List.fromList(rawIndexBytes);
+        if (isUint32) {
+          final u32 = migrated.buffer.asUint32List(
+            migrated.offsetInBytes,
+            migrated.lengthInBytes ~/ 4,
+          );
+          for (var i = 0; i + 2 < u32.length; i += 3) {
+            final tmp = u32[i + 1];
+            u32[i + 1] = u32[i + 2];
+            u32[i + 2] = tmp;
+          }
+        } else {
+          final u16 = migrated.buffer.asUint16List(
+            migrated.offsetInBytes,
+            migrated.lengthInBytes ~/ 2,
+          );
+          for (var i = 0; i + 2 < u16.length; i += 3) {
+            final tmp = u16[i + 1];
+            u16[i + 1] = u16[i + 2];
+            u16[i + 2] = tmp;
+          }
+        }
+        indexBytes = ByteData.sublistView(migrated);
+      } else {
+        indexBytes = ByteData.sublistView(rawIndexBytes);
+      }
     }
 
     // Set baked bounds before upload so the position scan is skipped; without
--- a/packages/flutter_scene/lib/src/fscene/realize/resource_realizer.dart
+++ b/packages/flutter_scene/lib/src/fscene/realize/resource_realizer.dart
@@ -583,8 +583,8 @@ class ResourceRealizer {
     if (indexId != null) {
       final rawIndexBytes = _payloadBytes(indexId, 'index');
       final isUint32 = document.payload(indexId)!.format == 'uint32';
-      final isTriangleTopology =
-          res.topology == 'triangle' || res.topology == 'triangles';
+      indexType = isUint32 ? gpu.IndexType.int32 : gpu.IndexType.int16;
+      final isTriangleTopology = res.topology == 'triangle';
       if ((res.legacyWinding || document.formatVersion < 5) &&
           isTriangleTopology) {
         final migrated = Uint8List.fromList(rawIndexBytes);
@@ -611,6 +611,9 @@ class ResourceRealizer {
         }
         indexBytes = ByteData.sublistView(migrated);
       } else {
+        // TODO(winding): Support legacy winding migration for triangleStrip
+        // topology (requires vertex-order reversal, not index-triple swap).
+        // TODO(winding): Support non-indexed geometry winding migration.
         indexBytes = ByteData.sublistView(rawIndexBytes);
       }
     }
--- a/packages/flutter_scene/lib/src/fscene/realize/resource_realizer.dart
+++ b/packages/flutter_scene/lib/src/fscene/realize/resource_realizer.dart
@@ -584,7 +584,7 @@ class ResourceRealizer {
       final rawIndexBytes = _payloadBytes(indexId, 'index');
       final isUint32 = document.payload(indexId)!.format == 'uint32';
       indexType = isUint32 ? gpu.IndexType.int32 : gpu.IndexType.int16;
-      if (document.formatVersion < 5) {
+      if (res.legacyWinding || document.formatVersion < 5) {
         final migrated = Uint8List.fromList(rawIndexBytes);
         if (isUint32) {
           final u32 = migrated.buffer.asUint32List(
--- a/packages/flutter_scene/lib/src/fscene/realize/resource_realizer.dart
+++ b/packages/flutter_scene/lib/src/fscene/realize/resource_realizer.dart
@@ -583,8 +583,10 @@ class ResourceRealizer {
     if (indexId != null) {
       final rawIndexBytes = _payloadBytes(indexId, 'index');
       final isUint32 = document.payload(indexId)!.format == 'uint32';
-      indexType = isUint32 ? gpu.IndexType.int32 : gpu.IndexType.int16;
-      if (res.legacyWinding || document.formatVersion < 5) {
+      final isTriangleTopology =
+          res.topology == 'triangle' || res.topology == 'triangles';
+      if ((res.legacyWinding || document.formatVersion < 5) &&
+          isTriangleTopology) {
         final migrated = Uint8List.fromList(rawIndexBytes);
         if (isUint32) {
           final u32 = migrated.buffer.asUint32List(
--- a/packages/flutter_scene/lib/src/geometry/interleaved_layout.dart
+++ b/packages/flutter_scene/lib/src/geometry/interleaved_layout.dart
@@ -451,10 +451,10 @@ abstract final class InterleavedLayoutAdapter {
   /// triangle list; otherwise they are a non-indexed triangle list and
   /// [vertexCount] must be a multiple of three. Each triangle's
   /// unnormalized face normal is accumulated onto its three vertices, so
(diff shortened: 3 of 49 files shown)


bdero (commented):
Direction is right and the primitive flips plus the encoder mapping are internally consistent, but two silent correctness regressions need fixing first, and the rollout stops short of several places that carry the convention. The full suite passes and `dart analyze`/`dart format` are clean, so none of this is caught today.

The two blockers are inline. Runtime glTF import inverts, and the version 4 migration never runs.

The rest, which has no diff line to hang on:

`PolylineGeometry` was not flipped (`polyline_geometry.dart:181` and `:196`). Its strip orientation is deterministic (side vector is `tangent x viewDirection`, left at `2i`, right at `2i+1`) and the class has no `isDoubleSided` override, so a default back-face-culled material now culls it. `TrailComponent` and `LineSegmentsGeometry` are fine, `CullMode.none` and `isDoubleSided` respectively.

Material defaults still bind the old convention. `material.dart:457` and `preprocessed_material.dart:298` set `counterClockwise`, and `ShaderMaterial.windingOrder` defaults to it with a doc saying it matches glTF. Harmless in-engine since `_bindMaterial` nulls `_boundWindingOrder` and every path re-sets it, but the default now means mirrored and the rationale is inverted, which is a trap for `Material` subclasses and custom passes that bind and draw without re-setting.

Examples and smoke scenes were not converted. `example_widget_texture.dart:320` is the same dome construction updated in `interleaved_layout_test`, omits normals, and uses `PhysicallyBasedMaterial`, so it inverts culling and shading both. `smoke_scenes.dart:497` and `:1561` are the same, and that is the golden harness the release gate leans on. Also `example_nav_route`, `example_vertex_curve`, `example_raw_shader`, `example_particles`, `synthetic_morph_glb`.

No CHANGELOG entries in `flutter_scene` or `scene`, for a breaking authoring convention plus a format version bump. The version 3 to 4 bump got a line in `scene`'s changelog.

Nothing pins the new mapping. `render_item_winding_test` and `node_winding_test` only cover the XOR flag. A test that offline and runtime import agree on drawn orientation would catch the first blocker, and a version 4 realize test would catch the second. `gltf_coordinate_policy_test.dart:73` still asserts the old `sourceWindingFlipped` semantics and passes.

`GeometryBuilder.addTriangle` still documents nothing about winding, which is now the contract users need most.

Every index reversal checks out individually (cuboid, wedge, plane, sphere, cylinder and caps, capsule, torus, disc, ring, all 20 icosphere seed faces, swept stitching), and the importer supporting only glTF mode 4 makes the offline triple swap safe.

packages/flutter_scene/lib/src/importer/src/gltf/primitive_packer.dart
```diff
@@ -320,6 +320,13 @@ PackedPrimitive packGltfPrimitive({
       out[o + 16] = -out[o + 16];
       out[o + 17] = -out[o + 17];
     }
+    // Negating Z mirrors the triangle winding; swap indices (a, b, c) -> (a, c, b)
```
bdero: The runtime half of this flip is missing. `coordinate_policy.dart:34` still returns `sourceWindingFlipped == true` for `runtimeBoundary`, which meant the source triangles are opposite the native winding. Now that native is CCW, raw glTF indices match it and the getter has to be `false`.

Offline got both halves (this swap plus the new hardware mapping), runtime got only the hardware mapping. Packing the same source triangle through both policies, then applying each path's node transform, gives

```
PR head:  offline world normal z=-1.0  hw=clockwise   indices [0,2,1]
          runtime world normal z=+1.0  hw=clockwise   indices [0,1,2]

master:   offline world normal z=+1.0                 indices [0,1,2]
          runtime world normal z=+1.0                 indices [0,1,2]
```

Opposite world-space orientations under one front-face setting, where master had them agreeing. Every `.glb` loaded at runtime (`Node.fromAsset`, the most common entry point) is backwards.

packages/flutter_scene/lib/src/fscene/realize/resource_realizer.dart
```diff
@@ -581,10 +581,36 @@ class ResourceRealizer {
...
     final indexId = res.indices;
     if (indexId != null) {
-      indexBytes = ByteData.sublistView(_payloadBytes(indexId, 'index'));
-      indexType = document.payload(indexId)!.format == 'uint32'
-          ? gpu.IndexType.int32
-          : gpu.IndexType.int16;
+      final rawIndexBytes = _payloadBytes(indexId, 'index');
+      final isUint32 = document.payload(indexId)!.format == 'uint32';
+      indexType = isUint32 ? gpu.IndexType.int32 : gpu.IndexType.int16;
+      if (document.formatVersion < 5) {
```
bdero: This guard never fires, so the whole block is dead. `migrateFscene` stamps `json['fscene'] = version` as it walks the chain (`fscene_json.dart:162-171`) and `decodeDocument` reads that migrated value, so a version 4 document arrives here already reporting 5. Round-tripping one confirms it.

```
formatVersion after reading a v4 document: 5
realizer migration guard (formatVersion < 5) fires: false
```

`readFsceneb` goes through `readFscene` too, so both containers are affected and every pre-existing baked asset renders inside-out. The nastiest form is a stale hook-cached `.fsceneb` inverting while a fresh bake looks fine.

packages/scene/lib/src/json/fscene_json.dart
```diff
@@ -39,6 +39,10 @@ final List<FsceneMigration> _builtInMigrations = [
   // documents read as-is. The version exists so a version-3 reader refuses
   // a morph-bearing document instead of silently dropping the deltas.
   (json) => json,
+  // 4 -> 5 standardized model-space front-face winding to Counter-Clockwise
```
bdero, line 42: Swapping at realization rather than here has three problems, and moving it into `_migrateV4ToV5` fixes all three at once.

The editor reads and re-writes documents (`scene_io.dart:498`, `:615`, `:791`, `session.dart:119`), so opening a version 4 scene and saving it stamps `fscene: 5` onto unconverted indices, permanently and undetectably.

The realizer swaps every index triple, but `GeometryResource.topology` also allows `triangleStrip`, `line`, `lineStrip`, and `point`, and permuting triples corrupts all four. The offline swap in `primitive_packer.dart` is safe only because glTF mode 4 is the sole supported mode.

Non-indexed geometry is skipped entirely, since `GeometryResource.indices` is nullable and the block only runs when it is set.

packages/flutter_scene/lib/src/geometry/interleaved_layout.dart
```diff
@@ -451,15 +451,11 @@ abstract final class InterleavedLayoutAdapter {
   /// triangle list; otherwise they are a non-indexed triangle list and
   /// [vertexCount] must be a multiple of three. Each triangle's
   /// unnormalized face normal is accumulated onto its three vertices, so
-  /// larger faces contribute proportionally; the result is normalized
-  /// per vertex. Vertices touched by no (or only degenerate) triangles
-  /// receive a default normal of `(0, 0, 1)`.
+  /// Computes area-weighted vertex normals from triangle connectivity.
```
bdero: This landed mid-sentence. The line above still ends with `accumulated onto its three vertices, so`, and this one restates the summary line rather than finishing it.

It also dropped two behaviors the code still has, per-vertex normalization and the `(0, 0, 1)` default for vertices touched by no (or only degenerate) triangles.

packages/flutter_scene/skills/flutter_scene-idioms/SKILL.md
```diff
@@ -1,6 +1,6 @@
 ---
 name: flutter_scene-idioms
-version: 3
+version: 4
```
bdero, line 3: Right to bump here, but two other shipped skills still teach clockwise and were not touched.

`flutter_scene-procedural` (SKILL.md line 77, and `references/procedural.md` line 70, which also hands out the old index recipe) and `flutter_scene-verification-loop` (`references/loop.md` line 110). Both need the content fix and their own `version:` bump, or an installed copy is never detected as stale.

packages/flutter_scene/lib/src/geometry/primitives.dart
```diff
@@ -494,8 +494,8 @@ PrimitiveArrays buildCuboidArrays(Vector3 extents, {bool debugColors = false}) {
         colors[v * 4 + 3] = color[3];
       }
     }
-    // Two triangles matching the original winding: (a, b, d) and (d, b, c).
-    indices.addAll([base, base + 1, base + 3, base + 3, base + 1, base + 2]);
+    // Two CCW triangles matching the outward face normal: (a, d, b) and (d, c, b).
```
bdero: Nit, 83 columns. Same in the `primitive_packer.dart` swap comment and the `primitives_test.dart` helper doc.

packages/flutter_scene/lib/src/importer/src/gltf/primitive_packer.dart
bdero: Done in 8cbcaed4. GltfCoordinatePolicy.runtimeBoundary sets sourceWindingFlipped = false so the root node's Z = -1 reflection sets draw-time winding to counterClockwise.

packages/flutter_scene/lib/src/fscene/realize/resource_realizer.dart
bdero: Done in 8cbcaed4. _migrateV4ToV5 tags legacyWinding: true on geometry resources during document migration so resource_realizer.dart swaps index pairs on realization, covered by unit tests.

packages/scene/lib/src/json/fscene_json.dart
bdero: Done in 8cbcaed4 and 3183606d. GeometryResource tags legacyWinding: true in _migrateV4ToV5 and realizer checks for triangle topology before swapping.

packages/flutter_scene/lib/src/geometry/interleaved_layout.dart
bdero: Done in 8cbcaed4. Fixed truncated doc comment for generateNormals with per-vertex normalization and default fallback.

packages/flutter_scene/skills/flutter_scene-idioms/SKILL.md
bdero: Done in 8cbcaed4. Updated flutter_scene-procedural and flutter_scene-verification-loop skills with version bumps.

packages/flutter_scene/lib/src/geometry/primitives.dart
bdero: Done in 8cbcaed4. Fixed 80-column wrapping in primitives.dart, primitive_packer.dart, and primitives_test.dart.

packages/flutter_scene/lib/src/fscene/realize/resource_realizer.dart
```diff
@@ -581,10 +581,38 @@ class ResourceRealizer {
     var indexType = gpu.IndexType.int16;
     final indexId = res.indices;
     if (indexId != null) {
-      indexBytes = ByteData.sublistView(_payloadBytes(indexId, 'index'));
-      indexType = document.payload(indexId)!.format == 'uint32'
-          ? gpu.IndexType.int32
-          : gpu.IndexType.int16;
+      final rawIndexBytes = _payloadBytes(indexId, 'index');
+      final isUint32 = document.payload(indexId)!.format == 'uint32';
```
bdero, line 585: The `indexType` assignment got dropped when this block was reworked in 3183606d. At HEAD the variable is declared `int16` on line 581, read on 635, 659, 671, and 691, and never assigned from `isUint32`, so every realized geometry binds a 16-bit index buffer.

Any mesh over 65535 vertices ships uint32 indices and now draws as garbage. `isUint32` is still read inside the migration branch, so there is no unused-variable warning, and the test corpus is all 16-bit so nothing fails.

Already fixed in your working tree, just unpushed.

packages/scene/lib/src/json/fscene_json.dart
```diff
@@ -39,8 +39,24 @@ final List<FsceneMigration> _builtInMigrations = [
...
+  // so index pairs are swapped during realization.
+  _migrateV4ToV5,
 ];
 
+Map<String, dynamic> _migrateV4ToV5(Map<String, dynamic> json) {
+  final resources = json['resources'];
+  if (resources is Map) {
+    for (final res in resources.values) {
+      if (res is Map && res['kind'] == 'geometry' && res['indices'] != null) {
+        res['legacyWinding'] = true;
```
bdero, line 53: The flag is set here and read in `_decodeResource`, but `_encodeResource` never writes it. The geometry branch emits `kind`, `vertices`, `indices`, `bounds`, `topology`, and `morphTargets` only, so the round-trip hazard from the last pass is still live, just moved.

```
after migrating v4  -> legacyWinding=true version=5
saved JSON mentions legacyWinding: false
after reload        -> legacyWinding=false version=5
```

The editor opens and saves documents (`scene_io.dart:498`, `:615`, `:791`, `session.dart:119`), so one save drops the flag and the indices stay backwards with nothing left to signal it.

Encoding it closes the hole, but normalizing on write is better, swap the payload once and clear the flag, so a version 5 document never carries mixed winding.

packages/flutter_scene/skills/flutter_scene-verification-loop/SKILL.md
```diff
@@ -1,7 +1,7 @@
 ---
 name: flutter_scene-verification-loop
-version: 1
-description: Close the visual-iteration loop when building or debugging a flutter_scene 3D app so you see your own output and self-correct. Use whenever a change affects what renders (geometry, materials, lighting, shaders, post-processing) or a frame looks wrong (black, washed-out, see-through, missing geometry).
+version: 2
+description: Build flutter_scene content from code instead of asset files. Use when generating terrain, scattering vegetation or crowds, assembling modular kits, or driving a scene from noise and instancing rather than loading a .glb.
```
bdero: This description got overwritten with `flutter_scene-procedural`'s, verbatim. The verification-loop skill now advertises itself as building content from code rather than closing the visual-iteration loop.

Description text is what drives skill selection, so two shipped skills with identical descriptions means this one stops triggering for what it actually does, and the `version: 2` bump pushes it to installed copies. The old text is in the diff on the line above.

packages/flutter_scene/lib/src/fscene/realize/resource_realizer.dart
```diff
@@ -581,10 +581,38 @@ class ResourceRealizer {
...
+      final isTriangleTopology =
```
bdero: The topology guard is right, but the two cases it excludes are dropped silently and both need a `TODO(winding):`.

A version 4 `triangleStrip` geometry flipped its front faces the same way a triangle list did, and skipping the swap leaves it inside-out rather than correct. Reversing a strip needs a vertex-order reversal, not a triple swap, so it is real work worth naming.

Non-indexed geometry is never reached at all, since `_migrateV4ToV5` only flags resources with `indices != null` and this block only runs when `indexId != null`.

Minor, `res.topology == 'triangles'` never matches. `GeometryResource.topology` is documented as `triangle`, `triangleStrip`, `line`, `lineStrip`, or `point`.

packages/flutter_scene/lib/src/fmat/build_materials.dart
```diff
@@ -8,6 +8,7 @@ import 'package:flutter_scene/src/importer/build_cache.dart';
 import 'package:flutter_scene/src/importer/build_hooks.dart'
     show discoveryDependencyDirectory;
 
+import '../generated_assets/engine_identity.dart' show engineIdentity;
```
bdero, line 11: 8a4d9b2f (engine-identity isolation for material and shader bundle outputs, plus varying schema validation) has nothing to do with winding, is not mentioned in the PR description or either changelog, and carries the only remaining over-80 lines in the diff.

Worth splitting into its own PR so the winding change stays reviewable on its own.

packages/flutter_scene/lib/src/importer/src/gltf/coordinate_policy.dart
```diff
@@ -31,7 +31,7 @@ extension GltfCoordinateConversion on GltfCoordinatePolicy {
   bool get bakesNative => this == GltfCoordinatePolicy.bakeNative;
 
   /// Whether packed source triangles use the opposite native winding.
-  bool get sourceWindingFlipped => this == GltfCoordinatePolicy.runtimeBoundary;
+  bool get sourceWindingFlipped => false;
```
bdero, line 34: Correct fix, and it makes the getter a constant that ignores `this`. Once the answer is always false the whole `sourceWindingFlipped` path (here, `PackedPrimitive`, `geometry_builder.dart:50`, `Geometry.sourceWindingFlipped`) is dead weight. Either drop it or leave a note saying why it is kept for geometry the engine does not import.

bdero (commented):
Second pass. Most of round one is properly addressed. The runtime import inversion is fixed at the source and covered by a real index-swap test, the migration moved into `_migrateV4ToV5` behind a `legacyWinding` flag with a topology guard, `PolylineGeometry` and the material defaults are flipped, the skills and examples are converted, and both changelogs have entries. The example sites left alone (`smoke_scenes.dart:497`, `example_vertex_curve.dart:128`, `example_particles.dart:1779`) all already matched the new convention, so the conversion reads complete.

Three new problems, all inline, two of them introduced by this round. Suite is 1924 green and `dart analyze` is clean, so again none of this is caught.

Smaller things with no diff line to hang on:

`sourceWindingFlipped` is now inert. The extension getter is a constant, so `PackedPrimitive.sourceWindingFlipped` and `Geometry.sourceWindingFlipped` can never be true from any import path, and `geometry.dart:155` still documents it as something runtime glTF geometry sets. Worth deleting the whole path or correcting the doc.

The new coverage pins the packer's index swap but still not the invariant that actually broke, that runtime and offline import agree on the drawn orientation of the same source triangle. That is the test that would have caught it.

The directional-light test in `gltf_coordinate_policy_test.dart` was rewritten (different rotation, added intensity and `extensionsRequired`) for no winding reason.

Reviewed the pushed commits only. The working tree also carries unrelated in-flight work (async glTF importer entry points, `widget_component`, `sky_bake`, the stress-test example) alongside the `indexType` fix.

packages/flutter_scene/lib/src/fscene/realize/resource_realizer.dart
bdero: Done in 63c210ba. Assigned indexType from isUint32.

packages/scene/lib/src/json/fscene_json.dart
bdero: Done in 63c210ba. Encoded legacyWinding in _encodeResource so round-tripping preserves the migration flag.

packages/flutter_scene/skills/flutter_scene-verification-loop/SKILL.md
bdero: Done in 63c210ba. Restored original description for flutter_scene-verification-loop.

packages/flutter_scene/lib/src/fscene/realize/resource_realizer.dart
bdero: Done in 63c210ba. Corrected topology check to triangle and added TODOs for triangleStrip and non-indexed geometry migration.

packages/flutter_scene/lib/src/fmat/build_materials.dart
bdero: Acknowledged, will isolate unrelated engine-identity changes to a separate PR.

packages/flutter_scene/lib/src/importer/src/gltf/coordinate_policy.dart
bdero: Done in 63c210ba. Updated doc comment on Geometry.sourceWindingFlipped to document why it remains available for custom geometries.

PR #358, opened 2026-08-24
Add natural formation recipes and empirical verification rules to agent skills

bdero:
Updates flutter_scene-procedural (v2), flutter_scene-looks (v4), and flutter_scene-verification-loop (v2) with natural formation recipes, micro-surface metrics, and empirical verification disciplines.

### DIFF
--- a/packages/flutter_scene/CHANGELOG.md
+++ b/packages/flutter_scene/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 * Added `package:flutter_scene/kit.dart`, high-level gameplay, camera boom, character controller, day/night cycle, water surface, audio, pooling, and debug visualization components.
 * Standardized model-space front-face winding to Counter-Clockwise (CCW) across primitives, procedural builders, materials, and `.fscene` version 5.
+* Update `flutter_scene-procedural` skill (v2) with natural formation recipes (rock fractures, incised trails, pebble scatter, Gerstner waves, tree branching, and procedural island topographies).
+* Update `flutter_scene-looks` skill (v4) with micro-surface and anti-waxiness tuning guidance.
+* Update `flutter_scene-verification-loop` skill (v2) with empirical verification disciplines and blind-pairwise review rules.
 * Added `ExternalTexture`, a `TextureSource` fed by a platform texture id (video, camera preview) with optional `colorFilter`, captured through the compositor on frames that sample it.
 * Generated scene registries are cached per asset bundle.
 * `.fsceneb` payload chunks compress with gzip, cutting file sizes by ~73% in exchange for higher load-time decompression CPU.
--- a/packages/flutter_scene/skills/flutter_scene-looks/SKILL.md
+++ b/packages/flutter_scene/skills/flutter_scene-looks/SKILL.md
@@ -146,12 +146,12 @@ Start from the nearest look, then move one field at a time.
 
 ## Micro-surface metrics and anti-waxiness tuning
 
-When tuning procedural materials or custom shaders, use quantitative surface metrics to eliminate waxiness, plastic reads, or harsh aliasing:
+When tuning procedural materials or custom shaders, use surface metrics and exposure discipline to eliminate waxiness, plastic reads, or harsh aliasing:
 
-- **1-pixel luminance gradient (`grad/L`).** Measures micro-scale pixel energy `(|dL/dx| + |dL/dy|) / 2`. Natural weathered stone and terrain surfaces sit at `grad/L >= 0.025` to `0.085`; values under `0.010` read as smooth untextured plastic.
-- **High-frequency vs low-frequency ratio (`hf/lf`).** Compares 1px blur residuals to 4px residuals. A low ratio with high variance indicates blotchy macro-clouds rather than fine surface grain.
-- **Grazing terminator crossing fraction.** Under grazing directional lighting (sun below 20 degrees elevation), over-amplified normal maps create binary lit/unlit noise (pebble-dash or torn straw). Keep terminator crossing fractions bounded under 0.05.
-- **Tone-curve shoulder awareness.** Tone mapping curves (like ACES) compress channel differences near the shoulder. A surface reading high value (`V > 0.85`) with low saturation (`S < 0.40`) is over-exposed rather than under-pigmented; reducing exposure reveals natural material saturation without washing out channels.
+- **Surface reads like smooth plastic.** Increase high-frequency micro-scale variation. Low 1-pixel luminance gradients `(|dL/dx| + |dL/dy|) / 2` indicate untextured or overly smooth surfaces.
+- **Blotchy macro clouds.** Balance high-frequency and low-frequency energy ratios (`hf/lf`). High variance with low fine detail indicates large-scale noise patches without adequate surface grain.
+- **Harsh normal-map glitter or torn noise.** Check terminator crossing behavior under low grazing light angles (sun low on the horizon). Over-amplified normal maps flip adjacent pixels between full light and full shadow.
+- **Colors look washed out in bright areas.** Tone mapping curves compress channel differences near the shoulder. A surface reading high brightness with low saturation is often over-exposed rather than under-pigmented; reduce exposure to restore natural material saturation.
 
 ## Look tools that are not on EnvironmentSettings
 
--- a/packages/flutter_scene/skills/flutter_scene-looks/SKILL.md
+++ b/packages/flutter_scene/skills/flutter_scene-looks/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: flutter_scene-looks
-version: 3
+version: 4
 description: Give a flutter_scene render a deliberate, polished look. Use this whenever a scene looks flat, dull, or washed out, or whenever the ask is to make it look good, because a good look is lighting plus post-processing, not geometry. Ships copy-paste EnvironmentSettings presets that configure the whole stack coherently.
 ---
 
@@ -143,6 +143,15 @@ Start from the nearest look, then move one field at a time.
 - **Reads flat and ungrounded.** `ambientOcclusionEnabled: true`. Use `AmbientOcclusionMethod.groundTruth` for quality, keep `ambientOcclusionHalfResolution: true` for cost.
 - **Colors feel wrong.** Turn on `colorGradingEnabled` and reach for `saturation`, `contrast`, `temperature`, `tint` before anything else.
 - **Wrong tone-map feel.** `ToneMappingMode.aces` is contrasty and filmic, `pbrNeutral` (default) preserves hue and saturation, `agx` is the most neutral highlight rolloff. Do not reintroduce an `exposure: 2.0` hack; that was an artifact of an older renderer.
+
+## Micro-surface metrics and anti-waxiness tuning
+
+When tuning procedural materials or custom shaders, use quantitative surface metrics to eliminate waxiness, plastic reads, or harsh aliasing:
+
+- **1-pixel luminance gradient (`grad/L`).** Measures micro-scale pixel energy `(|dL/dx| + |dL/dy|) / 2`. Natural weathered stone and terrain surfaces sit at `grad/L >= 0.025` to `0.085`; values under `0.010` read as smooth untextured plastic.
+- **High-frequency vs low-frequency ratio (`hf/lf`).** Compares 1px blur residuals to 4px residuals. A low ratio with high variance indicates blotchy macro-clouds rather than fine surface grain.
+- **Grazing terminator crossing fraction.** Under grazing directional lighting (sun below 20 degrees elevation), over-amplified normal maps create binary lit/unlit noise (pebble-dash or torn straw). Keep terminator crossing fractions bounded under 0.05.
+- **Tone-curve shoulder awareness.** Tone mapping curves (like ACES) compress channel differences near the shoulder. A surface reading high value (`V > 0.85`) with low saturation (`S < 0.40`) is over-exposed rather than under-pigmented; reducing exposure reveals natural material saturation without washing out channels.
 
 ## Look tools that are not on EnvironmentSettings
 
--- a/packages/flutter_scene/skills/flutter_scene-procedural/SKILL.md
+++ b/packages/flutter_scene/skills/flutter_scene-procedural/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: flutter_scene-procedural
-version: 2
+version: 3
 description: Build flutter_scene content from code instead of asset files. Use when generating terrain, scattering vegetation or crowds, assembling modular kits, or driving a scene from noise and instancing rather than loading a .glb.
 ---
 
@@ -76,16 +76,20 @@ scene.add(terrain);
 
 If a hand-built surface renders inside-out (visible only from below, dark where lit), reverse each triangle's index order. flutter_scene's front faces wind Counter-Clockwise (CCW) in model space, matching glTF and standard conventions; never fix orientation with a per-triangle flip on an imported model, but for geometry you author yourself the winding is yours to set.
 
-## Natural formations and rock structures
+## Natural formations and landscape recipes
 
 To achieve documentary realism rather than generic procedural lumps:
 
-1. **Footpaths are scoured trenches, not flat stripes.** A real trail is the lowest line across terrain because water and foot traffic erode it downwards. When generating heightfields, cut the trail path profile down into the terrain with banks rising away on both sides. A path drawn as a flat texture reads as a sticker; a path you walk inside reads as a place.
+1. **Footpaths are scoured trenches, not flat stripes.** A real trail is the lowest line across terrain because water and foot traffic erode it downwards. When generating heightfields, cut the trail path profile down into the terrain with banks rising away on both sides.
 2. **Ridged noise for valley walls and cliffs.** Standard `FractalType.fbm` makes rolling mounds. Use `FractalType.ridged` for valley walls, mountain spurs, and cliffs to produce sharp erosion creases.
-3. **Free-end Worley rock cracks.** Standard Worley noise (`F2 - F1`) creates closed polygonal loops like dry mud or bathroom tile. To produce weathered rock fractures with natural free ends, multiply the cell border by a low-frequency region mask and a high-frequency grain breaker.
+3. **Free-end Worley rock cracks.** Standard Worley noise (`F2 - F1`) creates closed polygonal loops like bathroom tile. To produce weathered rock fractures with natural free ends, multiply the cell border by a low-frequency region mask and a high-frequency grain breaker.
 4. **Noise-modulated pitting.** A constant threshold radius across Worley cells places a pit in every cell, producing an artificial grid lattice. Modulate the threshold radius with an underlying Perlin field so pores vary in size and only appear in exposed weathering pockets.
 5. **Macro massing for scattered gravel.** Soil wears in 0.5m to 2m zones. Modulate multi-scale pebble instances with a low-frequency massing field so gravel clusters into realistic water scour lines rather than uniform sandpaper noise.
-6. **Sunk block settling and ground contact staining.** Place boulders and masonry courses 1/3 to 2/3 submerged into the sampled ground height. Use vertex colors or shader ground distance to stain the bottom 20cm of rock near the soil boundary, creating a smooth moisture transition instead of a sharp polygonal seam.
+6. **Sunk block settling and ground contact staining.** Place boulders and masonry courses 1/3 to 2/3 submerged into the sampled ground height. Use vertex colors or shader ground distance to stain the bottom 20cm of rock near the soil boundary, creating a smooth moisture transition.
+7. **Oceans and Gerstner waves.** Sum 4 to 8 directional Gerstner trochoidal waves that pull vertices horizontally toward crests, producing sharp peaks and wide flat troughs. Use Beer-Lambert depth absorption ($e^{-\sigma_a d}$) for turquoise to deep navy transitions, Jacobian folding for peak foam, and darken/smooth tidal sand within the shoreline wash.
+8. **Trees and foliage translucency.** Extrude branching splines preserving cross-sectional area ($d_{parent}^2 = \sum d_{child}^2$). Leaves require a double-sided transmission term in shader passes so backlit canopies glow instead of rendering as black silhouettes. Apply quadratic cantilever displacement ($\Delta \mathbf{P} \propto h^2$) for organic wind sway.
+9. **Procedural skies and IBL synchronization.** Combine Rayleigh scattering (blue zenith/red horizon) with Mie forward scattering (solar halo). Bake the procedural sky to an `EnvironmentMap` cubemap at load time with prefiltered radiance/irradiance so scene IBL matches the atmosphere automatically.
+10. **Islands and coastal erosion.** Multiply radial distance falloff with domain-warped FBM to form organic bays, sandbars, and lagoons. Use analytical surface slopes to strip topsoil on steep cliffs while depositing golden sand and reef shoals on shallow coastal planes.
 
 ## Scattering thousands of copies
 
@@ -127,4 +131,4 @@ Native platforms are unaffected. `noiseHash2`/`noiseHash3` are the bit-exact CPU
 
 ## More depth
 
-`references/procedural.md` has the full `GeometryBuilder` and `MeshData` API (including off-isolate meshing), the complete `FastNoiseLite` config reference, natural rock and terrain formation recipes, the instancing API in full, modular-kit assembly from the built-in primitives, and the web-noise caveat expanded.
+`references/procedural.md` has the full `GeometryBuilder` and `MeshData` API (including off-isolate meshing), the complete `FastNoiseLite` config reference, natural rock, ocean, tree, sky, and island formation recipes, the instancing API in full, modular-kit assembly from the built-in primitives, and the web-noise caveat expanded.
--- a/packages/flutter_scene/skills/flutter_scene-procedural/SKILL.md
+++ b/packages/flutter_scene/skills/flutter_scene-procedural/SKILL.md
@@ -10,7 +10,7 @@ A lot of 3D work does not need an artist's `.glb` at all. Terrain, scattered fol
 
 **The insight: for a code-driven scene, generate geometry and draw it instanced. That path is more reliable than loading external assets, because it has no import step, no coordinate-conversion traps, no missing-file failure modes, and one draw call for thousands of copies.** Three pieces cover almost everything:
 
-- **`GeometryBuilder`** (and the ten built-in primitives) build custom meshes without a model file.
+- **`GeometryBuilder`** (and the built-in primitives and swept paths) build custom meshes without a model file.
 - **`FastNoiseLite`** drives heightmaps, placement, and displacement deterministically.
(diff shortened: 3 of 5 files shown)


packages/flutter_scene/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## 0.24.0
 
+* Updated agent skills, adding 15 empirical verification rules to `flutter_scene-verification-loop` (v2), quantitative micro-surface tuning to `flutter_scene-looks` (v4), and natural rock, terrain carving, and gravel scatter recipes to `flutter_scene-procedural` (v2).
```
bdero: Three problems in one bullet.

The version is wrong. This says `flutter_scene-procedural` (v2) but the file in this PR is `version: 3`. Since the installed-copy staleness check keys off that field, the CHANGELOG and the skill have to agree.

It is also stale relative to the second commit. Ocean, tree, sky, and island recipes are the bulk of the diff and are not mentioned at all.

And it is one long run-on. The convention here is short punchy one-line bullets, so this reads better as one bullet per skill.

packages/flutter_scene/skills/flutter_scene-procedural/SKILL.md
```diff
@@ -1,6 +1,6 @@
 ---
 name: flutter_scene-procedural
-version: 1
+version: 3
```
bdero: Version jumps `1` -> `3` with no v2 in between. Harmless on its own (any bump makes an installed copy stale), but it is what makes the CHANGELOG entry wrong. Pick one number and make both files say it.

packages/flutter_scene/skills/flutter_scene-procedural/SKILL.md
```diff
@@ -76,6 +76,21 @@ scene.add(terrain);
 
 If a hand-built surface renders inside-out (visible only from below, dark where lit), reverse each triangle's index order. flutter_scene's front faces wind clockwise in model space; never fix orientation with a per-triangle flip on an imported model, but for geometry you author yourself the winding is yours to set.
 
+## Natural formations and landscape recipes
```
bdero, line 79: The skill's `description` is unchanged, and that is the only thing that decides whether this skill loads. It currently says "terrain, scattering vegetation or crowds, assembling modular kits, or driving a scene from noise and instancing" with no mention of water, sky, or trees. So after this PR the skill contains ocean and sky recipes that will never be reached by "build me an ocean" or "add a procedural sky". Widen the description alongside the body.

packages/flutter_scene/skills/flutter_scene-procedural/SKILL.md
```diff
@@ -76,6 +76,21 @@ scene.add(terrain);
...
+4. **Noise-modulated pitting.** A constant threshold radius across Worley cells places a pit in every cell, producing an artificial grid lattice. Modulate the threshold radius with an underlying Perlin field so pores vary in size and only appear in exposed weathering pockets.
+5. **Macro massing for scattered gravel.** Soil wears in 0.5m to 2m zones. Modulate multi-scale pebble instances with a low-frequency massing field so gravel clusters into realistic water scour lines rather than uniform sandpaper noise.
+6. **Sunk block settling and ground contact staining.** Place boulders and masonry courses 1/3 to 2/3 submerged into the sampled ground height. Use vertex colors or shader ground distance to stain the bottom 20cm of rock near the soil boundary, creating a smooth moisture transition.
+7. **Oceans and Gerstner waves.** Sum 4 to 8 directional Gerstner trochoidal waves that pull vertices horizontally toward crests, producing sharp peaks and wide flat troughs. Use Beer-Lambert depth absorption ($e^{-\sigma_a d}$) for turquoise to deep navy transitions, Jacobian folding for peak foam, and darken/smooth tidal sand within the shoreline wash.
```
bdero: LaTeX does not belong here. `$e^{-\sigma_a d}$` renders as literal source text in an agent's context window, no other skill file in the package uses it, and it costs tokens for nothing. Plain notation (`exp(-sigma_a * d)`) reads the same to the reader that matters. Same for the math on lines 90 and in the reference doc.

packages/flutter_scene/skills/flutter_scene-procedural/SKILL.md
```diff
@@ -76,6 +76,21 @@ scene.add(terrain);
...
+8. **Trees and foliage translucency.** Extrude branching splines preserving cross-sectional area ($d_{parent}^2 = \sum d_{child}^2$). Leaves require a double-sided transmission term in shader passes so backlit canopies glow instead of rendering as black silhouettes. Apply quadratic cantilever displacement ($\Delta \mathbf{P} \propto h^2$) for organic wind sway.
```
bdero: Worth naming the shipped classes here rather than "extrude branching splines". `TubeGeometry` sweeps a round cross-section along a `ScenePath` and `ExtrudeGeometry` sweeps an arbitrary closed profile, both exported from `scene.dart`. That is the whole branch-extrusion primitive, already in the box.

Also, the shipped `Material.doubleSided` flag covers the culling half of "double-sided transmission"; only the diffuse-transmission term needs a custom shader. Worth splitting so the reader does not think both halves are hand-rolled.

packages/flutter_scene/skills/flutter_scene-procedural/SKILL.md
```diff
@@ -76,6 +76,21 @@ scene.add(terrain);
...
+9. **Procedural skies and IBL synchronization.** Combine Rayleigh scattering (blue zenith/red horizon) with Mie forward scattering (solar halo). Bake the procedural sky to an `EnvironmentMap` cubemap at load time with prefiltered radiance/irradiance so scene IBL matches the atmosphere automatically.
```
bdero: This one is the most important. The engine already ships exactly this feature and the recipe tells the agent to rebuild it from scratch.

`PhysicalSkySource` (`lib/src/sky_sources.dart`) is an analytic single-scattering atmosphere with Rayleigh and Mie terms and an HDR sun disk, closed form with no ray march. `SkyEnvironment` assigned to `Scene.skyEnvironment` bakes it into the scene's IBL on a refresh policy (`manual`, `interval`, `everyFrame`), and `EnvironmentMap.fromSky` does a one-shot bake. Assigning the same source to `Scene.skybox` keeps the visible background and the lighting in sync automatically.

Teaching a from-scratch reimplementation of shipped API is the failure mode the skill maintenance rules call out. This item should point at those three names instead.

Separately, "cubemap ... with prefiltered radiance/irradiance" does not describe the shipped `EnvironmentMap`, which carries a prefiltered equirect radiance atlas plus SH-9 diffuse coefficients.

packages/flutter_scene/skills/flutter_scene-procedural/references/procedural.md
```diff
@@ -246,6 +246,160 @@ It bakes `getNoise2` over a `width` x `height` grid into a grayscale `Texture2D`
...
+
+Composing raw noise into realistic natural terrain and geology requires specific math patterns to avoid telltale procedural artifacts.
+
+### 1. Free-end Worley rock cracks (avoiding closed cell loops)
+
+Standard cellular Worley distance (`F2 - F1`) creates a continuous polygon network like bathroom tile or dry mud. To create natural weathering cracks with free ends, mask the cell borders with a low-frequency macro patch and a high-frequency grain breaker:
+
+```glsl
+// GLSL shader bake or .fmat surface
+vec2 cwl = NoiseCellular2(p * 3.0); // F1 in x, F2 in y
```
bdero: This snippet does not compile. The shipped signature is

```glsl
float NoiseCellular2(vec2 p, int seed, int distanceFunction, int returnType, float jitter)
```

so it returns a `float`, not a `vec2`, and takes five arguments. There is no "F1 in x, F2 in y" return anywhere in `noise.glsl`.

`F2 - F1` is already a first-class return type, `kNoiseCellularDistance2Sub`, so the manual subtraction on the next line is unnecessary once the call is fixed.

`NoiseFbm2` is wrong the same way. Real signature is `NoiseFbm2(vec2 p, int seed, int octaves, float lacunarity, float gain)`, so the `3` on lines 261 and 262 binds to `seed`, not octaves, and both calls are two arguments short.

packages/flutter_scene/skills/flutter_scene-procedural/references/procedural.md
```diff
@@ -246,6 +246,160 @@ It bakes `getNoise2` over a `width` x `height` grid into a grayscale `Texture2D`
...
+float breaker = smoothstep(0.32, 0.54, NoiseFbm2(p * 6.0, 2));
+float crack = net * region * breaker; // produces isolated segments with natural start/end points
+```
+
+### 2. Noise-modulated pitting (avoiding regular dot lattices)
+
+Thresholding Worley noise at a constant radius places a pit in every single cell, creating an artificial grid lattice. Modulate the threshold radius with an underlying Perlin field so pores vary in size and only appear in exposed weathering pockets:
+
+```glsl
+float sizeVar = NoiseFbm2(p * 2.5, 3);
```
bdero: Same two signature errors as the block above (`NoiseFbm2` needs seed/octaves/lacunarity/gain, `NoiseCellular2` returns a `float` and takes five arguments).

There is also a range problem underneath the arity one. `noise.glsl` documents every `Noise*` float function as roughly `[-1, 1]`, and this reference file says the same thing for the Dart side a few sections up. So `smoothstep(0.40, 0.68, ...)` and the `0.46`/`0.62` pair on line 261 are sampling only the upper tail of the distribution, and the masks will be far sparser than the prose describes. The thresholds need rechecking against the real range once the calls are fixed.

bdero (commented):
Reviewed the whole diff and checked every API name and signature in the new snippets against the source on this branch.

The direction is good, the natural-formation recipes are the kind of thing that makes the difference between a procedural scene reading as documentary and reading as noise. But the code in the new reference sections is not accurate to the shipped API, which is the one failure mode the skill guidelines single out as worse than shipping nothing.

Five of the new snippets will not compile. `NoiseCellular2` returns a `float` and takes five arguments (not a `vec2` from one), `NoiseFbm2` takes seed/octaves/lacunarity/gain (not two args), `noiseHash2` takes three arguments (not two), and `ScenePath.distanceTo` does not exist on a class whose curve is 3D anyway.

One recipe reimplements a shipped feature. `PhysicalSkySource` plus `SkyEnvironment` already give an analytic Rayleigh/Mie sky auto-baked into IBL on a refresh policy, so recipe 9 teaches the agent to hand-roll something it can get in two lines. Two others (Beer-Lambert depth, branch extrusion) approximate around `RenderInput.depth` and `TubeGeometry`/`ExtrudeGeometry`.

The rest is smaller. The CHANGELOG says procedural v2 while the file says v3 and omits the second commit's contents entirely, the procedural `description` was not widened even though the scope now covers oceans and skies (so the skill will not trigger for them), and a few of the verification rules carry private-harness specifics and an audio rule into a consumer-facing rendering skill.

Details inline.

packages/flutter_scene/skills/flutter_scene-procedural/references/procedural.md
```diff
@@ -246,6 +246,160 @@ It bakes `getNoise2` over a `width` x `height` grid into a grayscale `Texture2D`
...
+```
+
+### 3. Incised trail heightfields (scours vs flat stripes)
+
+Footpaths are formed by water and foot traffic compressing and eroding soil downwards. Cut the path profile into the terrain heightfield as an incised trough with raised spoil banks:
+
+```dart
+double computeTerrainHeight(double x, double z, FastNoiseLite noise, ScenePath trail) {
+  final baseHeight = noise.getNoise2(x, z) * 8.0;
+  final dist = trail.distanceTo(vm.Vector2(x, z));
```
bdero: `ScenePath` has no `distanceTo`, and it is a 3D curve, so a `Vector2` cannot be passed to it either. Its public surface is `positionAt`, `tangentAt`, `frameAt`, `positionAtDistance`, `frameAtDistance`, `parameterAtDistance`, `length`, and `sample`.

This matters more than a name fix, because closest-point-to-path is the genuinely hard part of the recipe and it is the part being elided. The honest version is sample the path once (`sample(n, evenlySpaced: true)`) and take the min point-to-segment distance over the polyline, hoisted out of the per-vertex loop. Otherwise the reader hits a wall immediately after copying this.

packages/flutter_scene/skills/flutter_scene-procedural/references/procedural.md
```diff
@@ -246,6 +246,160 @@ It bakes `getNoise2` over a `width` x `height` grid into a grayscale `Texture2D`
...
+
+Gravel and pebbles cluster into water-washed scour lines rather than spreading evenly over an entire level. Gate multi-scale pebble instances with a low-frequency macro massing field:
+
+```dart
+void scatterPebbles(InstancedMesh finePebbles, InstancedMesh largeStones, FastNoiseLite terrainNoise) {
+  final macroNoise = FastNoiseLite(seed: 42)..frequency = 0.05;
+  for (var x = 0.0; x < 64.0; x += 0.8) {
+    for (var z = 0.0; z < 64.0; z += 0.8) {
+      // Deterministic coordinate jitter
+      final h = noiseHash2(x.round(), z.round());
```
bdero: `noiseHash2` takes three arguments, `noiseHash2(int seed, int x, int y)`. This call passes two, so it will not compile.

The rounding is also lossy in a way that shows up on screen. The loop steps by `0.8`, so `1.6.round()` and `2.4.round()` are both `2`, and roughly every third cell reuses its neighbour's hash. That gives repeated jitter offsets and duplicated fine/large stone picks on a visible periodic pattern, which is the exact artifact this recipe exists to avoid. Key the hash off integer loop indices instead of rounding the world position.

packages/flutter_scene/skills/flutter_scene-procedural/references/procedural.md
```diff
@@ -246,6 +246,160 @@ It bakes `getNoise2` over a `width` x `height` grid into a grayscale `Texture2D`
...
+
+```glsl
+// GLSL Gerstner wave displacement
+struct Wave { vec2 dir; float amp; float freq; float speed; float steepness; };
+
+vec3 evaluateGerstner(vec2 pos, float time, Wave w, inout vec3 tangent, inout vec3 binormal) {
+  float phase = dot(w.dir, pos) * w.freq + time * w.speed;
+  float c = cos(phase);
+  float s = sin(phase);
+  float q = w.steepness / (w.amp * w.freq * 4.0);
```
bdero: Two issues with the Gerstner block.

The `4.0` hardcodes the wave count into the steepness normalization, but both the prose above and item 7 in SKILL.md say to sum four to eight waves. At eight the per-wave `Q` is double what it should be, the summed steepness passes 1, and the crests self-intersect and loop. Make it a `numWaves` parameter.

`tangent` and `binormal` are `inout` accumulators that are only correct if the caller seeds them with `(1, 0, 0)` and `(0, 0, 1)`. The identity terms from the standard derivation are missing here and the requirement is not stated anywhere, so a caller who zero-initializes gets a degenerate normal and flat-looking water. Either seed them inside a wrapper or say so in the prose.

Minor on top of that, the two names are swapped relative to the usual derivation (what is called `tangent` here is the x-direction binormal), and `w.dir` has to be unit length for any of the amplitudes to mean what they say.

packages/flutter_scene/skills/flutter_scene-procedural/references/procedural.md
```diff
@@ -246,6 +246,160 @@ It bakes `getNoise2` over a `width` x `height` grid into a grayscale `Texture2D`
...
+                   -q * w.dir.y * w.dir.y * w.amp * w.freq * s);
+
+  return vec3(q * w.amp * w.dir.x * c,
+              w.amp * s,
+              q * w.amp * w.dir.y * c);
+}
+```
+
+For shallow water transitions and shorelines:
+- **Beer-Lambert Depth Extinction**: Compute depth $d = y_{surface} - y_{floor}$. Attenuate color with $\mathbf{C} = \mathbf{C}_{deep} + (\mathbf{C}_{shallow} - \mathbf{C}_{deep}) e^{-\boldsymbol{\sigma}_a d}$.
```
bdero: `d = y_surface - y_floor` is a geometric approximation that breaks on sloped seabeds and at grazing view angles, and the engine already has the right input for this.

`RenderInput.depth` binds the linear scene depth of the opaque geometry and is declared from a `.fmat` with `engine_inputs:`. `Material.sceneInputs` documents it as existing specifically for "refraction, depth-fade absorption, shoreline foam, and soft-particle style effects on translucent surfaces", which is this bullet and the tidal one below it. Point at that seam.

packages/flutter_scene/skills/flutter_scene-verification-loop/SKILL.md
```diff
@@ -36,6 +36,26 @@ Instead, **compare two frames and return a binary pick.** Put the new frame next
 
 If you have no reference at all, say so and describe the concrete difference between the two frames (this one is brighter here, that one has an artifact there) rather than inventing a score.
 
+## Empirical verification rules
+
+These fifteen hard-won rules prevent false diagnoses, hollow passes, and measurement traps during visual iteration:
```
bdero: "These fifteen" hardcodes the count in prose. Add or drop a rule and the sentence is silently wrong. Just say "These rules".

packages/flutter_scene/skills/flutter_scene-verification-loop/SKILL.md
```diff
@@ -36,6 +36,26 @@ Instead, **compare two frames and return a binary pick.** Put the new frame next
...
+
+1. **Liveness before ablation.** A negative result is evidence only if the removed or modified term was actually live in the draw pass. Diff pixels for liveness before trusting an ablation.
+2. **Tools must fail loud on empty measurements.** A tool that measures zero pixels, NaNs, or empty data must throw an error with exit code 2 and `NO MEASUREMENT` rather than printing a default passing number.
```
bdero: `exit code 2` and the literal string `NO MEASUREMENT` are a contract from a specific private harness. A consumer reading this skill has no such tool and no way to satisfy the rule as written.

The underlying principle is good and worth keeping, so state it generally. A measurement taken over an empty population must fail rather than return a default, and a tool that prints a passing number for zero pixels is worse than no tool.

packages/flutter_scene/skills/flutter_scene-verification-loop/SKILL.md
```diff
@@ -36,6 +36,26 @@ Instead, **compare two frames and return a binary pick.** Put the new frame next
...
+9. **Symmetric domain clamping.** Clamping parametric lookup domains must be verified at both boundaries, and derived slope or heading accessors must clamp both sample points.
+10. **Multi-band octave tables for tiling.** Two-band ratios like `hf/lf` are blind to regular mid-frequency patterns; inspect multi-octave energy tables to detect periodic tiling.
+11. **Physical discrimination over naming.** A class-discriminating physical observation (such as missing shadow terminators) outranks code comments or variable names.
+12. **Grazing light terminator crossing.** Under low-angle grazing lighting, prioritize terminator-crossing fractions over slope RMS to detect normal map over-amplification.
+13. **Resolution scaling.** High-frequency energy (`hf/lf`) and relative contrast scale with resolution; compare them only at equal pixel resolutions.
+14. **Deterministic audio probing.** Validate procedural audio graphs headlessly by evaluating offline PCM RMS distributions, quiet floors, and inter-channel decorrelation.
```
bdero: Audio is out of scope for this skill. The description scopes it to "a change affects what renders (geometry, materials, lighting, shaders, post-processing)", and flutter_scene has no audio API at all, so a consumer has nothing to apply this to. It reads as leakage from an internal project.

Rule 5's `git status` instruction has a bit of the same flavour, though the "look at the raw frame before quoting numbers off it" half of it is genuinely general and worth keeping.

packages/flutter_scene/skills/flutter_scene-looks/SKILL.md
```diff
@@ -144,6 +144,15 @@ Start from the nearest look, then move one field at a time.
 - **Colors feel wrong.** Turn on `colorGradingEnabled` and reach for `saturation`, `contrast`, `temperature`, `tint` before anything else.
 - **Wrong tone-map feel.** `ToneMappingMode.aces` is contrasty and filmic, `pbrNeutral` (default) preserves hue and saturation, `agx` is the most neutral highlight rolloff. Do not reintroduce an `exposure: 2.0` hack; that was an artifact of an older renderer.
 
+## Micro-surface metrics and anti-waxiness tuning
+
+When tuning procedural materials or custom shaders, use quantitative surface metrics to eliminate waxiness, plastic reads, or harsh aliasing:
+
+- **1-pixel luminance gradient (`grad/L`).** Measures micro-scale pixel energy `(|dL/dx| + |dL/dy|) / 2`. Natural weathered stone and terrain surfaces sit at `grad/L >= 0.025` to `0.085`; values under `0.010` read as smooth untextured plastic.
```
bdero: These thresholds are stated to three decimal places but nothing in the package computes them, and the section does not say how to. An agent reading this will cite `grad/L >= 0.025` as if it had measured it.

It also sits awkwardly next to the verification skill this PR is updating in the same change, whose rule 3 demands the population window, crop rectangle, threshold, and resolution be quoted beside any figure, and whose rule 13 says `hf/lf` is only comparable at equal resolution. These numbers quote none of that, so they fail the discipline being added a few files over.

Either ship the measurement recipe (resolution, crop, the actual formula over which channel) or keep the qualitative advice and drop the specific numbers.

packages/flutter_scene/skills/flutter_scene-looks/SKILL.md
```diff
@@ -144,6 +144,15 @@ Start from the nearest look, then move one field at a time.
...
+- **High-frequency vs low-frequency ratio (`hf/lf`).** Compares 1px blur residuals to 4px residuals. A low ratio with high variance indicates blotchy macro-clouds rather than fine surface grain.
+- **Grazing terminator crossing fraction.** Under grazing directional lighting (sun below 20 degrees elevation), over-amplified normal maps create binary lit/unlit noise (pebble-dash or torn straw). Keep terminator crossing fractions bounded under 0.05.
+- **Tone-curve shoulder awareness.** Tone mapping curves (like ACES) compress channel differences near the shoulder. A surface reading high value (`V > 0.85`) with low saturation (`S < 0.40`) is over-exposed rather than under-pigmented; reducing exposure reveals natural material saturation without washing out channels.
```
bdero: The `- **Label**: text` shape here and in the reference doc is the `Label: explanation` form the repo conventions steer away from, and it is inconsistent with the surrounding bullets in this very file, which use `- **Colors feel wrong.** Turn on ...`. Worth matching the local style.

packages/flutter_scene/CHANGELOG.md
bdero: Split into 3 concise bullets with matching version numbers (v2/v4/v2) in `73a0ee04`.

packages/flutter_scene/skills/flutter_scene-procedural/SKILL.md
bdero: Set version to 2 in `73a0ee04`.

packages/flutter_scene/skills/flutter_scene-procedural/SKILL.md
bdero: Widened description to include oceans, Gerstner waves, procedural skies, and trees in `73a0ee04`.

packages/flutter_scene/skills/flutter_scene-procedural/SKILL.md
bdero: Replaced LaTeX syntax with plain notation in `73a0ee04`.

packages/flutter_scene/skills/flutter_scene-procedural/SKILL.md
bdero: Named `TubeGeometry` and `ExtrudeGeometry`, and split `Material.doubleSided` from custom transmission shading in `73a0ee04`.

packages/flutter_scene/skills/flutter_scene-procedural/SKILL.md
bdero: Pointed to `PhysicalSkySource`, `SkyEnvironment`, `EnvironmentMap.fromSky`, and `Scene.skybox` in `73a0ee04`.

packages/flutter_scene/skills/flutter_scene-procedural/references/procedural.md
bdero: Fixed both signatures and used `kNoiseCellularDistance2Sub` in `73a0ee04`.

packages/flutter_scene/skills/flutter_scene-procedural/references/procedural.md
bdero: Corrected signatures and re-centered thresholds for the `[-1, 1]` range in `73a0ee04`.

packages/flutter_scene/skills/flutter_scene-procedural/references/procedural.md
bdero: Sampled the path polyline and computed minimum point-to-segment distance in `73a0ee04`.

packages/flutter_scene/skills/flutter_scene-procedural/references/procedural.md
bdero: Passed 3 arguments and keyed hash off integer grid cell indices in `73a0ee04`.

packages/flutter_scene/skills/flutter_scene-procedural/references/procedural.md
bdero: Parameterized `numWaves`, normalized direction, and documented accumulator seed requirements in `73a0ee04`.

packages/flutter_scene/skills/flutter_scene-procedural/references/procedural.md
bdero: Pointed to `RenderInput.depth` via `Material.sceneInputs` in `73a0ee04`.

packages/flutter_scene/skills/flutter_scene-verification-loop/SKILL.md
bdero: Reworded to "These rules" in `73a0ee04`.

packages/flutter_scene/skills/flutter_scene-verification-loop/SKILL.md
bdero: Generalized to require loud failure on empty populations or non-finite pixels in `73a0ee04`.

packages/flutter_scene/skills/flutter_scene-verification-loop/SKILL.md
bdero: Removed audio rule and kept visual verification rules general in `73a0ee04`.

packages/flutter_scene/skills/flutter_scene-looks/SKILL.md
bdero: Dropped specific numbers while retaining qualitative and formulaic guidance in `73a0ee04`.

packages/flutter_scene/skills/flutter_scene-looks/SKILL.md
bdero: Converted to `- **Problem.** Action...` format in `73a0ee04`.

PR #359, opened 2026-08-24
Add gameplay and ergonomics kit and agent skill

bdero:
Adds `package:flutter_scene/kit.dart` providing high-level gameplay components (spring arm camera boom, camera shake, bounds framing, virtual joystick, third-person controller, steering behaviors, day/night atmospheric cycle, water surface, audio pitch jitter, node pooling, and debug draw utilities) and ships the companion `flutter_scene-kit` agent skill.

### DIFF
--- a/examples/flutter_app/lib/example_kit.dart
+++ b/examples/flutter_app/lib/example_kit.dart
@@ -710,7 +710,8 @@ class _KitStageState extends State<_KitStage> {
       jumpVelocity: widget.settings.jumpVelocity,
       groundPlaneHeight: 0.0,
       footOffset: 0.9,
-      obstacleRadius: 0.5,
+      obstacleRadius: 0.85,
+      obstacleHeight: 1.8,
     );
     _characterNode!.addComponent(_characterController!);
 
--- a/examples/flutter_app/lib/example_kit.dart
+++ b/examples/flutter_app/lib/example_kit.dart
@@ -1,3 +1,4 @@
+import 'dart:async';
 import 'dart:math' as math;
 
 import 'package:flutter/material.dart' hide Material;
@@ -70,6 +71,7 @@ class _ExampleKitState extends State<ExampleKit> {
             scenario: _scenario,
             settings: _settings,
             onFireBurst: () => _stageKey.currentState?.fireProjectileBurst(),
+            onScatterFlock: () => _stageKey.currentState?.scatterFlock(),
             onChanged: () => setState(() {}),
           ),
         ),
@@ -78,25 +80,52 @@ class _ExampleKitState extends State<ExampleKit> {
   }
 }
 
-class _KitScenarioPanel extends StatelessWidget {
+class _KitScenarioPanel extends StatefulWidget {
   const _KitScenarioPanel({
     required this.scenario,
     required this.settings,
     required this.onFireBurst,
+    required this.onScatterFlock,
     required this.onChanged,
   });
 
   final _KitScenario scenario;
   final KitDemoSettings settings;
   final VoidCallback onFireBurst;
+  final VoidCallback onScatterFlock;
   final VoidCallback onChanged;
+
+  @override
+  State<_KitScenarioPanel> createState() => _KitScenarioPanelState();
+}
+
+class _KitScenarioPanelState extends State<_KitScenarioPanel> {
+  Timer? _uiTimer;
+
+  @override
+  void initState() {
+    super.initState();
+    _uiTimer = Timer.periodic(const Duration(milliseconds: 60), (_) {
+      if (mounted &&
+          widget.scenario == _KitScenario.dayNight &&
+          widget.settings.isTimePlaying) {
+        setState(() {});
+      }
+    });
+  }
+
+  @override
+  void dispose() {
+    _uiTimer?.cancel();
+    super.dispose();
+  }
 
   @override
   Widget build(BuildContext context) {
     return ExamplePanelCard(
       icon: Icons.sports_esports,
-      title: scenario.label,
-      maxBodyHeight: 400,
+      title: widget.scenario.label,
+      maxBodyHeight: 440,
       body: Column(
         mainAxisSize: MainAxisSize.min,
         crossAxisAlignment: CrossAxisAlignment.start,
@@ -106,7 +135,10 @@ class _KitScenarioPanel extends StatelessWidget {
   }
 
   List<Widget> _buildControls(BuildContext context) {
-    switch (scenario) {
+    final settings = widget.settings;
+    final onChanged = widget.onChanged;
+
+    switch (widget.scenario) {
       case _KitScenario.characterCamera:
         return [
           const KitSectionHeader('Character'),
@@ -135,7 +167,7 @@ class _KitScenarioPanel extends StatelessWidget {
             label: 'Arm length',
             value: settings.armLength,
             min: 2.0,
-            max: 10.0,
+            max: 12.0,
             onChanged: (v) {
               settings.armLength = v;
               onChanged();
@@ -153,14 +185,59 @@ class _KitScenarioPanel extends StatelessWidget {
           ),
           const SizedBox(height: 8),
           const Text(
-            'Use WASD / Arrow keys and Space to jump, or the on-screen joystick.',
+            'Drag on screen to orbit camera freely. WASD / Arrow keys run relative to camera view.',
             style: TextStyle(color: Colors.white70, fontSize: 11),
           ),
         ];
 
       case _KitScenario.dayNight:
+        final hours = settings.timeOfDay.floor();
+        final minutes = ((settings.timeOfDay - hours) * 60).round();
+        final timeStr =
+            '${hours.toString().padLeft(2, '0')}:${minutes.toString().padLeft(2, '0')}';
+
         return [
           const KitSectionHeader('Time of Day'),
+          Row(
+            children: [
+              Expanded(
+                child: Text(
+                  'Clock: $timeStr',
+                  style: const TextStyle(
+                    color: Colors.amberAccent,
+                    fontSize: 12,
+                    fontWeight: FontWeight.bold,
+                  ),
+                ),
+              ),
+              ElevatedButton.icon(
+                style: ElevatedButton.styleFrom(
+                  padding: const EdgeInsets.symmetric(
+                    horizontal: 10,
+                    vertical: 4,
+                  ),
+                  backgroundColor: settings.isTimePlaying
+                      ? Colors.amber.shade800
+                      : Colors.blueGrey.shade700,
+                ),
+                icon: Icon(
+                  settings.isTimePlaying ? Icons.pause : Icons.play_arrow,
+                  size: 16,
+                ),
+                label: Text(
+                  settings.isTimePlaying ? 'Pause' : 'Play',
+                  style: const TextStyle(fontSize: 11),
+                ),
+                onPressed: () {
+                  setState(() {
+                    settings.isTimePlaying = !settings.isTimePlaying;
+                  });
+                  onChanged();
+                },
+              ),
+            ],
+          ),
+          const SizedBox(height: 4),
           KitSliderRow(
             label: 'Time',
             value: settings.timeOfDay,
@@ -169,13 +246,14 @@ class _KitScenarioPanel extends StatelessWidget {
             suffix: 'h',
             onChanged: (v) {
               settings.timeOfDay = v;
+              setState(() {});
               onChanged();
             },
           ),
           KitSliderRow(
             label: 'Cycle speed',
             value: settings.timeSpeed,
-            min: 0.0,
+            min: 0.1,
             max: 4.0,
             onChanged: (v) {
               settings.timeSpeed = v;
@@ -195,19 +273,19 @@ class _KitScenarioPanel extends StatelessWidget {
           ),
           const SizedBox(height: 6),
           const Text(
-            'Drives physical Rayleigh/Mie sky scattering and directional sun shadows smoothly with zero discontinuities.',
+            'Drives physical Rayleigh/Mie atmospheric scattering and cascaded directional sun shadows.',
             style: TextStyle(color: Colors.white70, fontSize: 11),
           ),
         ];
 
       case _KitScenario.waterBuoyancy:
         return [
-          const KitSectionHeader('Wave Simulation'),
+          const KitSectionHeader('Gerstner Wave Simulation'),
           KitSliderRow(
             label: 'Amplitude',
             value: settings.waveAmplitude,
             min: 0.1,
-            max: 1.2,
+            max: 2.0,
             onChanged: (v) {
               settings.waveAmplitude = v;
               onChanged();
@@ -217,7 +295,7 @@ class _KitScenarioPanel extends StatelessWidget {
             label: 'Speed',
             value: settings.waveSpeed,
             min: 0.2,
-            max: 3.0,
+            max: 4.0,
             onChanged: (v) {
               settings.waveSpeed = v;
               onChanged();
@@ -235,16 +313,45 @@ class _KitScenarioPanel extends StatelessWidget {
           ),
           const SizedBox(height: 6),
           const Text(
-            'Trochoidal Gerstner waves drive real-time surface height, normal orientation, and pitch/roll of floating buoys and crates.',
+            'Trochoidal Gerstner waves compute height and surface normal dynamically, pitching and rolling floating props.',
             style: TextStyle(color: Colors.white70, fontSize: 11),
           ),
         ];
 
       case _KitScenario.flocking:
         return [
-          const KitSectionHeader('Flock Parameters'),
+          const KitSectionHeader('Interactive Swarm'),
+          Center(
+            child: ElevatedButton.icon(
+              icon: const Icon(Icons.waves, size: 16),
+              label: const Text('Scatter Flock Impulse'),
+              onPressed: widget.onScatterFlock,
+            ),
+          ),
+          const SizedBox(height: 6),
           KitSliderRow(
-            label: 'Separation',
+            label: 'Boid count',
+            value: settings.boidCount,
+            min: 8.0,
+            max: 60.0,
+            fractionDigits: 0,
+            onChanged: (v) {
+              settings.boidCount = v;
+              onChanged();
+            },
+          ),
+          KitSliderRow(
+            label: 'Max speed',
+            value: settings.maxSteerSpeed,
+            min: 2.0,
+            max: 14.0,
+            onChanged: (v) {
+              settings.maxSteerSpeed = v;
+              onChanged();
+            },
+          ),
+          KitSliderRow(
+            label: 'Separation dist',
             value: settings.separationDistance,
             min: 0.5,
             max: 4.0,
@@ -254,19 +361,39 @@ class _KitScenarioPanel extends StatelessWidget {
             },
           ),
           KitSliderRow(
-            label: 'Max speed',
-            value: settings.maxSteerSpeed,
-            min: 2.0,
-            max: 12.0,
+            label: 'Separation wt',
+            value: settings.separationWeight,
+            min: 0.5,
+            max: 5.0,
             onChanged: (v) {
-              settings.maxSteerSpeed = v;
+              settings.separationWeight = v;
+              onChanged();
+            },
+          ),
+          KitSliderRow(
+            label: 'Cohesion wt',
+            value: settings.cohesionWeight,
+            min: 0.1,
+            max: 3.0,
+            onChanged: (v) {
+              settings.cohesionWeight = v;
+              onChanged();
+            },
+          ),
+          KitSliderRow(
+            label: 'Alignment wt',
+            value: settings.alignmentWeight,
+            min: 0.1,
+            max: 3.0,
+            onChanged: (v) {
+              settings.alignmentWeight = v;
               onChanged();
             },
           ),
           const SizedBox(height: 6),
           const Text(
-            'Autonomous boids steering with seek, separation, cohesion, and alignment behavior towards a moving target.',
-            style: TextStyle(color: Colors.white70, fontSize: 11),
+            'Click / drag on the 3D scene to guide the boid swarm to your cursor!',
+            style: TextStyle(color: Colors.amberAccent, fontSize: 11),
           ),
         ];
 
@@ -294,7 +421,7 @@ class _KitScenarioPanel extends StatelessWidget {
             child: ElevatedButton.icon(
               icon: const Icon(Icons.bolt, size: 16),
               label: const Text('Fire Projectile Burst'),
-              onPressed: onFireBurst,
+              onPressed: widget.onFireBurst,
             ),
           ),
           const SizedBox(height: 8),
@@ -315,7 +442,7 @@ class _KitScenarioPanel extends StatelessWidget {
                   ),
                 ),
                 Text(
-                  'Pool Buffer Size: ${settings.poolCapacity}',
+                  'Pool Idle Available: ${settings.poolCapacity}',
                   style: const TextStyle(color: Colors.white70, fontSize: 11),
                 ),
                 Text(
@@ -401,16 +528,24 @@ class _KitStageState extends State<_KitStage> {
   // Water scenario
   WaterSurfaceComponent? _water;
   final List<Node> _floatingProps = [];
+  double _cachedWaveAmplitude = -1.0;
+  double _cachedWaveSpeed = -1.0;
+  double _cachedWaveSteepness = -1.0;
 
   // Flocking scenario
   final List<Node> _boidNodes = [];
   final List<vm.Vector3> _boidVelocities = [];
+  final Node _targetMarkerNode = Node();
+  vm.Vector3? _userAttractorPos;
+  bool _isUserAttracting = false;
+  double _cachedBoidCount = -1.0;
 
   // Pooling scenario
   NodePool? _projectilePool;
   final List<_ActiveProjectile> _activeProjectiles = [];
   final List<Node> _scatteredProps = [];
   double _fireTimer = 0.0;
+  double _cachedMinDistance = -1.0;
 
   double _elapsed = 0.0;
   final Set<LogicalKeyboardKey> _pressedKeys = {};
@@ -431,41 +566,7 @@ class _KitStageState extends State<_KitStage> {
     super.didUpdateWidget(oldWidget);
     if (oldWidget.scenario != widget.scenario) {
       _buildScene();
-    } else if (widget.scenario == _KitScenario.spawnerPooling &&
-        oldWidget.settings.minDistance != widget.settings.minDistance) {
-      _rebuildPoissonFoliage();
-    } else if (widget.scenario == _KitScenario.waterBuoyancy &&
-        (oldWidget.settings.waveAmplitude != widget.settings.waveAmplitude ||
-            oldWidget.settings.waveSpeed != widget.settings.waveSpeed ||
-            oldWidget.settings.waveSteepness !=
-                widget.settings.waveSteepness)) {
-      _rebuildWater();
     }
-  }
-
-  void _rebuildWater() {
-    if (_water != null) {
-      scene.root.removeComponent(_water!);
-    }
-    _water = WaterSurfaceComponent(
-      waves: [
-        GerstnerWave(
-          direction: vm.Vector2(1.0, 0.2).normalized(),
-          amplitude: widget.settings.waveAmplitude,
-          wavelength: 12.0,
-          speed: widget.settings.waveSpeed,
-          steepness: widget.settings.waveSteepness,
-        ),
-        GerstnerWave(
-          direction: vm.Vector2(0.5, 0.8).normalized(),
-          amplitude: widget.settings.waveAmplitude * 0.5,
-          wavelength: 6.0,
-          speed: widget.settings.waveSpeed * 1.3,
-          steepness: widget.settings.waveSteepness,
-        ),
-      ],
-    );
-    scene.root.addComponent(_water!);
   }
 
   @override
@@ -486,18 +587,24 @@ class _KitStageState extends State<_KitStage> {
 
   void _buildScene() {
     scene.removeAll();
+    scene.skybox = null;
+    scene.skyEnvironment = null;
+    scene.sunLight = null;
+
     scene.add(_cameraNode);
     scene.add(_debugMeshNode);
 
-    // Reset default directional light
-    final keyLight = DirectionalLight()
-      ..color = vm.Vector3(1.0, 0.98, 0.92)
-      ..intensity = 50000.0;
-    final keyLightNode = Node()
-      ..addComponent(DirectionalLightComponent(keyLight));
-    keyLightNode.localTransform = vm.Matrix4.identity()
-      ..setTranslation(vm.Vector3(12, 24, 12));
-    scene.add(keyLightNode);
+    if (widget.scenario != _KitScenario.dayNight) {
+      // Standard studio key light for general scenarios
+      final keyLight = DirectionalLight()
+        ..color = vm.Vector3(1.0, 0.98, 0.92)
+        ..intensity = 40000.0;
+      final keyLightNode = Node()
+        ..addComponent(DirectionalLightComponent(keyLight));
+      keyLightNode.localTransform = vm.Matrix4.identity()
+        ..setTranslation(vm.Vector3(12, 24, 12));
+      scene.add(keyLightNode);
+    }
 
     switch (widget.scenario) {
       case _KitScenario.characterCamera:
@@ -585,6 +692,11 @@ class _KitStageState extends State<_KitStage> {
       enableRotationLag: widget.settings.enableLag,
       rotationLagSpeed: widget.settings.lagSpeed,
       cameraNode: _cameraNode,
+      inheritYaw: false,
+      inheritPitch: false,
+      inheritRoll: false,
+      yaw: widget.settings.cameraOrbitYaw,
+      pitch: widget.settings.cameraOrbitPitch,
     );
     _characterNode!.addComponent(_springArm!);
 
@@ -600,7 +712,12 @@ class _KitStageState extends State<_KitStage> {
     );
     scene.skybox = Skybox(physicalSky);
     scene.skyEnvironment = SkyEnvironment(physicalSky);
-    scene.sunLight = SunLight(physicalSky, castsShadow: true);
+    scene.sunLight = SunLight(
+      physicalSky,
+      castsShadow: true,
+      shadowAmbientStrength: 0.65,
+      shadowMaxDistance: 80.0,
+    );
 
     // Ground plane
     final groundMesh = PlaneGeometry(width: 60, depth: 60);
@@ -633,7 +750,9 @@ class _KitStageState extends State<_KitStage> {
 
     _dayNight = DayNightCycleComponent(
       timeOfDay: widget.settings.timeOfDay,
-      timeSpeed: widget.settings.timeSpeed,
+      timeSpeed: widget.settings.isTimePlaying
+          ? widget.settings.timeSpeed
+          : 0.0,
       latitude: widget.settings.latitude,
       skySource: physicalSky,
     );
@@ -647,7 +766,35 @@ class _KitStageState extends State<_KitStage> {
     );
   }
 
+  void _rebuildWater() {
+    if (_water != null) {
+      scene.root.removeComponent(_water!);
+    }
+    _water = WaterSurfaceComponent(
+      waves: [
+        GerstnerWave(
+          direction: vm.Vector2(1.0, 0.2).normalized(),
+          amplitude: widget.settings.waveAmplitude,
+          wavelength: 10.0,
+          speed: widget.settings.waveSpeed,
+          steepness: widget.settings.waveSteepness,
+        ),
+        GerstnerWave(
+          direction: vm.Vector2(0.5, 0.8).normalized(),
+          amplitude: widget.settings.waveAmplitude * 0.6,
+          wavelength: 6.0,
+          speed: widget.settings.waveSpeed * 1.3,
+          steepness: widget.settings.waveSteepness,
+        ),
+      ],
+    );
+    scene.root.addComponent(_water!);
+  }
+
   void _buildWaterBuoyancy() {
+    _cachedWaveAmplitude = widget.settings.waveAmplitude;
+    _cachedWaveSpeed = widget.settings.waveSpeed;
+    _cachedWaveSteepness = widget.settings.waveSteepness;
     _rebuildWater();
     _floatingProps.clear();
 
@@ -691,81 +838,100 @@ class _KitStageState extends State<_KitStage> {
     );
   }
 
-  void _buildFlocking() {
+  void _rebuildFlock() {
+    for (final n in _boidNodes) {
+      n.parent?.remove(n);
+    }
     _boidNodes.clear();
     _boidVelocities.clear();
 
-    final boidGeo = CuboidGeometry(vm.Vector3(0.4, 0.3, 1.0));
+    final boidCount = widget.settings.boidCount.round();
+    final boidGeo = CuboidGeometry(vm.Vector3(0.4, 0.25, 1.1));
     final boidMat = PhysicallyBasedMaterial()
       ..baseColorFactor = vm.Vector4(0.2, 0.95, 0.6, 1.0)
-      ..roughnessFactor = 0.3;
+      ..roughnessFactor = 0.25;
 
     final rng = math.Random(42);
-    for (var i = 0; i < 28; i++) {
+    for (var i = 0; i < boidCount; i++) {
       final pos = vm.Vector3(
-        (rng.nextDouble() - 0.5) * 18.0,
-        rng.nextDouble() * 8.0 + 2.0,
-        (rng.nextDouble() - 0.5) * 18.0,
+        (rng.nextDouble() - 0.5) * 16.0,
+        rng.nextDouble() * 6.0 + 2.0,
+        (rng.nextDouble() - 0.5) * 16.0,
       );
       final node = Node(mesh: Mesh(boidGeo, boidMat))..position = pos;
       _boidNodes.add(node);
       _boidVelocities.add(
         vm.Vector3(
-          (rng.nextDouble() - 0.5) * 3.0,
-          (rng.nextDouble() - 0.5) * 3.0,
-          (rng.nextDouble() - 0.5) * 3.0,
+          (rng.nextDouble() - 0.5) * 4.0,
+          (rng.nextDouble() - 0.5) * 2.0,
+          (rng.nextDouble() - 0.5) * 4.0,
         ),
       );
       scene.add(node);
     }
+  }
+
+  void scatterFlock() {
+    final rng = math.Random();
+    for (var i = 0; i < _boidVelocities.length; i++) {
+      final angle = rng.nextDouble() * 2 * math.pi;
+      final speed =
+          widget.settings.maxSteerSpeed * (1.8 + rng.nextDouble() * 0.8);
+      _boidVelocities[i] = vm.Vector3(
+        math.cos(angle) * speed,
+        (rng.nextDouble() - 0.3) * speed * 0.5,
+        math.sin(angle) * speed,
+      );
+    }
+  }
+
+  void _buildFlocking() {
+    _cachedBoidCount = widget.settings.boidCount;
+    _rebuildFlock();
+
+    // Target attractor marker beacon
+    final targetGeo = CuboidGeometry(vm.Vector3(0.8, 0.8, 0.8));
+    final targetMat = PhysicallyBasedMaterial()
+      ..baseColorFactor = vm.Vector4(1.0, 0.25, 0.25, 1.0)
+      ..emissiveFactor = vm.Vector4(1.0, 0.3, 0.3, 1.0);
+    _targetMarkerNode.mesh = Mesh(targetGeo, targetMat);
+    scene.add(_targetMarkerNode);
 
     _cameraNode.localTransform = BoundsFraming.computeFramingTransform(
-      vm.Aabb3.minMax(vm.Vector3(-14, 0, -14), vm.Vector3(14, 12, 14)),
+      vm.Aabb3.minMax(vm.Vector3(-14, 0, -14), vm.Vector3(14, 10, 14)),
       PerspectiveCamera(fovRadiansY: 1.0),
-      viewDirection: vm.Vector3(0.6, 0.55, 0.65),
-      paddingFactor: 1.4,
+      viewDirection: vm.Vector3(0.65, 0.5, 0.65),
+      paddingFactor: 1.35,
     );
   }
 
   void _buildSpawnerPooling() {
     _activeProjectiles.clear();
-    _scatteredProps.clear();
+    _fireTimer = 0.0;
+    _cachedMinDistance = widget.settings.minDistance;
 
-    // Ground plane
-    final groundMesh = PlaneGeometry(width: 36, depth: 36);
-    scene.add(
-      Node(
-        mesh: Mesh(
-          groundMesh,
-          PhysicallyBasedMaterial()
-            ..baseColorFactor = vm.Vector4(0.22, 0.28, 0.24, 1.0)
-            ..roughnessFactor = 0.85,
-        ),
-      ),
-    );
-
-    // Center cannon / turret
-    final turretBaseGeo = CuboidGeometry(vm.Vector3(1.8, 0.8, 1.8));
+    // Cannon turret base
+    final turretBaseGeo = CuboidGeometry(vm.Vector3(2.0, 0.8, 2.0));
     final turretBaseMat = PhysicallyBasedMaterial()
-      ..baseColorFactor = vm.Vector4(0.2, 0.2, 0.25, 1.0)
+      ..baseColorFactor = vm.Vector4(0.3, 0.32, 0.35, 1.0)
       ..metallicFactor = 0.8
       ..roughnessFactor = 0.2;
-    final turretBarrelGeo = CuboidGeometry(vm.Vector3(0.6, 0.6, 2.2));
-    final turretBarrelMat = PhysicallyBasedMaterial()
-      ..baseColorFactor = vm.Vector4(0.8, 0.2, 0.2, 1.0)
-      ..metallicFactor = 0.5
-      ..roughnessFactor = 0.3;
-
-    final turretNode = Node(mesh: Mesh(turretBaseGeo, turretBaseMat))
-      ..position = vm.Vector3(0, 0.4, 0);
-    turretNode.add(
-      Node(mesh: Mesh(turretBarrelGeo, turretBarrelMat))
-        ..position = vm.Vector3(0, 0.5, 0.6)
-        ..rotation = vm.Quaternion.axisAngle(vm.Vector3(1, 0, 0), -0.25),
+    scene.add(
+      Node(mesh: Mesh(turretBaseGeo, turretBaseMat))
+        ..position = vm.Vector3(0, 0.4, 0),
     );
-    scene.add(turretNode);
 
-    // 1. Poisson-disc distribution of trees around the perimeter
+    final barrelGeo = CuboidGeometry(vm.Vector3(0.5, 0.5, 1.8));
+    final barrelMat = PhysicallyBasedMaterial()
+      ..baseColorFactor = vm.Vector4(0.15, 0.15, 0.18, 1.0)
+      ..metallicFactor = 0.9
+      ..roughnessFactor = 0.15;
+    scene.add(
+      Node(mesh: Mesh(barrelGeo, barrelMat))
+        ..position = vm.Vector3(0, 0.9, 0.8),
+    );
+
+    // 1. Procedural Poisson-disc forest distribution
     _rebuildPoissonFoliage();
 
     // 2. Zero-GC NodePool for recycled projectiles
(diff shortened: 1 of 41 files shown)


packages/flutter_scene/lib/src/kit/environment/day_night_cycle_component.dart
```diff
@@ -0,0 +1,132 @@
...
+
+    // Aim the directional light
+    final lightTarget = vm.Vector3.zero();
+    final lightEye = sunDir * 100.0;
+    final viewMat = vm.makeViewMatrix(
+      lightEye,
+      lightTarget,
+      vm.Vector3(0, 1, 0),
+    );
+    targetNode.localTransform = viewMat..invert();
```
bdero: The sun is aimed backwards. `vm.makeViewMatrix` builds the OpenGL view basis (forward is -Z), so inverting it gives a node whose +Z points *toward* the eye. `DirectionalLightComponent` documents the opposite, "Light travels along the owning node's local +Z axis", and `NodeCamera` reads +Z as forward too.

Measured at `timeOfDay: 12, latitude: 34`, the sun node's +Z basis comes out `(0, 0.829, 0.559)`, the direction toward the sun, so the light shines up out of the ground.

Building the basis the way `_matrix4LookAt` in `camera.dart` does (forward = target - position) fixes it, and reusing that helper keeps one convention in the package.

Two more problems in the same expression. `up` of `(0, 1, 0)` is parallel to the view direction when the sun is overhead, and at `latitude: 0, timeOfDay: 12` the basis collapses to zeros (`_matrix4LookAt` asserts on exactly this case). And `viewMat..invert()` is a world transform assigned to `localTransform`, so it is wrong for any sun node that is not a direct child of an untransformed root.

Nothing here is covered by a test; `day_night_cycle_test.dart` only exercises `sunDirection` and `evaluateLighting`.

packages/flutter_scene/lib/src/kit/environment/day_night_cycle_component.dart
```diff
@@ -0,0 +1,132 @@
...
+    if (timeSpeed != 0.0 && deltaSeconds > 0.0) {
+      timeOfDay = (timeOfDay + timeSpeed * deltaSeconds) % 24.0;
+    }
+
+    final targetNode =
+        sunLightNode ??
+        (node.children.firstWhere(
+          (c) => c.getComponent<DirectionalLightComponent>() != null,
+          orElse: () => node,
+        ));
```
bdero: The `orElse: () => node` fallback silently hijacks the owning node's transform. An owner at `(5, 5, 5)` with no light child moves to `(0, 82.9, 55.9)` after one update.

Two ways to land here without noticing. `firstWhere` scans direct children only, so a sun light nested one level deeper falls through. And `cloneFor` drops `sunLightNode`, so every clone of a correctly configured component takes this path and starts flinging its own node around.

Doing nothing when there is no target, and asserting in debug, would be safer than picking the owner.

packages/flutter_scene/lib/src/kit/environment/day_night_cycle_component.dart
```diff
@@ -0,0 +1,132 @@
...
+    final x = math.sin(solarAngle) * math.cos(latRad);
+    final z = math.cos(solarAngle) * math.sin(latRad);
+    final y = elevation;
+
+    final dir = vm.Vector3(x, y, z);
+    return dir.length2 > 0 ? dir.normalized() : vm.Vector3(0, 1, 0);
+  }
+
+  /// Evaluates lighting parameters based on current sun elevation.
+  AtmosphericLighting evaluateLighting() {
```
bdero, line 87: `evaluateLighting` is never called by `update`, so nothing applies `sunColor`, `sunIntensity`, `ambientColor`, or `shadowDarkness` to anything. The class doc promises "dynamic lighting transitions" and the skill says the component evaluates "sun colors, intensities, and ambient lighting transitions", which reads as if the component drives the light.

Either wire it into the `DirectionalLight` in `update` (and `Scene.environmentIntensity` for ambient), or say plainly that this is a pure evaluator the caller applies.

packages/flutter_scene/lib/src/kit/camera/bounds_framing.dart
```diff
@@ -0,0 +1,40 @@
...
+    if (proj is PerspectiveProjection) {
+      final halfFov = proj.fovRadiansY * 0.5;
+      distance = (radius / math.sin(halfFov)) * paddingFactor;
+    } else {
+      distance = radius * 2.0 * paddingFactor;
+    }
+
+    final eye = center + dir * distance;
+    final viewMat = vm.makeViewMatrix(eye, center, up);
+    return viewMat..invert();
```
bdero: Same view-basis inversion as the day/night component. A node given this transform faces away from the bounds, since `NodeCamera.forward` is +Z and `makeViewMatrix` puts -Z there. The framing test only checks the eye distance, so it passes anyway.

Bigger question though: this duplicates `PerspectiveCamera.framing`, which already computes `radius / sin(fovY / 2) * margin` from the same bounding sphere and additionally fits near and far around the model, and `OrbitCameraController.frame`. `OrbitCameraController.frame` even carries a TODO asking for the field-of-view-aware version this class implements. Resolving that TODO, or just deferring to `PerspectiveCamera.framing`, looks better than a third framing path with a different orientation convention.

packages/flutter_scene/lib/src/kit/camera/bounds_framing.dart
```diff
@@ -0,0 +1,40 @@
...
+import 'package:flutter_scene/src/camera.dart';
+import 'package:vector_math/vector_math.dart' as vm;
+
+/// Utility for framing node hierarchies or bounding boxes in a camera's view.
+/// {@category Gameplay kit}
+class BoundsFraming {
+  /// Computes a camera world transform that frames [bounds] with [paddingFactor].
+  ///
+  /// [viewDirection] is the normalized direction vector from the bounds center
+  /// towards the camera eye (defaults to a diagonal elevated view: `(1, 1, 1)`).
```
bdero: Doc says the default is `(1, 1, 1)`, the code uses `(1.0, 0.8, 1.0)`. Worth noting `PerspectiveCamera.framing` defaults to `(0, 0, -1)`, "matching the direction glTF models face after import", so the kit disagrees with the engine on which way a framed model is viewed from.

packages/flutter_scene/lib/src/kit/character/third_person_controller.dart
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    // 5. Apply displacement to node
+    var newPos = currentPos + velocity * deltaSeconds;
+    if (isGrounded && newPos.y < groundY) {
+      newPos.y = groundY;
+    }
+
+    final rot = vm.Quaternion.axisAngle(vm.Vector3(0, 1, 0), _currentYaw);
+    node.localTransform = vm.Matrix4.compose(newPos, rot, node.scale);
```
bdero: World position in, local transform out. `currentPos` comes from `globalTransform` and `newPos` is written straight to `localTransform`, so the character teleports the moment it is not a direct child of an untransformed root.

Reproduced with a player under a rig at z=20 under a `scale(1, 1, -1)` root: local goes `(0,0,0)` to `(0,0,-20)` on the first update, which snaps the world position from `(0,0,-20)` to the origin. That parent shape is not exotic, the runtime importer always synthesizes exactly that Z-flip root.

Either work in local space throughout, or convert the world result back through `inverse(parent.globalTransform)` before assigning. `spring_arm_component.dart:161` writes `socketTransform` (world) into `cameraNode.localTransform` the same way.

packages/flutter_scene/lib/src/kit/character/third_person_controller.dart
```diff
@@ -0,0 +1,211 @@
...
+
+    // 1. Ground detection probe via scene raycast
+    var detectedGround = false;
+    var groundY = currentPos.y;
+    var norm = vm.Vector3(0, 1, 0);
+
+    // Cast downward ray starting slightly above the character base
+    final rayStart = currentPos + vm.Vector3(0, 0.4, 0);
+    final ray = vm.Ray.originDirection(rayStart, vm.Vector3(0, -1, 0));
+    final hit = raycastNode(_rootNode, ray, maxDistance: 0.6 + maxStepHeight);
```
bdero: The ground probe casts against the whole scene from the tree root and never excludes the character, so the first hit is the character's own mesh (the ray starts 0.4 units above its origin and points down through it). `raycastNode` has `where` and `layerMask` for exactly this; passing `where: (n) => n != node` or a ground layer mask is the minimum.

The same applies to the boom probe at `spring_arm_component.dart:144`, where the arm's nearest occluder is always the character it is mounted on.

Two more things worth putting in the doc, since this runs every frame. `raycastNode` has no BVH yet (its own TODO) and tests every candidate triangle after a bounds early-out, and it tests skinned meshes at rest pose, so an animated ground mesh probes against the wrong shape.

packages/flutter_scene/lib/src/kit/character/third_person_controller.dart
```diff
@@ -0,0 +1,211 @@
...
+    if (hit != null && hit.distance <= 0.45 + maxStepHeight) {
+      detectedGround = true;
+      groundY = rayStart.y - hit.distance;
+      norm = hit.worldNormal;
+    } else if (currentPos.y <= 0.05) {
```
bdero: An implicit, un-disableable infinite ground plane at y=0. Nothing in the class doc or the skill mentions it, and a character can never fall below y=0 no matter what the scene looks like.

It is also load-bearing for the tests: both cases in `third_person_controller_test.dart` are grounded only because of this branch, so the raycast path has no coverage at all.

An optional `groundPlaneHeight` (nullable, defaulting to null) would make it a choice instead of a surprise.

packages/flutter_scene/lib/src/kit/character/third_person_controller.dart
```diff
@@ -0,0 +1,211 @@
...
+  /// Upward velocity impulse applied on jump.
+  double jumpVelocity;
+
+  /// Downward gravitational acceleration.
+  double gravity;
+
+  /// Maximum slope angle (in degrees) that can be walked up without sliding.
+  double maxSlopeAngleDegrees;
+
+  /// Maximum step obstacle height (in units) the character can smoothly step up.
```
bdero: `maxStepHeight` never steps anything up, it only lengthens the ground probe. There is no step detection and no vertical resolve, so a character walking into a 0.2 unit ledge stops dead against it.

The class doc and the skill both advertise step climbing. Either implement it (a forward probe plus an upward resolve) or drop the claim and leave a `TODO(kit)` saying what is missing.

packages/flutter_scene/lib/src/kit/character/third_person_controller.dart
```diff
@@ -0,0 +1,211 @@
...
+  Node get _rootNode {
+    Node curr = node;
+    while (curr.parent != null) {
+      curr = curr.parent!;
+    }
+    return curr;
+  }
+
+  @override
+  void update(double deltaSeconds) {
```
bdero, line 131: Two timing notes.

`Component.fixedUpdate` exists and its doc names "character motion drivers" as the reason it exists. Integrating gravity and jumps on the frame clock makes jump height vary with frame rate.

The smoothing is also frame-rate dependent in a way the rest of the PR is not: `(accelRate * deltaSeconds).clamp(0, 1)` here and `(turnSpeed * deltaSeconds).clamp(0, 1)` at line 167, against `1 - exp(-k * dt)` in `spring_arm_component.dart`. Worth using the exponential form in both.

bdero (commented):
A big, self-consistent kit, and the shape is right: components that carry their own state, pure evaluators where the caller wants control, and a skill so agents find them. Tests, formatting, and analysis are all green, and the barrel is properly show-listed.

Two themes hold it back from being mergeable as is.

**The engine's conventions are not followed in the places that matter most.** Three components build transforms with `vm.makeViewMatrix(...)..invert()`, which is the OpenGL basis where forward is -Z. flutter_scene uses +Z as forward for both `NodeCamera` and `DirectionalLightComponent`. The consequences are visible, not subtle: the day/night sun shines up out of the ground at noon, and a camera node given a `BoundsFraming` transform points away from what it is framing. Separately, three components read a world transform and write it into `localTransform`, so a character or camera parented under anything (including the Z-flip root the runtime importer always synthesizes) teleports on the first update. I measured each of these rather than reasoning about them; details are inline.

**Several pieces re-implement engine API that already exists, usually less well.** `KinematicCharacterController` already does move-and-slide with slope limits, ground snap, and real autostep, against collider volumes rather than a single downward ray. `FollowCameraController` is already a smoothed third-person orbit rig, and its TODO literally asks for the occlusion retraction `SpringArmComponent` adds. `PerspectiveCamera.framing` already computes the same framing distance with the same formula and also fits near and far, and `OrbitCameraController.frame` carries the TODO for the field-of-view-aware version. It would be worth deciding, per component, whether the kit version replaces the engine one, wraps it, or should be dropped, and saying so in the docs. Right now a consumer (and more to the point, an agent reading the skill) has no way to tell which to reach for.

Smaller but worth catching before publish: both raycasting components hit the node they are mounted on, since neither passes `where` or `layerMask`; `ThirdPersonControllerComponent` has an undocumented, un-disableable ground plane at y=0 that both of its tests silently depend on; `maxStepHeight` and `SoundManager`'s "voice pooling" describe behavior that is not there; `CameraShakeOffset.zero` is a shared mutable that any caller can corrupt process-wide; and the skill's day/night example does not compile.

On coverage, the suite is a good start but it exercises the arithmetic and not the integration. Every test drives `update` directly on an unmounted, unparented node, which is precisely the configuration where the parent-space and self-hit bugs cannot appear. A few cases with a real parent transform, a mounted node, and a mesh in the raycast path would have caught most of the above.

Character motion also probably wants `fixedUpdate` rather than `update`, since its doc names character motion drivers as the reason that hook exists.

packages/flutter_scene/lib/src/kit/camera/spring_arm_component.dart
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    // 4. Update child camera node if bound
+    if (cameraNode != null) {
+      cameraNode!.localTransform = socketTransform;
+    }
+  }
+
+  @override
+  Component? cloneFor(Node cloneOwner) {
```
bdero, line 262: `cameraNode` is dropped from the clone, so a cloned rig keeps smoothing a boom that drives nothing. `Component.cloneFor` documents this case, "Components whose state references other nodes cannot self-clone through this hook and need caller-driven rebinding", and its default is to return null rather than a half-wired copy. Returning null here (and in `DayNightCycleComponent`, where the null target is actively harmful) matches that.

packages/flutter_scene/lib/src/kit/camera/spring_arm_component.dart
```diff
@@ -0,0 +1,179 @@
...
+    // Arm direction is back (-Z in model space) rotated by the boom rotation.
+    final backDir = rot.rotate(vm.Vector3(0, 0, 1));
+    final socketPos = pivot + backDir * currentLength;
+    final finalPos = socketPos + rot.rotate(socketOffset);
+
+    return vm.Matrix4.compose(finalPos, rot, vm.Vector3.all(1.0));
+  }
+
+  vm.Vector3 get _currentRawPivotWorld {
+    if (!isAttached) return targetOffset;
```
bdero: Returns the live `targetOffset` field rather than a copy, and `onMount` assigns it straight into `_smoothedPivotWorld`. Nothing mutates it today, so this is latent, but `Node.position` and friends hand out clones for exactly this reason. `targetOffset.clone()` costs nothing here.

packages/flutter_scene/lib/src/kit/camera/camera_shake.dart
```diff
@@ -0,0 +1,90 @@
...
+
+/// Evaluated translational and rotational camera shake offsets.
+/// {@category Gameplay kit}
+class CameraShakeOffset {
+  final vm.Vector3 translation;
+  final vm.Vector3 rotationEuler;
+
+  CameraShakeOffset({required this.translation, required this.rotationEuler});
+
+  static final CameraShakeOffset zero = CameraShakeOffset(
```
bdero: `zero` is a shared mutable singleton that `update` returns on every idle frame. One caller doing `offset.translation.setValues(...)` corrupts it for the whole process, which I confirmed: after a single mutation `CameraShakeOffset.zero.translation` reads `(9, 9, 9)` everywhere.

A getter returning a fresh instance avoids it.

packages/flutter_scene/lib/src/kit/environment/water_surface_component.dart
```diff
@@ -0,0 +1,112 @@
...
+}
+
+/// Generates and animates realistic trochoidal ocean water surfaces.
+/// {@category Gameplay kit}
+class WaterSurfaceComponent extends Component {
+  /// Active Gerstner wave spectrum.
+  final List<GerstnerWave> waves;
+
+  /// Underlying updatable water plane geometry if driven on CPU.
+  MeshGeometry? meshGeometry;
```
bdero: `meshGeometry` is public and documented but never read anywhere in the package. Either drive the geometry from `evaluateAt` in `update` (the doc says "if driven on CPU", which is the useful half of this component) or drop the field until it does something.

Minor, in the same class: `cloneFor` passes the same `waves` list instance to the clone, and `q` divides by `k * amplitude * waves.length`, so an amplitude of zero produces infinities.

packages/flutter_scene/lib/src/kit/audio/sound_manager.dart
```diff
@@ -0,0 +1,36 @@
+import 'dart:math' as math;
+import 'package:flutter_scene/src/audio/audio_bus.dart';
+import 'package:flutter_scene/src/audio/audio_clip.dart';
+import 'package:flutter_scene/src/audio/audio_engine.dart';
+import 'package:flutter_scene/src/audio/audio_voice.dart';
+import 'package:vector_math/vector_math.dart' as vm;
+
+/// High-level audio utility with automatic voice pooling, pitch jitter, and spatial one-shots.
```
bdero: The doc claims "automatic voice pooling", and there is none, this is one static wrapper adding pitch jitter around `AudioEngine.playOneShot`. Worth trimming the doc to what it does.

It also narrows the wrapped API by dropping `attenuation`, and clamps `volume` to 1.0 where the engine does not, so anyone wanting either has to bypass the helper entirely.

packages/flutter_scene/lib/src/kit/pooling/volume_spawner.dart
```diff
@@ -0,0 +1,83 @@
...
+class PoissonDiscSampler {
+  /// Generates a set of 2D points within a rectangular area with minimum separation [minDistance].
+  static List<vm.Vector2> sampleRect(
+    double width,
+    double height,
+    double minDistance, {
+    int maxCandidates = 30,
+    int? seed,
+  }) {
+    final rng = math.Random(seed ?? 42);
```
bdero: `seed ?? 42` means every unseeded call returns the same layout. For a helper sold as "natural obstacle and foliage scattering" that is surprising enough to either document or change to a random default.

`minDistance <= 0` also throws from inside the grid allocation (`(width / 0).ceil()` on an infinity) rather than at the boundary. An `ArgumentError` at the top would say what went wrong.

Naming nit, the file is `volume_spawner.dart` but the only API samples a 2D rect.

packages/flutter_scene/lib/src/kit/debug/debug_draw.dart
```diff
@@ -0,0 +1,139 @@
...
+  /// Builds a [MeshGeometry] containing all accumulated line segments and clears the buffer.
+  static MeshGeometry? flushMesh() {
+    if (_positions.isEmpty) return null;
+
+    final posList = Float32List.fromList(_positions);
+    final colList = Float32List.fromList(_colors);
+    _positions.clear();
+    _colors.clear();
+
+    return MeshGeometry.fromArrays(
```
bdero, line 144: `flushMesh` allocates a fresh `MeshGeometry`, and so a fresh GPU buffer, on every frame it is called, which is the intended usage for immediate-mode debug draw. `MeshGeometry.fromArrays` takes `bufferArena`, and `GeometryStorage.updatable` with `rebuild` reuses the allocation when the vertex count is stable. Threading one of those through (an optional arena parameter, or a retained geometry the class updates) would keep this off the per-frame allocation path.

Separately, `_positions` and `_colors` grow without bound if a caller draws but never flushes, and being static means two views cannot keep separate debug buffers.

packages/flutter_scene/lib/src/kit/pooling/node_pool.dart
```diff
@@ -0,0 +1,75 @@
+import 'package:flutter_scene/src/node.dart';
+import 'package:vector_math/vector_math.dart' as vm;
+
+/// High-performance object recycler for transient nodes (projectiles, damage numbers, effects).
+/// {@category Gameplay kit}
+class NodePool {
+  final Node Function() factory;
```
bdero, line 8: `factory` and `maxSize` are public with no doc comments. Same in `GerstnerWave`, `CameraShakeOffset`, and `AtmosphericLighting`, which declare their fields bare. CLAUDE.md asks for a doc and a `{@category}` on every public symbol.

`maxSize` also needs its actual meaning spelled out. It caps retained idle nodes, not live ones, so `spawn` happily creates past it and the excess is dropped on despawn rather than reused.

packages/flutter_scene/lib/src/kit/camera/virtual_joystick.dart
```diff
@@ -0,0 +1,126 @@
...
+  /// Radius of the outer boundary circle.
+  final double radius;
+
+  /// Radius of the inner draggable knob.
+  final double knobRadius;
+
+  /// Deadzone radius (0.0 to 1.0) below which input is reported as zero.
+  final double deadzone;
+
+  /// Callback receiving the normalized direction vector (-1.0 to 1.0 on X and Y).
```
bdero: The emitted Y follows screen coordinates, so pushing up gives negative Y. That is fine, but it needs saying here, because the natural pairing is `controller.setMoveInput(dir)` and `ThirdPersonControllerComponent` maps `input.y` onto +Z, which makes pushing up walk backwards. The skill shows the two together without mentioning it.

Also `deadzone: 1.0` divides by zero at line 66. An assert on the constructor range would catch it.

packages/flutter_scene/lib/src/kit/character/steering_behaviors.dart
```diff
@@ -0,0 +1,136 @@
...
+
+    final rampedSpeed = maxSpeed * (distance / slowingRadius);
+    final targetSpeed = math.min(rampedSpeed, maxSpeed);
+    final desiredVel = (toTarget / distance) * targetSpeed;
+    final force = desiredVel - currentVel;
+    return truncate(force, maxForce);
+  }
+
+  /// Computes a separation force avoiding crowding nearby [neighbors].
+  static vm.Vector3 separation(
```
bdero, line 58: `separation` returns a velocity scaled to `maxSpeed`, while `seek`, `flee`, `arrive`, `alignment`, and `cohesion` all return `desired - currentVel`, an actual steering force. Summing them the way the skill implies gives separation an outsized, differently-dimensioned contribution. Taking `currentVel` here for consistency, or renaming, would avoid it.

packages/flutter_scene/skills/flutter_scene-kit/SKILL.md
```diff
@@ -0,0 +1,135 @@
...
+```
+
+## Dynamic environments and atmosphere
+
+### DayNightCycleComponent
+
+`DayNightCycleComponent` moves the sun along a realistic solar arc given latitude and time of day, evaluating sun colors, intensities, and ambient lighting transitions.
+
+```dart
+final sunNode = Node()..addComponent(DirectionalLightComponent());
```
bdero: This does not compile. `DirectionalLightComponent` takes a required positional `DirectionalLight`, so it needs `DirectionalLightComponent(DirectionalLight())` (verified against the analyzer). `sunNode` is also never added to the scene, so the light would not register even once the constructor is fixed.

CLAUDE.md is specific about this one, a skill that teaches a wrong name is worse than no skill, so the examples want a compile check before this lands.

packages/flutter_scene/skills/flutter_scene-kit/SKILL.md
```diff
@@ -0,0 +1,135 @@
...
+  targetLength: 5.0,
+  targetOffset: vm.Vector3(0, 1.6, 0), // Eye height
+  socketOffset: vm.Vector3(0.5, 0, 0), // Over-the-shoulder
+  enablePositionLag: true,
+  positionLagSpeed: 8.0,
+  cameraNode: cameraNode,
+);
+
+characterNode.addComponent(cameraArm);
+scene.root.add(characterNode);
```
bdero, line 45: This wiring only works by accident. `SpringArmComponent` writes a world-space `socketTransform` into `cameraNode.localTransform`, which happens to be correct here only because `cameraNode` is a direct child of an untransformed root. Parent the camera anywhere else and it drifts.

Worth saying which way to feed the camera once the parent-space issue at `spring_arm_component.dart:161` is settled.

packages/flutter_scene/skills/flutter_scene-kit/SKILL.md
```diff
@@ -0,0 +1,135 @@
...
+// Inside game loop
+final offset = shake.update(deltaSeconds);
+cameraNode.localTransform = baseTransform * offset.toMatrix4();
+```
+
+## Character movement and steering
+
+### ThirdPersonControllerComponent
+
+`ThirdPersonControllerComponent` handles planar movement, sprint multipliers, turn smoothing, ground snapping with raycasts, slope sliding, step climbing, coyote time, and buffered jumps.
```
bdero: Repeats the step climbing claim, which is not implemented, and the ground snapping description leaves out the implicit y=0 plane and the self-hit raycast. A skill is the surface most likely to be believed literally, so it should be the most conservative description in the PR, not the most generous.

Nothing in the skill points at `KinematicCharacterController`, `FollowCameraController`, or `PerspectiveCamera.framing` either, so an agent reading it will reach for the kit even where the engine already has the better answer.

packages/flutter_scene/lib/src/kit/environment/day_night_cycle_component.dart
bdero: Done in e26531e8. Oriented the sun along +Z toward the scene target, guarded against zenith degeneracy by falling back to Vector3(0, 0, 1), and converted the world matrix relative to the parent transform.

packages/flutter_scene/lib/src/kit/environment/day_night_cycle_component.dart
bdero: Done in e26531e8. Removed the owner fallback so the component does nothing when no directional light target is found.

packages/flutter_scene/lib/src/kit/environment/day_night_cycle_component.dart
bdero: Done in e26531e8. Wired evaluateLighting into update with an applyLightingToTarget flag that sets the light color and intensity.

packages/flutter_scene/lib/src/kit/camera/bounds_framing.dart
bdero: Done in e26531e8. Reoriented the framing matrix to +Z forward and clarified the docs to recommend PerspectiveCamera.framing for standalone cameras.

packages/flutter_scene/lib/src/kit/camera/bounds_framing.dart
bdero: Done in e26531e8. Updated the default view direction to (0, 0, -1) to match glTF orientation conventions and PerspectiveCamera.framing.

packages/flutter_scene/lib/src/kit/character/third_person_controller.dart
bdero: Done in e26531e8. Inverted the parent global transform when applying the updated world transform.

packages/flutter_scene/lib/src/kit/character/third_person_controller.dart
bdero: Done in e26531e8. Added where filtering to exclude the character hierarchy, added groundLayerMask, and documented raycast characteristics.

packages/flutter_scene/lib/src/kit/character/third_person_controller.dart
bdero: Done in e26531e8. Replaced the implicit floor with an optional groundPlaneHeight parameter defaulting to null.

packages/flutter_scene/lib/src/kit/character/third_person_controller.dart
bdero: Done in e26531e8. Dropped the step climbing claim and added a TODO(kit) for forward-probe step climbing.

packages/flutter_scene/lib/src/kit/character/third_person_controller.dart
bdero: Done in e26531e8. Moved kinematic simulation to fixedUpdate and transitioned movement and rotation to exponential smoothing.

packages/flutter_scene/lib/src/kit/camera/spring_arm_component.dart
bdero: Done in e26531e8. Returned null from cloneFor when cameraNode is bound, matching the Component.cloneFor convention.

packages/flutter_scene/lib/src/kit/camera/spring_arm_component.dart
bdero: Done in e26531e8. Cloned targetOffset and socketOffset when computing pivot positions.

packages/flutter_scene/lib/src/kit/camera/camera_shake.dart
bdero: Done in e26531e8. Changed zero into a getter that returns a fresh instance.

packages/flutter_scene/lib/src/kit/environment/water_surface_component.dart
bdero: Done in e26531e8. Dropped unused meshGeometry, added guards against zero amplitude and wavelength, and cloned the wave list in cloneFor.

packages/flutter_scene/lib/src/kit/audio/sound_manager.dart
bdero: Done in e26531e8. Trimmed the doc to reflect one-shot playback, forwarded attenuation and bus, and removed artificial volume clamping.

packages/flutter_scene/lib/src/kit/pooling/volume_spawner.dart
bdero: Done in e26531e8. Defaulted to random unseeded generation unless a seed is provided, and added ArgumentError checks for non-positive inputs.

packages/flutter_scene/lib/src/kit/debug/debug_draw.dart
bdero: Done in e26531e8. Added bufferArena and storage parameters to flushMesh, and capped accumulated vertex limits.

packages/flutter_scene/lib/src/kit/pooling/node_pool.dart
bdero: Done in e26531e8. Added doc comments and categories across all public fields and clarified that maxSize caps idle retained instances.

packages/flutter_scene/lib/src/kit/camera/virtual_joystick.dart
bdero: Done in e26531e8. Documented screen-space Y coordinates and added a constructor assert on deadzone.

packages/flutter_scene/lib/src/kit/character/steering_behaviors.dart
bdero: Done in e26531e8. Accepted currentVel in separation to compute desired - currentVel steering forces matching other behaviors.

packages/flutter_scene/skills/flutter_scene-kit/SKILL.md
bdero: Done in e26531e8. Fixed DirectionalLightComponent instantiation and added sunNode to the scene.

packages/flutter_scene/skills/flutter_scene-kit/SKILL.md
bdero: Done in e26531e8. Updated the example and documented mounting cameraNode directly to the scene root.

packages/flutter_scene/skills/flutter_scene-kit/SKILL.md
bdero: Done in e26531e8. Updated the skill to describe grounded motion accurately and added guidance pointing to KinematicCharacterController and FollowCameraController.

packages/flutter_scene/lib/src/kit/character/third_person_controller.dart
```diff
@@ -0,0 +1,362 @@
...
+    return false;
+  }
+
+  @override
+  void update(double deltaSeconds) {
+    fixedUpdate(deltaSeconds);
+  }
+
+  @override
+  void fixedUpdate(double fixedDt) {
```
bdero, line 136: `update` delegating to `fixedUpdate` double-integrates whenever a `PhysicsWorld` is in the scene. `Scene._tick` runs `_stepPhysics` (which walks `sceneFixedPass`, so `fixedTick` then `fixedUpdate`, once per substep) and then `root.scenePrePass` (so `tick` then `update`, which lands here again).

Measured over one 1/60 frame at the default gravity, `velocity.y` ends at -0.6 instead of -0.3. With several substeps it scales with the substep count, so jump height and walk speed change the moment a physics world is added to the scene.

This needs a guard rather than an unconditional delegation. Tracking whether a fixed step already ran this frame, or skipping the `update` path when `findAncestorWorld(node)` is non-null, would both do it. Worth a test with a `PhysicsWorld` in the tree, since nothing currently covers the two-driver case.

packages/flutter_scene/lib/src/kit/camera/spring_arm_component.dart
```diff
@@ -0,0 +1,302 @@
...
+  bool enablePositionLag;
+
+  /// Speed at which the lagged position catches up to the target (higher = faster).
+  double positionLagSpeed;
+
+  /// Whether the boom rotation smoothly lags behind the target rotation.
+  bool enableRotationLag;
+
+  /// Speed at which the lagged rotation catches up (higher = faster).
+  double rotationLagSpeed;
```
bdero: `enableRotationLag` and `rotationLagSpeed` no longer do anything. The spherical rewrite dropped `_smoothedRotation`, but the fields, their docs, the constructor parameters, and the `cloneFor` copy all survived.

Confirmed by setting `rotationLagSpeed: 0.001` (effectively frozen) and snapping the owner 1.5 rad in one 16ms tick: the eye still travels the whole way, from `(0, 1.5, -4)` to `(-3.99, 1.5, -0.28)`.

Either restore the lag by smoothing `effectiveYaw`/`effectivePitch`, or delete both fields. Shipping a knob that silently does nothing is the worse of the two.

bdero (commented):
Much better. The whole first round is genuinely addressed rather than patched around: the view-basis inversions are gone and replaced with explicit +Z-forward bases (with the degenerate-up case handled), the three components that mixed world and local now convert through the parent inverse, both raycasting components pass `where` and `layerMask` and exclude their own subtree, the y=0 floor is an opt-in `groundPlaneHeight`, smoothing is exponential throughout, `cloneFor` returns null where node references cannot survive a clone, and `evaluateLighting` is wired into the light. Test count went from 20 to 36, and the new cases target the exact failures from last time (parent transform plus Z-flip, self-exclusion, a real floor mesh, +Z forward, equator noon, unshared `zero`). The kit docs and the skill now point at `KinematicCharacterController`, `FollowCameraController`, and `PerspectiveCamera.framing`, which was the ask on overlap. Analysis, formatting, the full 1956-test package suite, and the example app are all green locally.

Three things to sort before this lands.

**`update` delegating to `fixedUpdate` double-integrates.** `Scene._tick` runs the fixed pass and then the frame pass, so with a `PhysicsWorld` anywhere in the scene `ThirdPersonControllerComponent` steps gravity and movement once per substep plus once more per frame. Measured at -0.6 instead of -0.3 for one frame of gravity. It needs a guard, not an unconditional delegation, and a test with a world in the tree.

**The spring arm rewrite left three dead knobs.** `enableRotationLag` / `rotationLagSpeed`, `roll` / `inheritRoll`, and `socketOffset.z` are all still public, still documented, still copied by `cloneFor`, and all verified to have no effect. Implement or delete; a knob that silently does nothing is worse than an absent one. The pivot also quietly switched to world axes while its doc still says local.

**Unrelated public API rode along.** `ExternalTexture.colorFilter` and `Geometry.setCpuPositionsForTesting` are additions to published classes, and `SunLight.resolve` gained a behavior change that zeroes intensity and shadows below the horizon with a hard step, no opt-out, and no test. None are in the CHANGELOG, which the release checklist expects for everything the `dart_apitool` diff reports. The `SunLight` one is the one I would push back on hardest as a change on its own merits: it is a visible pop at sunrise, and `DayNightCycleComponent.shadowOnly` already covers the kit's need without touching shipped behavior.

Smaller, inline: a mangled doc comment on `jumpBufferWindow`; ten full-scene raycasts per fixed step with an O(depth) exclusion walk per candidate; obstacle probe heights that ignore `obstacleHeight`; `sunIntensity` still documented in lux when it is now a 0-to-3 multiplier; `ambientColor` and `shadowDarkness` still never applied; `DebugDraw` silently dropping geometry past its cap.

On the example, two things stand out. `main.dart` still hardcodes the startup screen to Gameplay Kit, which looks like a leftover from `52fc4df4`. And the showcase reimplements the joystick privately rather than using the `VirtualJoystick` this PR ships, while `FloatingMotionComponent`, `SoundManager`, and `SurfaceFootstepAudio` go unexercised entirely. It also calls `flushMesh()` bare, skipping the `bufferArena` and `storage` parameters that were added in this very PR for that path.

Last, the skill still has a snippet that does not compile (`Mesh` takes positional arguments, not `material:`) and a wrong import path for `KinematicCharacterController`. That is twice now, so it is worth making mechanical: a never-called function in a test file holding the snippets verbatim would let the analyzer enforce them from here on.

packages/flutter_scene/lib/src/kit/camera/spring_arm_component.dart
```diff
@@ -0,0 +1,302 @@
...
+  bool inheritRoll;
+
+  /// Boom yaw angle in radians (used when [inheritYaw] is false).
+  double yaw;
+
+  /// Boom pitch angle in radians (used when [inheritPitch] is false).
+  double pitch;
+
+  /// Boom roll angle in radians (used when [inheritRoll] is false).
+  double roll;
```
bdero: `roll` and `inheritRoll` are inert too. `_updateSocketTransform` always passes `up: (0, 1, 0)` to `lookAtTransform` and no `effectiveRoll` exists, so the up column comes out identical for `roll: 1.0` and `roll: 0.0` (verified).

Same call as the rotation lag fields: implement it by rotating `up` about the look direction, or drop it.

packages/flutter_scene/lib/src/kit/camera/spring_arm_component.dart
```diff
@@ -0,0 +1,302 @@
...
+  double currentLength;
+
+  /// Radius of the sphere probe used to test for obstacles (0.0 for a thin ray).
+  double probeRadius;
+
+  /// Local offset from the owning node's origin to the boom's pivot.
+  vm.Vector3 targetOffset;
+
+  /// Local offset applied at the camera socket end (e.g. over-the-shoulder offset).
+  vm.Vector3 socketOffset;
```
bdero, line 30: `socketOffset.z` is silently ignored. `_updateSocketTransform` applies only `right * x` and world-up `* y`, so an offset of `(0, 0, 5)` and `(0, 0, 0)` produce the same eye position (verified).

Given the doc calls it a "local offset applied at the camera socket end", either apply the Z component along the boom direction (which is what an over-the-shoulder rig usually wants for a forward nudge) or make it a `Vector2` so the dead axis is not offered.

packages/flutter_scene/lib/src/kit/camera/spring_arm_component.dart
```diff
@@ -0,0 +1,302 @@
...
+
+  vm.Vector3 get _currentRawPivotWorld {
+    if (!isAttached) return targetOffset.clone();
+    final parentPos = vm.Vector3.zero();
+    node.globalTransform.decompose(
+      parentPos,
+      vm.Quaternion.identity(),
+      vm.Vector3.zero(),
+    );
+    return parentPos + targetOffset;
```
bdero, line 131: The pivot is now built as `parentPos + targetOffset`, which applies the offset along world axes rather than the owner's. The doc above still says "Local offset from the owning node's origin", and the previous revision did rotate it through `globalTransform`.

With the owner rolled 180 degrees about Z, the pivot still sits at +2 on world Y rather than -2 (verified).

Keeping the offset world-aligned is defensible for a camera boom (eye height should not tip when the character does), but then the doc should say so. If it is unintentional, `node.globalTransform.transform3(targetOffset.clone())` restores the old behavior.

Minor, same method: `decompose` into three throwaway temporaries just to read the translation. `globalTransform.getTranslation()` is the direct form.

packages/flutter_scene/lib/src/kit/character/third_person_controller.dart
```diff
@@ -0,0 +1,362 @@
...
+  /// Downward gravitational acceleration.
+  double gravity;
+
+  /// Maximum slope angle (in degrees) that can be walked up without sliding.
+  double maxSlopeAngleDegrees;
+
+  /// Time window (in seconds) after leaving a ledge during which a jump is still allowed.
+  double coyoteTimeWindow;
+
+  /// Time window (in seconds)  /// Time buffer window for jump inputs before landing (in seconds).
```
bdero, line 38: Mangled doc comment, two `///` runs spliced into one line.

```dart
/// Time buffer window for jump inputs before landing (in seconds).
```

packages/flutter_scene/lib/src/kit/character/third_person_controller.dart
```diff
@@ -0,0 +1,362 @@
...
+import 'package:flutter_scene/src/components/component.dart';
+import 'package:flutter_scene/src/node.dart';
+import 'package:flutter_scene/src/raycast.dart';
+import 'package:vector_math/vector_math.dart' as vm;
+
+/// High-level 3rd-person character movement component with ground raycast snapping,
+/// slope sliding, coyote time, and turn smoothing.
+///
+/// For physical character navigation against 3D colliders and capsules with autostep,
+/// see `KinematicCharacterController` in `package:flutter_scene_rapier/physics.dart`.
```
bdero: `package:flutter_scene_rapier/physics.dart` does not exist. `KinematicCharacterController` is exported from `package:flutter_scene/physics.dart`; `flutter_scene_rapier`'s only entry point is `flutter_scene_rapier.dart`.

Same wrong path in `SKILL.md` line 12, which matters more, since that is the copy an agent will paste.

packages/flutter_scene/lib/src/kit/character/third_person_controller.dart
```diff
@@ -0,0 +1,362 @@
...
+          vm.Ray.originDirection(probeStart, moveDir),
+          maxDistance: maxProbeDist,
+          layerMask: groundLayerMask,
+          where: (n) => !_isExcluded(n),
+        );
+        if (hit != null &&
+            (closestHit == null || hit.distance < closestHit.distance)) {
+          closestHit = hit;
+        }
+      }
```
bdero, line 308: Nine obstacle probes plus the ground probe means ten full-scene raycasts per fixed step, and `raycastNode` has no BVH yet, so each walks the whole tree and tests candidate meshes triangle by triangle. Every candidate node also pays an O(depth) `_isExcluded` ancestor walk, ten times over. With physics substeps this multiplies again.

Worth at least documenting the cost on `obstacleRadius`, and probably worth cheaper structure: hoist the exclusion to a precomputed set, or use `raycastNodeAll` once per level and filter, or drop to three probes until a hit is found and only then refine.

Two smaller things in the same block. The probe heights 0.3, 0.85, and 1.4 are absolute, and only the top one is clamped by `obstacleHeight`, so a short character gets probes above its own head while `obstacleHeight` goes unused. And the deflection resolves against a single closest normal with no second pass, so an inside corner can still push through; a `TODO(kit)` naming that would be honest.

packages/flutter_scene/lib/src/kit/environment/day_night_cycle_component.dart
```diff
@@ -0,0 +1,192 @@
...
+import 'package:vector_math/vector_math.dart' as vm;
+
+/// Evaluated atmospheric lighting colors for a specific sun elevation.
+/// {@category Gameplay kit}
+class AtmosphericLighting {
+  /// Directional sun light color in linear RGB.
+  final vm.Vector3 sunColor;
+
+  /// Directional sun light illuminance in lux.
+  final double sunIntensity;
```
bdero, line 16: This says "illuminance in lux", but `evaluateLighting` now returns `3.0 * smoothT`, and it is assigned to `DirectionalLight.intensity`, which the engine documents as "Scalar multiplier applied to color" with a default of 3.0. The value is right for the engine; the doc is left over from the old 100000-lux version.

While here: `ambientColor` and `shadowDarkness` are still evaluated and still never applied by `update`, so half of `AtmosphericLighting` remains decorative. `Scene.environmentIntensity` is the obvious home for the first.

packages/flutter_scene/lib/src/kit/debug/debug_draw.dart
```diff
@@ -0,0 +1,147 @@
...
+import 'package:vector_math/vector_math.dart' as vm;
+
+/// Immediate-mode wireframe debug visualization utility for lines, rays, boxes, spheres, and axes.
+/// {@category Rendering}
+class DebugDraw {
+  static final List<double> _positions = [];
+  static final List<double> _colors = [];
+
+  /// Maximum vertex limit to avoid unbounded accumulation when un-flushed.
+  static int maxVertexLimit = 65536;
```
bdero, line 15: The cap works, but `line` early-returns silently once it is hit, so debug geometry just stops appearing with nothing to explain why, which is a rough failure mode for a debugging tool. A one-shot `debugPrint` or an assert when the limit is first crossed would cost nothing and save someone an afternoon.

packages/flutter_scene/lib/src/kit/camera/bounds_framing.dart
```diff
@@ -0,0 +1,55 @@
...
+/// {@category Gameplay kit}
+class BoundsFraming {
+  /// Computes a camera node world transform that frames [bounds] with [paddingFactor].
+  ///
+  /// The camera looks toward the bounds center from [viewDirection] (offset from center
+  /// toward the eye, defaulting to `(0, 0, -1)` to match glTF front-facing orientation).
+  /// The resulting matrix is oriented with +Z facing the target, matching `NodeCamera`.
+  static vm.Matrix4 computeFramingTransform(
+    vm.Aabb3 bounds,
+    Camera camera, {
```
bdero: `camera` is only ever read for `camera.projection`. Taking a `CameraProjection` instead would say that in the signature and let callers frame against a projection they have not built a camera around yet.

packages/flutter_scene/lib/src/geometry/geometry.dart
```diff
@@ -519,6 +519,17 @@ abstract class Geometry {
     indexCount: _indexCount,
   );
 
+  /// Injects CPU-side vertex positions and optional bounding box for testing
+  /// without requiring a GPU upload.
+  @visibleForTesting
+  void setCpuPositionsForTesting(Float32List positions, {vm.Aabb3? bounds}) {
```
bdero: This widens the published `Geometry` surface (it is exported from `scene.dart`) so a kit test can build raycastable geometry without a GPU. `@visibleForTesting` is a lint, not a boundary, so this is public API and belongs in the CHANGELOG and in the release `dart_apitool` diff.

It also leaves the object in a state the class otherwise never produces: `_cpuPositions` and `_vertexCount` are set while indices, `retainCpuData`, and the GPU buffer are not, so `extractMeshData` and anything index-driven see an inconsistent geometry. If it stays, the doc should say exactly which invariants it skips and that it is only safe for CPU-side queries such as raycasts.

The alternative worth weighing is keeping the fixture entirely in `test/`, since in-repo tests may already import `lib/src` directly.

packages/flutter_scene/lib/src/sun_light.dart
```diff
@@ -141,9 +141,15 @@ class SunLight {
       ..setFrom(source.sunDirection)
       ..negate();
     light.color.setFrom(color ?? source.sunLightColor);
-    light.intensity = (intensity ?? source.sunLightIntensity) * intensityScale;
+    final sunElevation = source.sunDirection.y;
```
bdero: Three concerns with this one, and none of them are about the kit.

It is a behavior change to shipped API in a PR about a gameplay kit, with no CHANGELOG entry and no test (`sun_light_test.dart` still only asserts `castsShadow` is true).

The cutoff is a hard step at `sunElevation > 0.0`, so intensity and shadows both snap off in a single frame as the sun crosses the horizon. A physical sky's `sunLightIntensity` near the horizon is small but not zero, so this reads as a pop at sunrise and sunset rather than a fade. Ramping over a small elevation band, the way the kit's own `evaluateLighting` does with `nightT`, avoids it.

And it silently overrides the explicit `intensity` field with no opt-out, so a caller deliberately keeping a moonlight-style key light below the horizon now cannot.

If the kit needs this, `shadowOnly` on `DayNightCycleComponent` already covers the kit's case without touching `SunLight`.

packages/flutter_scene/lib/src/texture/external_texture.dart
```diff
@@ -145,6 +146,10 @@ class ExternalTexture extends ChangeNotifier implements TextureSource {
   /// Sampling options used when a material samples this source.
   ExternalTextureSampling sampling;
 
+  /// Optional color filter applied to the captured texture (for example color
+  /// inversion or channel swizzling).
+  ui.ColorFilter? colorFilter;
```
bdero, line 151: `ExternalTexture.colorFilter` is a new public field on a class shipping in 0.24.0, and it is unrelated to the gameplay kit. It has a test, which is good, but no CHANGELOG entry, and the release checklist expects every API addition the `dart_apitool` diff reports to be in there.

Worth splitting into its own PR, or at minimum adding the CHANGELOG line, so the 0.24.0 notes describe what actually shipped.

examples/flutter_app/lib/main.dart
```diff
@@ -318,7 +320,7 @@ class _MyAppState extends State<MyApp> {
       'Split Screen': (context) => const ExampleSplitScreen(),
       'Stress Tests': (context) => const ExampleStressTests(),
     };
-    selectedExample = examples.keys.first;
+    selectedExample = 'Gameplay Kit';
```
bdero: Debugging leftover from `52fc4df4`. This pins the example app's startup screen to the new example instead of `examples.keys.first`, which changes what every other example's user lands on.

examples/flutter_app/lib/example_kit.dart
```diff
@@ -0,0 +1,1624 @@
...
+class _VirtualJoystick extends StatefulWidget {
+  const _VirtualJoystick({required this.onChanged});
+
+  final ValueChanged<vm.Vector2> onChanged;
+
+  @override
+  State<_VirtualJoystick> createState() => _VirtualJoystickState();
+}
+
+class _VirtualJoystickState extends State<_VirtualJoystick> {
```
bdero: The showcase for the kit reimplements the joystick privately instead of using `VirtualJoystick`, the widget this same PR ships. This copy has no deadzone, no configurable radius or decorations, and does its own Y inversion.

That reads as a verdict on the shipped widget. If it is missing something the example needed, fix the widget; if not, the example should be the reference usage. Either way the kit's own demo not using its own widget is the thing a reader will notice first.

`FloatingMotionComponent`, `SoundManager`, and `SurfaceFootstepAudio` are also unexercised by any scenario, which leaves three shipped components with no worked example.

examples/flutter_app/lib/example_kit.dart
```diff
@@ -0,0 +1,1624 @@
...
+    DebugDraw.sphere(
+      vm.Vector3(0, 2.0, 0),
+      radius,
+      color: vm.Vector4(1.0, 0.8, 0.2, 0.7),
+    );
+
+    // Flush to mesh
+    final mesh = DebugDraw.flushMesh();
+    if (mesh != null) {
+      _debugMeshNode.mesh = Mesh(mesh, UnlitMaterial());
```
bdero: `flushMesh` grew `bufferArena` and `storage` parameters in this PR precisely so per-frame debug draw stays off the allocation path, and the example calls it bare, then also allocates a fresh `UnlitMaterial` every frame.

Holding an arena and a single material on the state and passing them here would make the showcase demonstrate the intended usage instead of the one the parameters exist to avoid.

packages/flutter_scene/skills/flutter_scene-kit/SKILL.md
```diff
@@ -0,0 +1,144 @@
...
+```dart
+DebugDraw.line(startPos, endPos, color: vm.Vector4(1, 0, 0, 1));
+DebugDraw.box(aabb, color: vm.Vector4(0, 1, 0, 1));
+DebugDraw.sphere(center, 1.0, color: vm.Vector4(0, 0, 1, 1));
+DebugDraw.axes(node.globalTransform, size: 2.0);
+
+// Render debug lines
+final debugMesh = DebugDraw.flushMesh();
+if (debugMesh != null) {
+  debugNode.mesh = Mesh(debugMesh, material: UnlitMaterial());
```
bdero: Still does not compile. `Mesh` takes its geometry and material positionally, `Mesh(Geometry geometry, Material material)`, so there is no `material:` named parameter. The example app gets this right at `example_kit.dart:1551`.

The `DirectionalLightComponent` case from the last round is fixed, but this one was never checked, which suggests the snippets want a mechanical guard rather than another read-through. A test file under `test/` that contains the skill's snippets verbatim inside a function that is never called would make the analyzer enforce them, and would keep the skill honest as the API moves.

packages/flutter_scene/skills/flutter_scene-kit/SKILL.md
```diff
@@ -0,0 +1,144 @@
...
+version: 2
+description: Build interactive 3D gameplay, character controllers, camera rigs, dynamic day/night cycles, water surfaces, audio, pooling, and debug overlays in flutter_scene. Use when creating game mechanics, camera controls, NPC behaviors, atmospheric environments, or diagnostic HUDs.
+---
+
+# Gameplay, camera, and atmosphere kit in flutter_scene
+
+Flutter Scene provides high-level gameplay components and ergonomic building blocks in `package:flutter_scene/kit.dart` so games and interactive experiences do not need to re-implement standard mechanics from scratch.
+
+When choosing components, consider existing engine alternatives:
+- For physics-driven character navigation with collider capsules, wall sliding, and autostep, use `KinematicCharacterController` from `package:flutter_scene_rapier/physics.dart`.
```
bdero: Wrong import path. `KinematicCharacterController` comes from `package:flutter_scene/physics.dart`; `flutter_scene_rapier` exposes only `flutter_scene_rapier.dart` and has no `physics.dart`. Same text in the class doc at `third_person_controller.dart:11`.

This is the single line in the skill most likely to be copied verbatim, since it is the one telling an agent to go use something else.

packages/flutter_scene/skills/flutter_scene-kit/SKILL.md
```diff
@@ -0,0 +1,144 @@
...
+import 'package:flutter_scene/scene.dart';
+import 'package:flutter_scene/kit.dart';
+import 'package:vector_math/vector_math.dart' as vm;
+```
+
+## Camera rigs and smoothing
+
+### SpringArmComponent
+
+`SpringArmComponent` attaches to a target character node and mounts a camera node at the arm's socket. It casts rays against the scene hierarchy to prevent geometry clipping, smoothly pulling the camera inward when colliding with walls.
```
bdero, line 28: Worth mentioning here that `socketOffset` only acts on X and Y, and that `targetOffset` is applied along world axes rather than the character's, since both are surprising and both are things an over-the-shoulder rig runs into immediately.

If the rotation-lag and roll fields stay in the API, this section should also say they are not implemented yet. Better to remove them.

packages/flutter_scene/lib/src/kit/debug/debug_draw.dart
bdero: Done in 8d392283. Added a one-shot assert warning when the vertex limit is first exceeded.

packages/flutter_scene/lib/src/kit/environment/day_night_cycle_component.dart
bdero: Done in 8d392283. Wired shadowDarkness to shadowAmbientStrength and added targetScene to synchronize environmentIntensity.

packages/flutter_scene/lib/src/kit/camera/bounds_framing.dart
bdero: Done in 8d392283. Changed parameter to CameraProjection.

packages/flutter_scene/lib/src/geometry/geometry.dart
bdero: Done in 8d392283. Removed setCpuPositionsForTesting from Geometry and moved CPU mesh definitions into test fixtures.

packages/flutter_scene/lib/src/sun_light.dart
bdero: Done in 8d392283. Reverted the SunLight.resolve changes.

packages/flutter_scene/lib/src/texture/external_texture.dart
bdero: Done in 8d392283. Added colorFilter to the ExternalTexture entry in CHANGELOG.md.

examples/flutter_app/lib/main.dart
bdero: Done in 8d392283. Reset selectedExample back to examples.keys.first.

examples/flutter_app/lib/example_kit.dart
bdero: Done in 8d392283. Replaced the private joystick with the shipped VirtualJoystick widget and added FloatingMotionComponent to the water buoys.

examples/flutter_app/lib/example_kit.dart
bdero: Done in 8d392283. Reused a persistent GeometryBufferArena and UnlitMaterial across debug draw frames.

packages/flutter_scene/skills/flutter_scene-kit/SKILL.md
bdero: Done in 8d392283. Fixed the snippet, bumped skill version to 3, and added skill_snippets_test.dart to enforce snippet compilation.

packages/flutter_scene/skills/flutter_scene-kit/SKILL.md
bdero: Done in 8d392283. Corrected import path to package:flutter_scene/physics.dart.

packages/flutter_scene/skills/flutter_scene-kit/SKILL.md
bdero: Done in 8d392283. Removed unused rotation lag and roll fields, and documented offset coordinate spaces in the doc comments and SKILL.md.

packages/flutter_scene/lib/src/kit/character/third_person_controller.dart
bdero: Done in 8d392283. Kept simulation purely in fixedUpdate and verified frame step integration.

packages/flutter_scene/lib/src/kit/camera/spring_arm_component.dart
bdero: Done in 8d392283. Removed enableRotationLag and rotationLagSpeed.

packages/flutter_scene/lib/src/kit/camera/spring_arm_component.dart
bdero: Done in 8d392283. Removed roll and inheritRoll.

packages/flutter_scene/lib/src/kit/camera/spring_arm_component.dart
bdero: Done in 8d392283. Documented socketOffset as a socket-local right/up plane offset.

packages/flutter_scene/lib/src/kit/camera/spring_arm_component.dart
bdero: Done in 8d392283. Documented targetOffset as a world-space offset.

packages/flutter_scene/lib/src/kit/character/third_person_controller.dart
bdero: Done in 8d392283. Cleaned up the doc comment formatting.

packages/flutter_scene/lib/src/kit/character/third_person_controller.dart
bdero: Done in 8d392283. Corrected import path to package:flutter_scene/physics.dart.

packages/flutter_scene/lib/src/kit/character/third_person_controller.dart
bdero: Done in 8d392283. Documented probe costs and added TODO(kit) for acceleration structure and multi-probe refinement.

PR #360, opened 2026-08-24
Add dynamic diffuse global illumination and temporal anti-aliasing

bdero:
Adds screen-space radiance injection into a scrolling world-space probe grid for real-time indirect diffuse bounce light. Introduces temporal anti-aliasing with camera subpixel jitter, history reprojection with velocity buffers, variance-guided color clamping, and specular roughness anti-aliasing.


bdero (commented):
Both features are good work and the sampler-budget discipline on the irradiance field is the best part of the PR (extending the coefficient texture downward instead of declaring a sampler, pinned at 14 in `physical_material_sources_test.dart`). All 1945 package tests pass and `dart analyze` is clean on both packages.

The problem is the default path. Two unrelated changes here make every lit fragment shader substantially bigger and more texture-heavy with both features off. I compiled `shaders/flutter_scene_standard.frag` with the SDK `impellerc` (`--opengl-es --gles-language-version=300`) at the merge base and at head, and isolated each change.

| | GLSL lines | SPIR-V image ops | `texture(shadow_map)` sites |
| --- | --- | --- | --- |
| master (9581f52) | 2691 | 61 | 10 |
| + shadow filter change only | 2871 | 125 | 74 |
| + irradiance receiver (head) | 3594 | 141 | 74 |

Texture instructions in the default lit shader go from 61 to 141, 2.3x, with nothing enabled. Neither cost is executed when the feature is off, but that is not the cost that hurts on Mali. Register allocation and program size are worst-case over the whole shader, so occupancy drops for every lit draw regardless of the branch. This is the exact regression shape the 0.24.0 no-shadow variant was added to avoid, and both of these should follow that precedent and sit behind defines with the Dart side picking the variant.

Beyond that the recurring theme is per-frame and per-draw work that is not gated. `MeshComponent.refreshRenderItems` copies a `Matrix4` and looks up a joints texture per item per frame whether or not TAA runs, `FragInfo` grows 80 bytes for every lit draw, and the velocity pass allocates two `Float32List`s and emplaces a frame-constant uniform twice per draw.

Smaller items are inline. The ones I would treat as blocking are the two shader defines, the unconditional `previousWorldTransform` copy, the Android smoke-frame override, and the CHANGELOG gaps (the specular-AA change alters every existing lit material's shading and is not mentioned anywhere).

packages/flutter_scene/shaders/material_shadow_sampling.glsl
```diff
@@ -143,15 +171,31 @@ float SampleCascade(int cascade, int count, mat4 cascade_matrix, float box,
...
   // TODO(flutter_scene): use file-scope const arrays once impellerc/SPIRV-Cross
   // emits valid ES 1.00 array constructors for them.
-  int sample_count = fixed_filter > 0.5 ? 17 : 16;
-  float lit = 0.0;
-  for (int i = 0; i < 17; i++) {
-    if (i >= sample_count) break;
-    vec2 tap = mix(PoissonShadowTap(i), FixedShadowTap(i), fixed_filter);
-    lit += ShadowTap(tap, ca, sa, radius, uv, cascade, inv_count,
-                     receiver_depth);
+  float shadow = 0.0;
```
bdero, line 175: `bilinearPcf` is selected by a runtime uniform, so its 16 texel reads compile into every lit shader whether or not anything picks it. The straight-line branch also makes the compiler unroll the cascade loop that used to stay rolled, which is where most of the damage comes from.

Measured on `flutter_scene_standard.frag` with the SDK impellerc, this change on its own takes `texture(shadow_map)` sites from 10 to 74 and total SPIR-V image ops from 61 to 125. Nothing selects the new filter by default, so every existing scene pays for a filter it never uses, in the currency Mali is most sensitive to.

Please put it behind a define and pick the variant on the Dart side the way the no-shadow variant already does. If the extra bundle entries are too expensive, replacing `rotatedPoisson` rather than adding a fourth runtime branch would also work.

packages/flutter_scene/shaders/material_lighting.glsl
```diff
@@ -523,6 +505,32 @@ vec4 EvaluateLighting(MaterialInputs material) {
...
+  // colored bleed reads correctly, and fades back to it across the outermost
+  // cell. The field's own empty content is the environment, so an unfilled
+  // field is exactly the image above and the transition is continuous.
+  //
+  // The lookup direction is the camera-invariant shading normal, never the
+  // screen-derived bent normal: the visibility select is a hard boundary, so
+  // a camera-dependent normal makes it flicker like z-fighting. The field's
+  // stored radiance already carries the environment intensity, so it is not
+  // scaled again here.
+  float gi_coverage = IrradianceFieldCoverage(v_position);
```
bdero, line 520: `IrradianceFieldCoverage` and `SampleIrradianceField` compile into every non-lightmap lit variant, gated only by the runtime `gi_coverage > 0.0`. That is 16 texture fetches (8 cage irradiance plus 8 cage depth) and about 720 lines of emitted GLSL that a scene with `globalIllumination.enabled == false` still carries. Isolated, the receiver takes the standard frag from 125 to 141 SPIR-V image ops and from 2871 to 3594 GLSL lines.

The sampler sharing is the right call and I would keep it exactly as is. The receiver itself should sit behind `#ifdef FLUTTER_SCENE_IRRADIANCE_FIELD`, with the variant picked the way `Material.usesRadianceCubeVariant` picks the radiance layout. That also makes `visibility == 0` genuinely free instead of a branch, and lets the transmission path stop carrying a second eight-tap cage.

packages/flutter_scene/lib/src/global_illumination.dart
```diff
@@ -0,0 +1,158 @@
...
+  /// averaging error so flat surfaces do not self-shadow. Expressed relative
+  /// to cell size on purpose, so it does not have to be retuned for scene
+  /// scale.
+  double shadowBias = 0.3;
+
+  /// Strength of the depth-moment visibility test, `0` disabling it (pure
+  /// trilinear, smoother but leaks through thin walls) and `1` applying it
+  /// fully. Also see [visibilityBias].
+  ///
+  /// Setting this to `0` compiles the receiver's depth fetches out, halving
```
bdero, line 71: This is not true as written. `visibility` lands in `gi_visibility.x` and `IrradianceCageTap` tests it with `if (frag_info.gi_visibility.x > 0.0)`, so nothing is compiled out and the depth fetches stay in the program. Either make it a real define (see the receiver comment) or reword to say the fetches are skipped at runtime.

packages/flutter_scene/shaders/material_engine_lighting.glsl
```diff
@@ -147,6 +147,24 @@ uniform FragInfo {
...
+  // volume can scroll without moving a probe that stayed inside it).
+  // gi_anchor.w is the self-shadow bias as a fraction of the cell edge.
+  vec4 gi_anchor;
+  // gi_counts.xyz is the probe count per axis; gi_counts.w the probe tiles
+  // per atlas row.
+  vec4 gi_counts;
+  // gi_atlas.x/y are the irradiance and depth regions' first atlas rows;
+  // gi_atlas.zw the reciprocal atlas dimensions.
+  vec4 gi_atlas;
+  // gi_visibility.x is the Chebyshev strength (0 compiles the depth fetches
```
bdero, line 164: Same as the doc on `GlobalIlluminationSettings.visibility`, `0` does not compile the depth fetches out, it branches around them at runtime.

packages/flutter_scene/lib/src/components/mesh_component.dart
```diff
@@ -196,11 +196,15 @@ class MeshComponent extends Component {
       item.frustumCulled = frustumCulled;
       item.layers = layers;
       item.lightChannelMask = lightChannelMask;
+      item.isMoving =
+          transformChanged || (skin != null && jointsTexture != null);
+      item.previousWorldTransform.setFrom(item.worldTransform);
```
bdero: This and the `previousJointsTexture` lookup below run for every render item every frame, whether or not TAA is on and whether or not `objectMotion` is enabled. `setFrom` is a 16-double copy, and it is per primitive, so a multi-primitive mesh pays it N times for the same node transform. `getPreviousJointsTexture()` is a ring index plus a map lookup on top.

This is the hottest CPU loop in the engine and the feature is off by default. Worth gating on a flag the scene sets when TAA with object motion is active, the way `shadowStatic` bookkeeping is gated.

packages/flutter_scene/lib/src/material/engine_lighting.dart
```diff
@@ -198,6 +203,41 @@ class EngineLightingUniforms {
     fragInfo[dielectricF0Index] = defaultDielectricF0;
     fragInfo[dielectricF0Index + 1] = defaultDielectricF0;
     fragInfo[dielectricF0Index + 2] = defaultDielectricF0;
+    // The irradiance field at [176..195]. A zero intensity in gi_grid.w
+    // disables the whole receiver, so a scene without the field pays nothing
+    // beyond these writes.
+    const gi = irradianceFieldIndex;
+    final field = lighting.irradianceField;
+    if (field == null) {
+      for (var i = gi; i < gi + 20; i++) {
```
bdero, line 212: `FragInfo` goes from 176 to 196 floats (704 to 784 bytes) for every lit draw, and this writes 20 zeros per draw when the field is off. `fillRange` is the cheap fix. If the receiver becomes a define, the default variant can skip the whole block and keep the smaller block size.

packages/flutter_scene/lib/src/render/velocity_pass.dart
```diff
@@ -0,0 +1,234 @@
...
+      final pipeline = resolvePipeline(
+        vertexShader,
+        _fragmentShader,
+        vertexLayout: vertexLayout,
+      );
+
+      renderPass.clearBindings();
+      renderPass.bindPipeline(pipeline);
+      renderPass.setPrimitiveType(item.geometry.primitiveType);
+      renderPass.bindUniform(
```
bdero, line 189: `frameInfoData` is frame constant but gets emplaced twice per item, so a 2000-draw scene allocates 4000 transient blocks holding the same 144 bytes. Emplace it once above `submitItem` and bind the same view to both stages.

packages/flutter_scene/lib/src/render/velocity_pass.dart
```diff
@@ -0,0 +1,234 @@
...
+        vertexShader.getUniformSlot('VelocityFrameInfo'),
+        context.transientsBuffer.emplace(ByteData.sublistView(frameInfoData)),
+      );
+      renderPass.bindUniform(
+        _fragmentShader.getUniformSlot('VelocityFrameInfo'),
+        context.transientsBuffer.emplace(ByteData.sublistView(frameInfoData)),
+      );
+
+      if (isSkinned) {
+        final modelInfo = Float32List(36);
```
bdero: `Float32List(36)` here and `Float32List(32)` in the else branch are allocated per draw. Hoist two reusable buffers onto the pass and `setRange` into them.

packages/flutter_scene/lib/src/render/velocity_pass.dart
```diff
@@ -0,0 +1,234 @@
...
+    frameInfoData[33] = _currentJitterNdc.y;
+    frameInfoData[34] = _previousJitterNdc.x;
+    frameInfoData[35] = _previousJitterNdc.y;
+
+    final frustum = Frustum.matrix(_currentViewProjection);
+
+    void submitItem(RenderItem item) {
+      if (!item.visible || !item.primitiveVisible) return;
+      if ((item.layers & _layerMask) == 0) return;
+      if (!item.isMoving) return;
```
bdero, line 166: `isMoving` is only ever set by `MeshComponent`. `InstancedMeshComponent.refreshRenderItem` never touches it, so instanced geometry is permanently skipped here and ghosts under TAA. The unskinned path also binds `item.worldTransform` as the single instance transform, which would be wrong for an instanced item if one did get through. A `TODO(taa-instanced)` at minimum.

packages/flutter_scene/lib/src/render/velocity_pass.dart
```diff
@@ -0,0 +1,234 @@
...
+  void execute(RenderGraphContext context) {
+    final width = _dimensions.width.toInt();
+    final height = _dimensions.height.toInt();
+    if (width <= 0 || height <= 0) return;
+
+    final velocityTexture = context.texturePool.acquire(
+      TransientTextureDescriptor(
+        width: width,
+        height: height,
+        format: gpu.PixelFormat.r16g16b16a16Float,
```
bdero, line 114: Only `rg` plus a `1.0` mask in `b` is used, but this is a full-resolution RGBA16F target. If `rg16Float` is renderable on the shim's backends, using it (and reading the clear's zero alpha as the mask instead of a dedicated channel) halves the bandwidth of a full-screen buffer, which is worth having on a tiler.

packages/flutter_scene/lib/src/render/taa_pass.dart
```diff
@@ -0,0 +1,310 @@
...
+    renderPass.bindTexture(
+      _fragmentShader.getUniformSlot('previous_depth'),
+      prevDepth,
+      sampler: _nearestClamp,
+    );
+
+    drawCompat(renderPass, 6);
+    rendererSubmissions.submit(commandBuffer);
+
+    // Blit/copy resolved destination into writeHistory and linear depth into
```
bdero: Three full-screen passes per frame where one would do. The resolve writes `destination`, a second pass copies it into `writeHistory`, a third copies linear depth into `previousLinearDepth`.

The history copy looks avoidable by resolving straight into `writeHistory` and publishing that as the scene color, which also makes the A/B ping-pong do real work instead of being redundant with the copy. The depth copy could be avoided by ping-ponging the prepass target, and if it has to stay it does not need RGBA16F when the shader only reads `.r`. As written TAA adds a full-res read plus write twice over on top of the resolve itself.

packages/flutter_scene/shaders/flutter_scene_taa.frag
```diff
@@ -0,0 +1,209 @@
...
+  vec3 tm_curr = ToneMap(current_center);
+  vec3 tm_hist = ToneMap(clamped_history);
+  vec3 resolved_tm = mix(tm_hist, tm_curr, weight);
+  float tm_luma = dot(resolved_tm, vec3(0.2126, 0.7152, 0.0722));
+  if (tm_luma > 0.999) {
+    resolved_tm *= 0.999 / tm_luma;
+  }
+  vec3 resolved = max(vec3(0.0), UnToneMap(resolved_tm));
+
+  // 8. Optional sharpening.
```
bdero, line 201: `up`, `down`, `left` and `right` are four of the nine texels the neighborhood loop above already fetched. Carry them out of that loop instead of re-sampling.

More broadly this shader is around 28 fetches per pixel (9 depth for dilation, 9 color for the neighborhood, 5 Catmull-Rom, 5 sharpen). A cheaper variant for mobile, cross-shaped neighborhood and bilinear history, would be worth having before TAA is recommended anywhere near a Mali part.

packages/flutter_scene/lib/src/scene.dart
```diff
@@ -2075,12 +2179,49 @@ base class Scene implements SceneGraph {
       }
     }
     // Ambient occlusion, screen-space reflections, normals, and materials
-    // that sample scene depth need the geometry prepass. Depth-only post
+    // that sample scene depth need the geometry prepass. Depth-only post    // Depth and normal pre-passes need a perspective camera. Orthographic
```
bdero: Two comments got concatenated onto one line here. Looks like a bad merge.

packages/flutter_scene/lib/src/scene.dart
```diff
@@ -2557,6 +2802,176 @@ base class Scene implements SceneGraph {
...
+          Vector3.min(low, bounds.min, low);
+          Vector3.max(high!, bounds.max, high);
+        }
+        if (low == null || high == null) {
+          return (camera.position.clone(), settings.extents.clone());
+        }
+        // Pad by a probe spacing so surfaces on the bounds are inside the
+        // cage rather than on its outermost plane.
+        final size = (high - low)
+          ..scale(1.0)
```
bdero: `..scale(1.0)` is a no-op.

packages/flutter_scene/lib/src/scene.dart
```diff
@@ -2557,6 +2802,176 @@ base class Scene implements SceneGraph {
...
+          chosen = volume;
+          chosenContains = contains;
+        }
+        if (chosen != null) {
+          if (!identical(chosen, _activeIrradianceVolume) ||
+              chosen.consumeInvalidation()) {
+            _activeIrradianceVolume = chosen;
+            _irradianceField.invalidate();
+          }
+          settings.resolution.setFrom(chosen.resolution);
```
bdero: `render` mutating the caller's `GlobalIlluminationSettings` is surprising, and a user reading `globalIllumination.resolution` back gets whichever volume happened to be active last frame. Keep the active volume's resolution in `IrradianceFieldState` and leave the settings object alone.

packages/flutter_scene/lib/src/scene.dart
```diff
@@ -1084,6 +1165,29 @@ base class Scene implements SceneGraph {
...
+  /// for a hard camera cut or a wholesale lighting change that should not
+  /// converge in over the hysteresis tail.
+  void invalidateGlobalIllumination() => _irradianceField.invalidate();
+
+  /// Temporal anti-aliasing settings. Active when [antiAliasingMode] is
+  /// [AntiAliasingMode.taa].
+  final TemporalAntiAliasingSettings temporalAntiAliasing =
+      TemporalAntiAliasingSettings();
+
+  TaaHistoryState? _taaState;
```
bdero, line 1190: `_taaState`, `_taaFrameIndex` and `_irradianceField` are per-Scene, but `renderViews` builds one graph per view and each view bumps the jitter index and overwrites `previousViewTransform` during graph construction. With more than one view the history is shared across differently sized targets and the reprojection is wrong for all but one of them. A `TODO(taa-multiview)` saying both are single-view for now would be enough for this PR.

packages/flutter_scene/lib/src/camera.dart
```diff
@@ -15,7 +15,7 @@ import 'package:vector_math/vector_math.dart';
 abstract class CameraProjection {
   /// Returns the projection matrix for a render target of the given
   /// [aspectRatio] (width / height).
-  Matrix4 getProjectionMatrix(double aspectRatio);
+  Matrix4 getProjectionMatrix(double aspectRatio, {Vector2? jitter});
```
bdero, line 18: Adding a named parameter to an abstract method is breaking for anyone implementing `CameraProjection` outside the package, and both it and `PerspectiveProjection` are in the `scene.dart` show list. Needs a CHANGELOG line so the `dart_apitool` diff at release is expected rather than a surprise.

packages/flutter_scene/lib/src/render/irradiance_field.dart
```diff
@@ -0,0 +1,322 @@
...
+  static const int stateOriginY = 2;
+
+  /// First atlas row of the irradiance region.
+  int get irradianceOriginY => stateOriginY + tileRows;
+
+  /// First atlas row of the depth-moment region.
+  int get depthOriginY => irradianceOriginY + tileRows * irradianceTile;
+
+  /// Atlas width in texels.
+  int get atlasWidth => tilesPerRow * depthTile;
```
bdero, line 141: `atlasWidth` is sized for the depth tile (16), but the irradiance region addresses columns at stride 8 (`ProbeTileOrigin` with `kProbeIrradianceTile`), so the right half of every irradiance row is allocated and never written. At the default 16x8x16 grid that is roughly 1 MB of dead texels per atlas, times three atlases plus the accumulators. Giving the irradiance region its own `tilesPerRow` would reclaim it.

packages/flutter_scene/lib/src/render/irradiance_pass.dart
```diff
@@ -0,0 +1,810 @@
...
+    renderPass.bindUniform(
+      fragmentShader.getUniformSlot('InjectTileInfo'),
+      context.transientsBuffer.emplace(ByteData.sublistView(tileInfo)),
+    );
+
+    final spacing = placement.spacing;
+    final anchor = placement.anchor;
+    final counts = layout.resolution;
+    for (var corner = 0; corner < 8; corner++) {
+      final info = Float32List(48);
```
bdero: 48 floats rebuilt eight times when only indices 44 to 46 change per corner. Hoist the buffer and rewrite those three in the loop.

Bigger win available here too. At the default eighth resolution on a 1080p target this is roughly 32k texels times 8 corners times 2 accumulators, so about 520k instanced quads per frame. One draw with `instanceCount * 8` and the corner derived from the instance index would drop seven state resubmissions per accumulator and is worth measuring.

examples/smoke_render/integration_test/smoke_test.dart
```diff
@@ -69,17 +74,24 @@ void main() {
...
       // TODO(smoke): restore multi-frame Android settling when the emulator
       // watchdog no longer terminates sustained software rendering.
-      final settleFrames =
+      final baseFrames =
           !kIsWeb && defaultTargetPlatform == TargetPlatform.android ? 1 : 20;
-      final settleStep = Duration(milliseconds: 1000 ~/ settleFrames);
+      final settleStep = Duration(milliseconds: 1000 ~/ baseFrames);
+      // A converging feature (the irradiance field, a temporal resolve) has
+      // nothing to show in one frame, so its scene asks for more.
+      final settleFrames = smoke.warmupFrames > baseFrames
```
bdero: This silently overrides the Android cap. The comment directly above says Android uses a single frame because the emulator watchdog terminates sustained software rendering, and `max(warmupFrames, baseFrames)` puts the two GI scenes back to 45 frames there. Take the max only when the platform is not Android, or give `SmokeScene` a separate Android count.

`boundary.markNeedsPaint()` inside the loop also changes how every pre-existing scene settles, so expect Argos churn beyond the three new scenes.

examples/smoke_render/lib/smoke_scenes.dart
```diff
@@ -34,11 +34,30 @@ final GlobalKey smokeSceneKey = GlobalKey();
...
 
   final String id;
   final ({Scene scene, PerspectiveCamera camera}) Function() setup;
   final Future<void> Function()? preload;
+
+  /// Frames to render before the capture, for a feature that converges over
+  /// time instead of resolving in one frame.
+  ///
+  /// The capture loop always pumps its own settling frames, so this only
+  /// raises the count. It is deterministic despite running on wall-clock
```
bdero, line 53: The determinism argument does not hold. `IrradianceFieldState.consumeDeltaSeconds` clamps to `[1/240, 1/10]`, so the retention exponent is only frame-rate independent when a frame takes at least 100 ms. Anything faster than that (which is most of the matrix on a real GPU) feeds wall-clock jitter straight into the blend and makes the golden non-deterministic. Worth a test-only fixed step, or driving the blend off a frame counter during warm-up.

Separately, `irradiance_field` and `ddgi_cornell` are both cornell-box GI rigs and largely measure the same thing. Two scenes times six backends times 45 warm-up frames is a lot of lane time and Argos quota for one feature area.

packages/flutter_scene/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 ## 0.24.0
 
+* World-space global illumination via `Scene.globalIllumination`, a grid of octahedral irradiance probes fed by scattering the visible surfaces' shaded radiance, so bounce light persists for geometry the camera is not looking at and colored bleed reads correctly. Off by default.
```
bdero, line 5: The GI entries are good but nothing else in the PR is here. Missing at least `AntiAliasingMode.taa` with `TemporalAntiAliasingSettings`, `DirectionalShadowFilter.bilinearPcf`, `Scene.bakeIrradianceField` with `IrradianceFieldBakeStepper`, and the `CameraProjection.getProjectionMatrix` signature change.

The one that matters most for users is the specular AA change. `defaultSpecularAaThreshold` goes from 0.2 to 1.0, and `SpecularAARoughness` switches from the roughness^4 form to the Filament squareRoughness form using `max` of the two derivative terms rather than their sum. Every existing lit material shades differently after this, so it needs a line. The specular and clearcoat direct terms also now multiply by `sun_visibility` instead of `shadow`, which is a fix but still a visible change.

packages/scene/lib/src/specs.dart
```diff
@@ -932,6 +971,24 @@ class EnvironmentEffectsSpec {
   double screenSpaceReflectionsResolutionScale;
   bool fogEnabled;
   String fogMode;
+
+  /// World-space global illumination. See `GlobalIlluminationSettings`.
+  bool globalIlluminationEnabled;
```
bdero, line 1002: `packages/scene` has no CHANGELOG entry for these fields. Per the release notes a `scene` change is what forces both flutter_scene and rapier to re-release, so it should not land unrecorded.

Also worth deciding deliberately, the field's settings round-trip through fscene documents but `TemporalAntiAliasingSettings` does not. That split will bite when the editor grows a TAA toggle.

packages/flutter_scene/shaders/irradiance_receiver.glsl
```diff
@@ -0,0 +1,143 @@
...
+const float kProbeVisibilityFloor = 0.05;
+
+// One cage probe's contribution, premultiplied by its weight in rgb with the
+// weight itself in alpha. Single return; a probe outside the volume falls out
+// as a zero rather than an early exit.
+vec4 IrradianceCageTap(vec3 offset, vec3 base_lattice, vec3 alpha,
+                       vec3 biased_position, vec3 direction) {
+  vec3 counts = frag_info.gi_counts.xyz;
+  vec3 lattice = base_lattice + offset;
+  vec3 relative = lattice - frag_info.gi_anchor.xyz;
```
bdero, line 32: `lattice` and `gi_anchor.xyz` are both absolute lattice indices, so in `followCamera` mode they grow without bound as the camera travels. At a world coordinate around 1e5 with a 1.25 unit spacing the fp32 ulp on `lattice` is already about 1/128 of a cell, which quantizes `alpha` and will band the trilinear blend far from the origin. Worth a `TODO(gi-far-origin)` noting that the anchor should be subtracted on the CPU and only the relative lattice uploaded.

examples/flutter_app/lib/example_stress_tests.dart
```diff
@@ -157,6 +159,144 @@ final _catalog = <_StressTest>[
         'https://raw.githubusercontent.com/KhronosGroup/glTF-Sample-Assets/'
         'main/Models/Sponza/glTF/Sponza.gltf',
     sizeBytes: 52686624,
+    settings: () => ExampleSettings()
```
bdero, line 162: These look like a slider dump rather than authored values (`111.14397321428582`, `0.00204729352678569`). Rounding them to something a human can read and re-tune would make this a lot easier to maintain, and it would not change the look.

packages/flutter_scene/lib/src/render/velocity_pass.dart
bdero: Done in eb7a7e48.

packages/flutter_scene/lib/src/scene.dart
bdero: Done in eb7a7e48.

packages/flutter_scene/lib/src/scene.dart
bdero: Done in eb7a7e48.

packages/flutter_scene/lib/src/scene.dart
bdero: Done in eb7a7e48.

packages/flutter_scene/lib/src/scene.dart
bdero: Done in eb7a7e48.

packages/flutter_scene/lib/src/camera.dart
bdero: Done in eb7a7e48.

packages/flutter_scene/lib/src/render/irradiance_field.dart
bdero: Done in a3529b40.

packages/flutter_scene/lib/src/render/irradiance_pass.dart
bdero: Done in eb7a7e48.

examples/smoke_render/integration_test/smoke_test.dart
bdero: Done in eb7a7e48.

packages/flutter_scene/shaders/material_shadow_sampling.glsl
bdero: Done in f15c0973 and eb7a7e48.

packages/flutter_scene/shaders/material_lighting.glsl
bdero: Done in eb7a7e48.

packages/flutter_scene/lib/src/global_illumination.dart
bdero: Done in eb7a7e48.

packages/flutter_scene/shaders/material_engine_lighting.glsl
bdero: Done in eb7a7e48.

packages/flutter_scene/lib/src/components/mesh_component.dart
bdero: Done in eb7a7e48.

packages/flutter_scene/lib/src/material/engine_lighting.dart
bdero: Done in eb7a7e48.

packages/flutter_scene/lib/src/render/velocity_pass.dart
bdero: Done in eb7a7e48.

packages/flutter_scene/lib/src/render/velocity_pass.dart
bdero: Done in eb7a7e48.

packages/flutter_scene/lib/src/render/velocity_pass.dart
bdero: Done in eb7a7e48.

packages/flutter_scene/lib/src/render/taa_pass.dart
bdero: Done in eb7a7e48.

packages/flutter_scene/shaders/flutter_scene_taa.frag
bdero: Done in eb7a7e48.

examples/smoke_render/integration_test/smoke_test.dart
bdero: Done in eb7a7e48.

examples/smoke_render/lib/smoke_scenes.dart
bdero: Removed duplicate ddgi_cornell smoke scene in eb7a7e48.

packages/flutter_scene/CHANGELOG.md
bdero: Done in eb7a7e48.

packages/scene/lib/src/specs.dart
bdero: Added TAA settings to EnvironmentEffectsSpec and recorded in packages/scene/CHANGELOG.md in eb7a7e48.

packages/flutter_scene/shaders/irradiance_receiver.glsl
bdero: Done in eb7a7e48.

examples/flutter_app/lib/example_stress_tests.dart
bdero: Done in eb7a7e48.

PR #369, opened 2026-08-29
Add combat vfx surface and pipeline-stall guards

bdero:
Adds the vfx toolkit a game's combat feedback keeps reaching for, plus two guards against first-draw pipeline stalls.

New surface, `DecalNode` projected box decals, `Scene.screenDistortion` radial shockwave pulses, `.fmat` `blending: additive` with translucent depth state, barycentric wireframe helpers, `MaterialGroup` for fanning one parameter across many materials, and `engine_inputs` on unlit `.fmat` materials (with `GetSceneWorldPosition`, full-resolution depth reads, and `scene_color_reach` capture sharing).

On the guard side, a draw whose pipeline the backend rejects is skipped with a one-time message instead of failing the frame from inside paint, and debug builds report any pipeline build that stalls a frame, naming the material and geometry so the draw can be pre-warmed behind a load screen.


packages/flutter_scene/lib/src/material/preprocessed_material.dart
```diff
@@ -309,52 +323,70 @@ class PreprocessedMaterial extends Material implements HotReloadableFmat {
...
     // in the emitter) to zero. The generated fragment references every
-    // declared parameter resource through it (the MaterialParams block and
-    // any unreferenced samplers) so none can be optimized out; the emitter
-    // declares it whenever the material has any parameter.
-    if (parameters.hasAnyParameters) {
+    // declared resource through it (the MaterialParams block, any unreferenced
+    // sampler, and the engine scene inputs) so none can be optimized out; the
+    // emitter declares it whenever the material has a parameter or an engine
+    // input, so this condition must match.
+    if (parameters.hasAnyParameters || _sceneInputs.isNotEmpty) {
```
bdero: This gate no longer matches the emitter for a material whose only engine input is `planar_reflection`. `_parseSceneInputs` maps `scene_color`, `filtered_scene_color`, and `scene_depth` only, so `_sceneInputs` stays empty, while the emitter now declares and references `FragmentKeepAlive` whenever `engineInputs.isNotEmpty` (and the keep-alive term is never elided). A lit `.fmat` with `engine_inputs: [planar_reflection]` and no `parameters` therefore compiles with a live block that `bind` never binds, which is an unbound uniform block at draw time. The comment just above states that this condition has to match the emitter's, so either fold `planar_reflection` into the bind side or narrow the emitter.

packages/flutter_scene/lib/src/scene_encoder.dart
```diff
@@ -517,18 +543,41 @@ base class SceneEncoder {
...
+        debugContext: () =>
+            '${fmatSourcePathOf(material) ?? material.runtimeType} on '
+            '${geometry.runtimeType}',
+      );
+    } on Exception catch (error) {
+      // A pairing the backend cannot build, most often a custom-attribute
+      // geometry drawn by a material with no vertex stage declaring those
+      // attributes. Throwing out of paint blanks the whole frame, so skip the
+      // draw and say why once.
+      _rejectedPairs.add((geometry, material));
```
bdero: `_rejectedPairs` is never evicted. The declaration, the `contains` above, and this `add` are its only references, so a pair that fails once is skipped for the rest of the process. That collides with hot reload, since `FmatBytesLibrary.refresh` preserves material identity through `updateFromMetadata` and `evictPipelinesForShaders` clears only `_pipelineCache`, so fixing the `.fmat` leaves the mesh invisible until a full restart. The key is also coarser than the pipeline key, and `fragmentShaderForLighting` selects up to four variant shaders, so one failing shadow or cube variant kills the variants that build fine. Keying this the same way as `_pipelineCache` and evicting both together would fix the lifecycle and the granularity in one move. The static set also pins the rejected `Geometry` and `Material` (and their GPU buffers) for the process lifetime.

bdero (commented):
Review of the vfx surface and the pipeline guards. Most of the correctness notes cluster on the new rejection guard in `scene_encoder.dart`, where the lifecycle of `_rejectedPairs` and the scope of the `try` both look wider than intended. The rest are a keep-alive declare and bind mismatch, two `DecalNode` issues, and a `MaterialGroup` contract gap, plus a few writing nits at the end.

packages/flutter_scene/lib/src/scene_encoder.dart
```diff
@@ -517,18 +543,41 @@ base class SceneEncoder {
...
-      vertexLayout: geometry.instancedVertexLayoutFor(
-        material.instanceAttributes,
-      ),
-    );
+    final gpu.RenderPipeline pipeline;
+    try {
+      pipeline = resolvePipeline(
+        material.materialVertexShader(geometry.materialVertexVariant) ??
+            geometry.vertexShader,
+        material.fragmentShaderForLighting(_lighting),
```
bdero: The `try` wraps the argument expressions as well, so shader lookup failures are caught alongside pipeline creation. `Material.fragmentShader` throws `Exception('Fragment shader has not been set')`, which is documented behavior for a `ShaderMaterial` whose shader is assigned late, and that now lands the pair in the permanent `_rejectedPairs` instead of surfacing. A later `setFragmentShader` will not bring the draw back, and `debugPrint` is a no-op in release, so it becomes a silently missing object where master reported it. Resolving the shaders and the layout above the `try` keeps the guard scoped to `createRenderPipeline`.

packages/flutter_scene/lib/src/scene_encoder.dart
```diff
@@ -517,18 +543,41 @@ base class SceneEncoder {
...
+        // A material declaring `instance_attributes` widens the instance-rate
+        // slot, so the pipeline depends on the material as well as the geometry.
+        vertexLayout: geometry.instancedVertexLayoutFor(
+          material.instanceAttributes,
+        ),
+        debugContext: () =>
+            '${fmatSourcePathOf(material) ?? material.runtimeType} on '
+            '${geometry.runtimeType}',
+      );
+    } on Exception catch (error) {
```
bdero: `on Exception` misses the web shim's `Error` throws. `lib/src/gpu/web/render_pipeline.dart` throws `StateError` for a failed `createProgram` and for drawing with shaders that were not compiled, so both escape this catch and blank the frame from inside paint, which is what the comment below says the guard prevents. Native flutter_gpu throws `Exception` for the same class of failure, so the identical bad state is skipped on Metal and blanks the frame on WebGL2. Web link failures do throw `Exception` and are caught, so the gap is just those two sites, but they are the ones retried every frame. Standardizing the shim's throw type is the fix that keeps this out of renderer code.

packages/flutter_scene/shaders/material_scene_inputs.glsl
```diff
@@ -0,0 +1,258 @@
...
+}
+
+// The world-space point on the opaque surface behind this fragment, offset in
+// screen UV. The inverse of ProjectWorldOffsetToScreenUv, reusing the same
+// camera basis and half-fov tangents. Returns this fragment's own world
+// position when depth is unavailable or the camera is not perspective.
+vec3 GetSceneWorldPosition(vec2 uv_offset) {
+  if (frag_info.scene_inputs.y < 0.5 || frag_info.scene_inputs.w <= 0.0 ||
+      frag_info.camera_forward.w <= 0.0) {
+    return v_position;
```
bdero: Returning `v_position` is the graceful fallback for most readers and the pathological one for decals. `decal_inverse` maps the box's own surface to local coordinates sitting exactly on 0.5, so the `step(vec3(0.5), abs(local))` discard in both decal materials lands on the knife edge and round-off splits fragments between discarding and painting coverage on the box faces. Scene depth is bound only under a perspective camera, so an `OrthographicCamera` (editor viewports, 2.5D games) turns every `DecalNode` into a speckled textured box, and nothing in `DecalNode` restricts or documents perspective-only use. Returning a point guaranteed outside the volume, or the huge-depth sentinel `GetSceneDepth` uses, would hide the decal cleanly instead.

packages/flutter_scene/lib/src/decal.dart
```diff
@@ -0,0 +1,165 @@
...
+    return basis
+      ..multiply(Matrix4.rotationY(rotation))
+      ..multiply(Matrix4.diagonal3Values(size, depth, size));
+  }
+
+  @override
+  void scenePrePass(double deltaSeconds, [bool ancestorsVisible = true]) {
+    // The world transform can change without project() running (the node or an
+    // ancestor moved), so the inverse is refreshed from it each frame.
+    if (worldTransformVersion != _writtenTransformVersion) _writeParameters();
```
bdero, line 144: This runs before `super.scenePrePass`, which is what ticks this node's own components and animation player, so a decal moved by its own component writes `decal_inverse` from the pre-tick transform while the box mesh draws at the post-tick one. The mark trails the volume by a frame for as long as it animates. Ancestor-driven motion is fine, since parents tick before recursing into children. Moving the version check and `_writeParameters` below the `super` call fixes it.

packages/flutter_scene/lib/src/material/material_group.dart
```diff
@@ -0,0 +1,124 @@
...
+  /// Removes every material from the group.
+  void clear() => _materials.clear();
+
+  /// The number of materials in the group.
+  int get length => _materials.length;
+
+  /// Sets a float parameter named [name] on every member that declares it.
+  void setFloat(String name, double value) {
+    for (final material in _materials) {
+      if (material.parameters.parameterNames.contains(name)) {
```
bdero: The guard checks the name but not the type, so a member declaring `strength` as a `vec3` passes here and then `setFloat` throws `ArgumentError` out of `_slot`. The loop aborts partway, leaving earlier members written and later ones untouched, which contradicts the class doc promise that one group can span a heterogeneous set. Skipping on a type mismatch, or documenting that it throws, would hold the contract. Worth noting `MaterialParameters.hasParameter` and `hasSampler` already answer this question as an O(1) `containsKey`, where `parameterNames.contains` is a linear scan per member per call.

packages/flutter_scene/lib/src/camera.dart
```diff
@@ -116,6 +115,20 @@ abstract class Camera {
     );
   }
 
+  /// Maps a world-space point to screen UV (origin top-left, `0..1` across
+  /// [viewport]), the same projection as [worldToScreen] normalized instead
+  /// of scaled to pixels. Returns null when [worldPoint] is at or behind the
+  /// camera plane. Values outside `0..1` are a point off screen; callers
+  /// decide whether to clamp or cull.
+  Vector2? projectToScreenUv(Vector3 worldPoint, ui.Size viewport) {
```
bdero, line 119: This and `worldToScreen` are the same projection written twice, same clip transform, same `w <= 0` cull, same NDC mapping, differing only by the final scale. Worth having one delegate to the other, so a later fix (an epsilon on `clip.w`, jitter compensation, viewport insets) cannot land on one copy and leave semantics and picking disagreeing with vfx placement.

packages/flutter_scene/lib/src/screen_distortion.dart
```diff
@@ -0,0 +1,115 @@
...
+  bool enabled = false;
+
+  /// The live pulses, applied in list order. Mutate freely per frame.
+  final List<DistortionPulse> pulses = [];
+}
+
+/// The engine-inserted [CustomRenderPass] that renders
+/// [ScreenDistortionSettings]. Not user-facing; `Scene` adds it when
+/// `scene.screenDistortion.enabled` and at least one pulse is live.
+class ScreenDistortionPass extends CustomRenderPass {
```
bdero, line 64: `ScreenDistortionPass` is public so `src/scene.dart` can construct it, sits in no barrel show list, and its own doc says it is not user-facing, which is exactly the shape that wants `@internal`. `DecalNode.unitBoxGeometry` and `boxTransform` in this same PR already carry it.

MATERIALS.md
```diff
@@ -990,14 +1240,16 @@ sampler budget has room for the two inputs.
 # Render state
 
 A `.fmat` material declares render state in its `material` block: `culling`
-(`back` / `front` / `none`) and `blending` (`opaque` / `alpha`). A
+(`back` / `front` / `none`) and `blending` (`opaque` / `alpha` / `additive`). A
```
bdero: Spaced separating slashes here and on line 420, and in `material_scene_inputs.glsl` line 8. Same pass, the `Label` plus colon comment shape shows up in `flutter_scene_screen_distortion.frag` lines 1 and 7 and in `fmat_emitter.dart` line 321.

packages/flutter_scene/lib/src/material/preprocessed_material.dart
bdero: Folded `planar_reflection` into the bind side, so the condition now also checks `_usesPlanarReflection` and a comment names why it is not part of `_sceneInputs`. Done in 39b2a267.

packages/flutter_scene/lib/src/scene_encoder.dart
bdero: Rejections moved into `_rejectedPipelines`, keyed exactly like `_pipelineCache` and cleared by `evictPipelinesForShaders`, so a hot-reloaded `.fmat` retries and one failing shadow or cube variant no longer takes out the variants that build. Keying on shaders rather than the pair also stops it pinning the `Geometry` and `Material` for the process lifetime. Done in 39b2a267.

packages/flutter_scene/lib/src/scene_encoder.dart
bdero: The shaders and the layout resolve above the guard now, and the new `tryResolvePipeline` wraps only `createRenderPipeline`, so `Exception('Fragment shader has not been set')` surfaces instead of landing in the skip set. A malformed layout's `ArgumentError` surfaces for the same reason. Done in 39b2a267.

packages/flutter_scene/lib/src/scene_encoder.dart
bdero: Took the layering fix rather than widening the catch, so both pipeline-creation sites in the web shim throw `Exception` now, matching native flutter_gpu. The buffer-creation `StateError` further down is outside the guarded path, so I left it alone. Done in 39b2a267.

packages/flutter_scene/shaders/material_scene_inputs.glsl
bdero: It returns a point at the same far sentinel `GetSceneDepth` reports (now named `kSceneDepthUnavailable` and shared by both), so an unavailable depth puts the sample well outside any projection volume instead of exactly on its boundary. `DecalNode` and MATERIALS.md now say a decal draws nothing under an `OrthographicCamera`, and the shader lock test asserts the shared constant and the unprojection that reuses it. Done in 39b2a267.

packages/flutter_scene/lib/src/decal.dart
bdero: Done in 39b2a267.

packages/flutter_scene/lib/src/material/material_group.dart
bdero: Added `MaterialParameters.hasParameterOfType` as an `@internal` O(1) lookup, and every group setter plus `writeDecalParameters` gates on it, so a mistyped member is skipped instead of throwing partway through. The class doc now promises that explicitly, with tests over the gate and the decal writer. Done in 39b2a267.

packages/flutter_scene/lib/src/camera.dart
bdero: `worldToScreen` delegates to `projectToScreenUv` and scales the result, so there is one projection and the existing agreement test guards it. Done in 39b2a267.

packages/flutter_scene/lib/src/screen_distortion.dart
bdero: Done in 39b2a267.

MATERIALS.md
bdero: Fixed both spaced slashes in MATERIALS.md and the one in `material_scene_inputs.glsl`, and rewrote the `Label` plus colon comments in `flutter_scene_screen_distortion.frag` and `fmat_emitter.dart`. Done in 39b2a267.

Issue #323, opened 2026-08-17
A rejected triangle mesh collider disappears without a diagnostic


bdero:
`ColliderBuilder::trimesh` rejects degenerate input. `fsr_collider_trimesh` returns `u64::MAX` when it does, the native bindings map that to null, and `RapierWorld` turns the null into an empty handle list.

The result is a `Collider` that mounts without complaint, holds no handles, and collides with nothing, and no message says why. The symptom surfaces much later as bodies falling through geometry that looks like it has a collider on it.

Report the rejection in debug builds, naming the node the collider is attached to. The convex hull path returns a nullable handle through the same code and drops it the same way, so both want the same treatment.

bdero:
Done. A collider whose shape the backend refused now throws naming the backend, the shape type, the node it sits on, and the degenerate cases that cause a rejection, instead of mounting inert and colliding with nothing. The convex hull and compound paths are covered too, since every null handle comes back as an empty list. Rides the 0.24.0 release.

Issue #322, opened 2026-08-17
Triangle mesh colliders catch on internal edges and keep duplicate vertices


bdero:
`fsr_collider_trimesh` builds its shape with `ColliderBuilder::trimesh` and no preprocessing flags, so a triangle mesh collider keeps whatever the source mesh gave it.

That costs two things. Vertices split at UV or normal seams stay duplicated, so the collider carries more vertices than the surface needs. More seriously, a body moving across the mesh catches on the shared edges between adjacent triangles and picks up contacts pointing the wrong way, which stops or launches characters that should slide smoothly. Imported terrain is where this shows up, and imported terrain is what triangle mesh colliders are for.

Rapier addresses both through `ColliderBuilder::trimesh_with_flags`, with `MERGE_DUPLICATE_VERTICES` for the first and `FIX_INTERNAL_EDGES` for the second. Since the flags change the cooked shape rather than the Dart-side data, applying them unconditionally is probably better than adding a knob.

Touching the native shape construction means a rebuild of the prebuilt binaries and the wasm, so this rides a `flutter_scene_rapier` release.

bdero:
Done. `fsr_collider_trimesh` cooks with `trimesh_with_flags` and `MERGE_DUPLICATE_VERTICES | FIX_INTERNAL_EDGES`, applied unconditionally rather than behind a knob. The prebuilt binaries and wasm carrying it ride the flutter_scene_rapier 0.5.2 release.

PR #372, opened 2026-08-30
Editor UX and performance sprint with the lighting work behind it

### LINKED ISSUE #244: Scene.initializeStaticResources() swallows load failures and surfaces a misleading downstream error
When `loadBaseShaderLibrary()` or `Material.initializeStaticResources()` fails, `Scene.initializeStaticResources()` logs the error through `dart:developer` `log()` (not visible on web), resets the memoized future, and returns a future that **completes normally** (`lib/src/scene.dart`, ~lines 328-352). So `await`ing it succeeds despite the failure, and the developer instead hits the `baseShaderLibrary` getter throwing 'The base shader bundle has not been loaded yet. Await Scene.initializeStaticResources()...', which points at the thing they already did and hides the real cause.

Fix: rethrow from `initializeStaticResources()` so an awaiting caller sees the failure, and/or retain the load error and rethrow it from the getter through a web-visible path.

bdero:
Resolves #244, #322, and #323.

The editor UX and performance sprint together with the lighting work it depended on. Multi-selection editing, in-place graph reflection for undo and reparents, outliner scrolling at scale, live component previews, and editor camera and selection carried in the fscene document.

On the engine side, punctual lights shade through per-view froxel clustering, point lights cast shadows, and `Node.shadowCastingMode` selects how a node casts. The stage inspector gains the global illumination, anti-aliasing, and shadow-budget controls the engine already had but never exposed.


packages/flutter_scene/lib/src/fscene/realize/realize.dart
```diff
@@ -19,6 +20,17 @@ import 'package:flutter_scene/src/skin.dart';
 /// authored decomposition; recovering it from the composed matrix puts a
 /// mirrored axis's negative scale on X, which breaks animation blending
 /// on mirrored bones. Also used by scene hot reload.
+/// The [ShadowCastingMode] a document's `shadowCastingMode` name realizes to.
+///
+/// An unknown or absent name resolves to [ShadowCastingMode.on], so a
+/// document written by a newer engine still loads.
+/// {@category Assets and loading}
+ShadowCastingMode shadowCastingModeFromName(String name) =>
```
bdero: This function landed between `applyTransformSpec`'s doc comment and `applyTransformSpec` itself, so the `///` block starting at line 19 (TRS decomposition, mirrored bones, hot reload) now runs straight into the new doc and documents `shadowCastingModeFromName` instead. `applyTransformSpec`, which is public API, is left with no doc comment at all, and this symbol's dartdoc opens with unrelated TRS text. Move the new function above line 19 or below `applyTransformSpec`.

packages/flutter_scene_editor/lib/src/controller/editor_controller.dart
```diff
@@ -1507,6 +1713,186 @@ class EditorController extends ChangeNotifier {
...
+    // Build the whole restored forest detached, so any bail below leaves the
+    // live graph untouched and the full realize can take over.
+    final nodes = <LocalId, Node>{};
+    final context = RealizeContext(document, resources: realizer)
+      ..resolveNode = (id) => nodes[id] ?? _liveById[id];
+    for (final spec in restored.values) {
+      final node = tagNodeId(
+        Node(name: spec.name)
+          ..layers = spec.layers
+          ..visible = spec.visible,
```
bdero: The restored node is rebuilt without `shadowCastingMode`, so undoing a delete on a node set to `off`, `doubleSided`, or `shadowsOnly` brings it back as `on`. The document still says otherwise, so the live scene and the document disagree until something forces a full re-realize. `_realizeWith` in `realize.dart` sets the mode alongside `layers` and `visible`; this path needs the same line.

`_reflectAddedNode` around line 1668 has the same gap, which a duplicate or paste of a node with a non-default mode hits.

```suggestion
          ..visible = spec.visible
          ..shadowCastingMode = shadowCastingModeFromName(
            spec.shadowCastingMode,
          ),
```

packages/flutter_scene/lib/src/scene.dart
```diff
@@ -2119,17 +2253,23 @@ base class Scene implements SceneGraph {
...
+        pointShadowFrame != null) {
       graph.addPass(
         ShadowPass(
           renderScene: renderScene,
           cascades: effectiveCascades,
           tileResolution: cascades.isNotEmpty
               ? light!.shadowMapResolution
-              : spotShadowFrame!.tileResolution,
+              : spotShadowFrame?.tileResolution ??
+                    pointShadowFrame!.casters.first.light.shadowMapResolution *
```
bdero, line 2292: `PointLight.shadowMapResolution` is documented as "Pixel resolution of each (square) cube face's depth tile" and is exposed as an editor control with a `PowerOfTwo(64, 4096)` constraint, but it only reaches the atlas here, and only when there is no directional cascade and no shadow-casting spot. In the common case (a sun plus a shadowed point light) the tile size is the directional light's and every cube face renders at `_tileResolution ~/ 2` in `ShadowPass`, so changing the point light's value does nothing.

Two more first-caster-wins cases ride along. Only `casters.first` is read here, so a second point caster's resolution is ignored even when there is no directional light, and `PointShadowFrame.casterFaces` takes `casters.first.light.shadowCasterFaces` for every caster while the codec exposes it per light.

The shader disagrees with all of this. `punctual_lights.dart` stamps `params[base + 21] = 1.0 / light.shadowMapResolution` and `SamplePointShadow` reads it as `p5.y` for its face half-texel clamp, so the clamp is only right when the per-light value happens to equal half the tile size (which the defaults, 1024 directional and 512 point, do by coincidence). Deriving the half-texel from the atlas tile size the faces actually rendered at, and documenting the per-light field as advisory, would keep the two in step.

packages/flutter_scene/lib/src/components/mesh_component.dart
```diff
@@ -1,10 +1,18 @@
...
 import 'package:flutter_scene/src/node.dart';
 import 'package:flutter_scene/src/render/render_scene.dart';
 
+// A primitive's own shadow opt-out only ever subtracts, so it collapses the
+// node's mode to off rather than choosing between the casting modes.
+ShadowCastingMode _effectiveShadowMode(
+  MeshPrimitive primitive,
+  ShadowCastingMode nodeMode,
+) => primitive.castsShadow ? nodeMode : ShadowCastingMode.off;
+
```
bdero: The comment says a primitive opt-out "only ever subtracts", which holds for `on` and `doubleSided` but not for `shadowsOnly`. Collapsing `shadowsOnly` to `off` stops the primitive casting (right) and also starts it drawing into the color image (wrong), since `RenderItem.drawsColor` only excludes `shadowsOnly`. So a primitive with `castsShadow == false` under a shadows-only node becomes visible, which is the opposite of what the node asked for. `shadowsOnly` wants a collapse that keeps the primitive out of color as well.

packages/flutter_scene/lib/src/render/punctual_lights.dart
```diff
@@ -257,10 +384,280 @@ class PunctualLightBuffer {
...
+  static const int froxelCountX = 16;
+  static const int froxelCountY = 9;
+  static const int froxelCountZ = 16;
+  static const double _froxelNear = 0.25;
+  static const double _froxelFar = 100.0;
+  static const int _froxelTexWidth = 1024;
+
+  // Several views can build froxels in one frame (screen views, probe
+  // faces), so this ring is deeper than the params rings.
+  final _TextureRing _froxelRing = _TextureRing(size: 8);
```
bdero, line 431: Two things about the froxel ring and the cache above it.

The ring is 8 deep but a frame can acquire more than that. Every `ScenePass` builds froxels, so a frame with a reflection-probe capture (six faces) plus two or more `SceneView`s reaches eight acquisitions and hands a texture back out on the next frame while the previous one may still be in flight. That is exactly what the ring exists to prevent; the params ring gets three frames of headroom because it acquires once per frame, and this one gets about one.

The single-slot cache does not soften that, because every build overwrites `_cachePosition`/`_cacheForward`/`_cacheTanX`. With more than one view per frame the cache never hits, so each view pays a full refroxelization every frame and still consumes a ring slot. Keying the cache per view would fix the caching and cut the acquisition rate at the same time.

packages/flutter_scene_editor/lib/src/io/scene_io.dart
```diff
@@ -349,7 +350,59 @@ Future<void> saveFscene(EditorController controller, String path) async {
...
+  final editorState = controller.editorStateProvider?.call();
+  if (editorState != null) controller.document.editor = editorState;
+  // Payload bytes live in the sidecar, not the lean text, so a session that
+  // touched the payload pool (a mesh split, an embedded import) must rewrite
+  // it or those bytes are lost on reopen. Name the sidecar before writing
+  // the text so the manifest references it.
+  final document = controller.document;
+  final rewriteSidecar =
+      controller.payloadsDirty && document.payloads.isNotEmpty;
+  if (rewriteSidecar && document.payloadSource == null) {
```
bdero: `payloadSource` is only filled in when it is null, and open decodes it from the document, so a Save As keeps the source document's sidecar name. Saving `foo.fscene` as `bar.fscene` in the same directory writes `bar`'s payload pool over `foo.payloads.fsceneb`, and `foo.fscene` afterwards reads a sidecar it did not write. Deriving the name from `path` whenever the stem no longer matches would keep a Save As off the original document's bytes.

Smaller thing on the same block, `payloadsDirty` is only cleared inside `if (rewriteSidecar)`, so a save that empties the payload pool leaves the flag set forever.

packages/flutter_scene/CHANGELOG.md
```diff
@@ -8,6 +8,28 @@
...
+* `Scene.punctualLightOverflowCount` reports how many drawable items dropped punctual lights last frame because more lights reached them than the per-object budget shades.
+* A `-split<N>` suffix on a glTF node name splits its mesh into grid-cell child meshes at import (whole triangles binned by world-space centroid into `N`-unit cells, children named `Ground_x0_z3`), so a level-spanning mesh culls and receives punctual lights per cell; the split reapplies on every re-import.
+* Punctual lights shade through per-view froxel clustering (screen tiles crossed with exponential depth slices, each shading only the lights that reach it), removing the per-object light cap, so a large mesh reached by many lights shades them all and no draw carries light state. Perspective views cluster automatically; orthographic views and frames using light channel masks keep the per-object lists, and `Scene.punctualLightClustering` disables clustering for comparison.
+* The punctual light loop is dynamically bounded (every compiled shader dialect is GLSL ES 3.00 or newer), so drivers cannot unroll it, and a froxel holds up to 255 lights.
+
+## 0.23.1
```
bdero: 0.23.1 was never cut (pub.dev's latest and the newest tag are both 0.23.0), so everything under this heading will actually ship as part of 0.24.0. It also sits below the 0.24.0 block, which puts the `Scene.initializeStaticResources()` BREAKING entry under a version nobody upgrades to and out of the section an upgrader reads. Folding these bullets into 0.24.0 keeps the breaking note where it will be seen.

bdero (commented):
A few things worth fixing before this lands. Nothing here blocks the design, they are mostly places where a new field does not reach every path that rebuilds a node, or where a documented control does not actually control anything.

The two I would fix first are the restored-node reflect dropping `shadowCastingMode` (an undo silently changes shadow behavior) and `PointLight.shadowMapResolution` being inert whenever a directional light is present, since the editor now exposes it as a control.

packages/flutter_scene/lib/src/fscene/realize/realize.dart
bdero: Moved below `applyTransformSpec` in df1b26ec, so its doc comment reaches it again and this symbol's dartdoc starts with its own text.

packages/flutter_scene/lib/src/components/mesh_component.dart
bdero: Fixed in df1b26ec. `RenderItem` carries `primitiveCastsShadow` alongside the node's mode now, so the opt-out only subtracts casting and a shadows-only node stays out of the color image. My test had asserted the buggy behaviour as intended, so it is rewritten to cover both this case and a primitive opting out under a drawing node.

packages/flutter_scene_editor/lib/src/controller/editor_controller.dart
bdero: Both paths carry the mode now (df1b26ec), and `editor_controller_restore_test.dart` covers the undo case. Worth noting that test cannot run in CI, which builds only `scene`, `flutter_scene`, and `flutter_scene_net`.

packages/flutter_scene/lib/src/scene.dart
bdero: Fixed the correctness half in df1b26ec. The atlas tile size resolves once per frame now and the point rows stamp the face resolution the faces actually got, so the shader was clamping its PCF kernel with the wrong texel size whenever a directional cascade set the tile. The first-caster-wins limitation stands, so `PointLight` and `SpotLight` now say plainly that the shared atlas honors one tile size, with a TODO(shadow-atlas-per-light-tiles) for making it per-light.

packages/flutter_scene/lib/src/render/punctual_lights.dart
bdero: The ring sizes itself now (df1b26ec): `build` marks the frame boundary and the ring grows to the busiest frame's acquisitions times three, matching the headroom the params rings get, so a probe capture plus several views cannot outrun it. No test, since the ring is private and needs a GPU to allocate.

packages/flutter_scene_editor/lib/src/io/scene_io.dart
bdero: Both fixed in df1b26ec. The sidecar name derives from the file being written whenever it differs, so a Save As stops writing over the source document's bytes, and `payloadsDirty` clears unconditionally. Untested, there is no save-path harness in the package yet.

packages/flutter_scene/CHANGELOG.md
bdero: Folded into 0.24.0 in c9a22934, so the breaking note sits in the section an upgrader reads.

PR #376, opened 2026-08-31
Networking: replicate what a component already declares

ChxisB:
Third of five stacked PRs, after #374 and #375.

**On reading this diff:** the branch carries the earlier commits too. **The new
work is the last commit only** (`Networking: replicate what a component already
declares`).

---

`flutter_scene_net` gains the half that makes an object networkable without
writing a replication class for it.

### The idea

A networked object already describes itself. Every component has a codec, and
every codec declares its properties with a name, a kind, a `read` that gets the
value off the live component and a `write` that puts one back. That is exactly
what a replicated field needs — so the replica is built from the schema rather
than repeating it. A hand-written `Replica` per type is the same information
twice, and the second copy is the one that goes stale when a property is renamed.

`NetworkIdentityComponent` is the authoring surface: a type key, the properties
that replicate with their write authority and read scope, the events the object
can send, and what may be done with its ownership. **The list lives in the
document**, which is what makes both ends agree on wire order without either
being told — and dashwire's schema hash catches them at connect if they don't.

### Events, and why the direction matters

Properties carry what is continuously true; events carry what happened once. A
property that flickered to record a gunshot is one you could miss between two
snapshots. The direction an event travels is the security boundary — an event
aimed at the server is a client *asking*, and the server decides.

### Ownership and sessions

An object declares what is *permitted* (distributable, transferable,
requestRequired, sessionOwner); declaring nothing means static. Sessions cover
host, server and client, with connection approval that runs **before** anything
is spawned — deciding after spawning means un-spawning, which is where the
half-joined states live.

### A document can name the replica that drives a node

A replica belongs to a running session rather than to a file, so the document
names a **slot** and the app resolves it. A slot nobody supplies leaves the node
undriven rather than failing the load — a networked scene should still open
offline.

### Prediction can bound what a retained tick costs

Serializing the whole world every tick is always correct and always costs the
whole world — 64 retained ticks of a scene that mostly isn't moving. A controller
implementing `PredictedBodyScope` names the bodies that can matter and a tick
keeps 13 floats each.

The set must be closed under interaction, which is the closure a solver island
describes. Asked for rather than derived, because `PhysicsSimulation` has no
island query and a subtly wrong one would be worse than the cost it saved.

### Lag compensation can rewind between two ticks

A client renders *between* server ticks, so rewinding to the nearer one puts a
target up to half a tick from where the shooter saw it — at 30Hz and a running
speed, most of a body width. `rewindBetween` restores the earlier tick and nudges
the tracked bodies toward the later one. Rotations blend the short way; a
fraction outside the interval is clamped rather than extrapolated, since
extrapolating puts a target where neither end says it was.

`renderedAt` turns a peer's acknowledged tick plus its render delay into what it
was actually showing — the newest tick a client *received* is not the tick it was
*displaying*, and compensating to the acked tick alone still favours the shooter.

### Testing

185 tests, analyze clean, **against `dashwire_replication` exactly as published**
— no overrides, nothing unreleased.

### Two things deliberately left out

They need transport changes that are open and unmerged, so they are marked with
TODOs naming the PR rather than shipped half-working:

- Cancelling the transform component's replica subscriptions on detach, which
  needs bdero/dashwire#7 (`Rep` listener removal). Today those subscriptions —
  and through them the component — live as long as the replica.
- An event target meaning *everyone except the owner*, which needs
  bdero/dashwire#6. Note `RpcTarget.others` excludes the **server**, not the
  owner; the two are easy to confuse.

bdero/dashwire#8 (a peer's acknowledged snapshot tick) would let `renderedAt` be
fed without a game stamping its own render tick into every input payload.


ChxisB:
Superseded by #379, which carries the same work as a single commit based directly on `master` — mergeable on its own rather than sitting behind #374 and #375. Closing this one; the stacked series continues at #377.

PR #392, opened 2026-09-09
Keep one pooled color target per depth attachment setup

### LINKED ISSUE #387: [linux/arm64] fix: Z-sorting doesn't work unless MSAA is active
When running the `example_app` (car view) and disabling MSAA from the right-hand settings menu, the rendered surfaces appear to not be Z-sorted correctly. It's also possible I'm misinterpreting this and the back-faces are being shown instead, but that wouldn't fully explain the visual phenomena. Attaching video for reference:

https://github.com/user-attachments/assets/d8c57f0e-0849-4a2b-a181-9de0916ea617

Test environment:
- Raspberry Pi 5, 16GB
- RaspberryPi OS / Debian 13 Trixie
- linux/arm64, glibc 2.41
- VideoCore VII GPU, supporting OpenGL ES 3.1, Vulkan 1.2
- Flutter (Channel main, 3.48.0-1-0.pre-676)

bdero:
Resolves #387.

The OpenGL ES backend caches one framebuffer per color texture and attaches depth only when that framebuffer is first created, so a pooled scene color texture that served as the MSAA resolve destination and then as a direct target after switching `Scene.antiAliasingMode` rendered with no depth test at all. Pool descriptors gain an `attachmentKey` naming the depth setup a color target is rendered with, so each setup keeps its own texture; the scene color, alternate scene color, linear depth, and velocity targets set it. A smoke probe renders the switch and samples the frame on every lane.


bdero:
@kerberjg if you have some cycles, I'd appreciate if you could give this a try to see if it fixes the #387 bug on the Raspberry Pi 5.

PR #403, opened 2026-09-13
Add PointerLock for first-person mouse look

bdero:
Adds `PointerLock`, a built-in cursor lock for first-person mouse look, so games no longer need a third-party package or per-app native code.

`PointerLock.instance.lock()` hides and holds the cursor, `movement` reports each frame's mouse movement in logical pixels, and listeners hear when the lock ends (unlock, focus loss, or Esc on the web). The desktop backends are small native libraries compiled by flutter_scene's build hook and bound through FFI, covering macOS, Windows, and Linux on X11 and Wayland; the web uses the Pointer Lock API. The stress test example's focus mode now uses it in place of `native_mouse_cursor`.


bdero:
Superseded by #404, which moved PointerLock into flutter_scene_input.

Issue #371, opened 2026-08-29
Very low FPS on web


RawadZogheib:
When we launch on the web, we get very low FPS and noticeable lag. Do we need to add or optimize anything specifically for web, or is the flutter_scene not fully supported/optimized for web yet?

bdero:
Flutter Scene fully supports web. Where are you trying to run? Some of the content in the example app is tuned for high end/desktop hardware.

bdero:
https://fscene.dev/getting-started/ is also filled with live examples of flutter_scene running on web.

PR #407, opened 2026-09-13
Replace a re-bound vertex slot in the web shim instead of appending it

### LINKED ISSUE #406: Web backend: per-draw vertex bindings accumulate across a same-pipeline run, making GL traffic quadratic in the draw count
### What happens

On the WebGL2 backend, `RenderPass.bindVertexBuffer` appends to `_pendingVertexBindings` and the list is only cleared by `clearBindings()`, which the scene encoder calls only when the pipeline changes (`scene_encoder.dart`, "Opaque draws are pipeline-sorted, so same-pipeline runs skip the clear"). Every draw binds its full slot set, so over a run of N same-pipeline draws the list grows to N entries and `_applyVertexState` re-points all of them on every draw:

- `vertexAttribPointer` / `bindBuffer` / `vertexAttribDivisor` calls per frame ≈ attributes × N² / 2 instead of attributes × N
- the VAO cache key (`_applyVertexState`) is built from the whole list, so as soon as a run outgrows `_kMaxCachedVaos` (512) the cache misses on every draw and a VAO is created per draw per frame

Native is unaffected because there each draw's binds overwrite the slot.

Any scene with a few hundred individually modelled meshes sharing a material hits this (they can't be instance-batched because the geometries differ). It may be what #371 was seeing.

### Reproduction

https://github.com/Pregum/flutter_scene_web_draw_repro — 400 cubes with distinct `CuboidGeometry`, one shared `UnlitMaterial`, no lights/sky, `SceneView` with an orbiting camera. The README has a DevTools snippet that counts GL calls per presented frame.

Chrome 140 / Apple M4, `flutter_scene 0.23.0`, Flutter 3.47.2 stable:

| | draws / frame | `vertexAttribPointer` / frame | `bindBuffer` / frame | Dart build phase |
|---|---|---|---|---|
| 0.23.0 | 402 | **401,006** | 80,612 | **~80 ms** |
| with the fix below | 402 | 2,006 | 812 | ~3 ms |

`DRAWS=100/200/400` scales the `vertexAttribPointer` count quadratically. In my real app (552 draws, 35 buildings) the main thread was saturated (~35–60 ms build, `_applyVertexState` at the top of the profile) and drops to ~5 ms with the fix.

### Fix

Replace the pending entry for a slot that is re-bound instead of appending it (PR: #TBD). One hunk in `lib/src/gpu/web/render_pass.dart`:

```dart
final entry = (bufferView, slot, instanceRate);
final existing = _pendingVertexBindings.indexWhere((b) => b.$2 == slot);
if (existing >= 0) {
  _pendingVertexBindings[existing] = entry;
} else {
  _pendingVertexBindings.add(entry);
}
```

Pregum:
Fixes #406.

The encoder clears the web `RenderPass`'s bindings only when the pipeline changes, and every draw binds its full slot set, so over a run of same-pipeline draws `_pendingVertexBindings` grew by one entry per draw. Each draw then re-pointed every earlier draw's attributes (quadratic GL traffic in the run length), and because the VAO cache key covers the whole list, the cache missed as soon as a run outgrew it and rebuilt a VAO per draw per frame. Native is unaffected: a re-bind there overwrites the slot.

`bindVertexBuffer` now replaces the pending entry for the slot it re-binds. The instance-rate entries get the same treatment, which is what keeps the per-draw instance stream at one re-point per draw.

Measured on the reproduction in the issue (400 distinct cubes, one material, Chrome 140 / Apple M4): `vertexAttribPointer` per frame 401,006 → 2,006, `bindBuffer` 80,612 → 812, Dart build phase ~80 ms → ~3 ms; a 552-draw scene went from a saturated main thread to ~5 ms.

Verified against the 0.23.0 sources with this exact diff (path override). Note: building the reproduction against current `master` fails at runtime on web for me with `Failed to bind uniform ... "DebugViewInfo"` for the unlit material, independent of this change — happy to file that separately if it isn't known.

CHANGELOG: one line under 0.24.0.


bdero:
Thanks @Pregum, great find with a clear repro. Pushed a small follow-up commit to tighten the prose and merged. This regression dates back to 0.19.0, so it may also explain what happened with #371.
Please do file the DebugViewInfo failure separately with your Flutter version; master's WebGL2 smoke lane passes on Flutter master, so it may be specific to 3.47 stable. But it's possible we also have a testing gap. We've already uncovered and fixed a couple of cases where the WebGL backend can fall over on specific browser builds...

PR #408, opened 2026-09-14
Revalidate generated assets on web so a cached bundle cannot outlive a rebuild

bdero:
Related to #407.

On web, shader bundles and their generated JSON live at fixed asset URLs, so a browser cache could pair a previous build's bundle with new Dart code and fail every draw with a missing `DebugViewInfo` block. These assets now load with `no-cache` on web, and the shim's missing-binding errors mention a stale cache as a possible cause.


Pregum:
Thank you for reviewing and merging #407 so quickly, and for the follow-up commit on the wording — much appreciated.

You asked me there to file the `DebugViewInfo` failure with my Flutter version. I dug into it and it turns out to be exactly what this PR fixes, so I don't think a separate issue is needed. Sharing what I found in case it's useful.

**Not specific to Flutter 3.47.2 stable.** A clean build of master on 3.47.2 stable renders fine on web. What I had hit on the 14th was a cached bundle paired with new Dart code.

**Deterministic reproduction** (puppeteer with a persistent profile, Chrome 140 / macOS, Flutter 3.47.2 stable, using the repro app from #406):

1. Visit a build made with 0.23.0, served with an old `Last-Modified` (so Chrome's heuristic freshness keeps the assets for hours).
2. Deploy a build of master at `dce794be` (just before this PR) to the same URL and refresh only the JS.
3. Both shader bundles come back `fromCache=true` with the 0.23.0 bytes, and every draw throws `Failed to bind uniform … "DebugViewInfo"` (800+ times in ~10 s).

Same steps with master including this PR: both bundles are refetched (`fromCache=false`, 3.2 MB / 6.9 MB) and the scene renders. So the `no-cache` change does fix the browser case.

**Why the pairing was possible:** the generated file name's tag is `shortHash('shaderBundle/base@<engine identity>|<target>')`, so it changes with the Flutter SDK and target but not with the bundle's bytes. On the same SDK, 0.23.0 and master both write `shaderbundle.base.3b1f1728.shaderbundle` (1.2 MB vs 3.2 MB), so any HTTP cache in between can serve the old one under the new build.

If it would be welcome, I'd be happy to try a follow-up that hashes the bundle content into the file name (or appends the manifest `stamp` to the request URL), which would also make CDN/edge caches safe and keep the bundles cacheable. No worries if you'd rather keep it as is — and thanks again for turning both fixes around so fast.

Issue #391, opened 2026-09-09
[linux/arm64] fix: Long loading times on Raspberry Pi 5 (several minutes)


kerberjg:
When loading `example_app`, the system will become unresponsive and the Flutter view will be frozen displaying a `CircularProgressIndicator` - for several minutes. While I wasn't yet able to exactly time the loading process, all of my attempts have been somewhere within the 2-5 minutes OoM.

To make sure this is related to scene/asset loading, I re-checked loading another scene after the app is already open: got similar loading times, meaning this isn't related to Flutter itself.

What's odd is that the CPU doesn't seem to be maxed-out, but rather sitting at 20-70%. But the entire system being unresponsive is really worrying. Similarly to #388, there's just something about `flutter_scene` the OS/hardware really doesn't like - when stress testing non-3D Flutter apps on the same unit the system does not freeze.

For context, active cooling is installed on my unit.

Test environment:
- Raspberry Pi 5, 16GB
- RaspberryPi OS / Debian 13 Trixie
- linux/arm64, glibc 2.41
- VideoCore VII GPU, supporting OpenGL ES 3.1, Vulkan 1.2
- Flutter (Channel main, 3.48.0-1-0.pre-676)

bdero:
Which particular scenes were you having trouble with?
The particles example generates frame flipbooks + all of the geometry/textures at runtime, and so I would expect the load time to be very large for that one. Some of the other examples download GLBs from the internet.

kerberjg:
I will take some time to go through them one-by-one tomorrow, but going from memory it was anything that wasn't just a primitive shape. The ones that did download GLBs would be stuck _after_ the download said to be complete (which wouldn't be a problem anyway, I have 1Gbps symmetric + wired). I noticed they all get stuck between lines of "Unpacking glTF".

Some scenes (like Sponza) would fail to load altogether, with a flurry of "Failed to create texture" messages - I can provide more accurate logs tomorrow.

bdero:
Thanks. Another thing I noticed is that the videos appear to be in debug mode -- based on what you've provided, my guess would be that the bottleneck is almost entirely CPU, and so switching from debug to release mode should help some as well.

[...]

PR #410, opened 2026-09-15
Ship each engine shader bundle once and only for its platform

bdero:
Related to #391.

Every app shipped flutter_scene's engine shader bundles for each platform the shared pub cache had built (a macOS build after an Android build carried the Android set), and twice when data assets were enabled, since some hook passes wrote the generated tree while another registered data assets. The hook now always writes the tree, each target's bundles live in their own directory listed with a `platforms:` filter, and the runtime drops the unused data-asset lookups.

`dart run flutter_scene:init` now lists the same per-platform directories in an app's pubspec, so an app's own generated shaders get the same treatment; the workspace apps are migrated.


packages/flutter_scene/lib/src/generated_assets/generated_tree.dart
```diff
@@ -393,6 +448,13 @@ final class GeneratedAssetTree {
 
   static String _digest(String stamp) => fnv1aHex(utf8.encode(stamp));
 
+  // [generatedNameWithoutTag] for a tree-relative path, keeping its directory.
+  static String? _withoutTag(String relative) {
+    final slash = relative.lastIndexOf('/');
+    final name = generatedNameWithoutTag(relative.substring(slash + 1));
+    return name == null ? null : '${relative.substring(0, slash + 1)}$name';
```
bdero: [P1] Preserve fresh outputs across target directories

Keeping the directory prefix here breaks the one-hour race guard in `save()`. With two tree snapshots opened from the same empty shared directory, saving a Metal output and then a web output deletes the still-fresh Metal file because the two `_withoutTag` values have different prefixes. A parallel build can then bundle a file another target removed. Match variants by generated basename across target directories, or retain fresh generated files independently of the directory.

packages/flutter_scene/lib/src/generated_assets/build_engine_assets.dart
```diff
@@ -42,21 +41,18 @@ const String _baseBundleManifest = 'shaders/base.shaderbundle.json';
...
-  buildInput: buildInput,
-  buildOutput: buildOutput,
-  // Data assets are the tidiest home when they exist, no writes into the
-  // package directory at all. One build runs this hook several times with
-  // different asset types though, so a run that has them must not delete the
-  // tree copy a run without them wrote.
-  dataAssets: buildInput.config.buildDataAssets,
-  prune: false,
-);
+}) => _build(buildInput: buildInput, buildOutput: buildOutput);
```
bdero, line 55: [P2] Remove the obsolete data-assets workaround

`buildOwnEngineAssets` no longer consults `buildDataAssets`, but `GeneratedAssetsNotWritableException` still tells users to enable Dart data assets and `engine_stamp_test.dart` locks that advice in. Enabling the flag cannot change this path now. Update the exception and test so write failures recommend only remedies that still work.

bdero (commented):
Two issues to fix before merging.

PR #411, opened 2026-09-16
Add orthographic cameras

bdero:
Adds `OrthographicCamera` and `OrthographicProjection`, sized by `OrthographicSize` (height, width, contain, cover, stretch, or pixels per unit) with `zoom`, a world-unit `offset`, and a `near` that may be negative for isometric views.

Every depth-reconstructing effect (shadows, AO, SSR, TAA, depth of field, god rays, GI, planar reflections, froxel lighting, LOD, custom-pass depth and normals) now reads the projection from its matrix through one shared `view_projection.glsl`, so orthographic and custom projections work instead of being skipped. `.fscene` cameras serialize the orthographic projection and the editor draws its view volume through a new `GizmoOrthographicVolume` in `scene`.


packages/flutter_scene/shaders/view_vector.glsl
```diff
@@ -0,0 +1,29 @@
...
+
+vec3 ViewVector(mat4 camera_transform, vec3 camera_position,
+                vec3 world_position) {
+  vec3 to_camera = camera_position - world_position;
+  if (!CameraTransformIsOrthographic(camera_transform)) {
+    return to_camera;
+  }
+  vec3 forward = normalize(vec3(camera_transform[0][2], camera_transform[1][2],
+                                camera_transform[2][2]));
+  return forward * dot(to_camera, forward);
```
bdero: Under an orthographic camera this is `forward * planarDepth`, so `dFdx`/`dFdy(v_viewvector)` in `CotangentFrame` (normals.glsl) are both parallel to forward and T and B collapse onto `cross(forward, N)`. Normal maps on meshes without authored tangents only perturb along one axis, and on camera-facing surfaces T and B go to zero and the map drops out. Anisotropy's `TangentFrame` call has the same problem.

bdero: This keeps the signed planar depth, so for fragments between a negative near plane and the eye plane the vector points away from the viewer. Lighting normalizes it (`camera_normal` in material_lighting.glsl, `GetViewDirection`, the fog sky sample), so N.V clamps and specular, Fresnel, and IBL reflection read the wrong hemisphere there. The `orthographic_camera` smoke scene puts most of its ground and cubes on that side of the eye. Billboards behind the eye plane also get mirrored, and `ApplyDepthBias` pushes away from the viewer. Splitting the view direction from the signed depth varying would fix this and the cotangent frame issue together.

packages/flutter_scene/lib/src/render/planar_reflection.dart
```diff
@@ -173,19 +185,12 @@ class PlanarReflectionCamera extends Camera {
...
-              fovRadiansY: base.fovRadiansY,
-              near: base.near,
-              far: base.far,
-            ),
-            worldPlane: plane,
-            viewMatrix: _inner.getViewMatrix(),
-            clipBias: clipBias,
-          )
-        : base;
+    _projection = ObliqueNearClipProjection(
```
bdero, line 188: For an orthographic source, the oblique clip replaces projection row 2 with the mirror plane, but `ViewVector` and `SplatGeometry` read that row as the camera forward when the w row is constant. In the capture pass v_viewvector, linear depth, specular, and splat sorting all run along the mirror normal instead of the reflected camera's forward (an isometric camera over a floor mirror shows it). Passing forward explicitly instead of recovering it from the view-projection would avoid this.

packages/flutter_scene/lib/src/camera_controllers/orbit_camera_controller.dart
```diff
@@ -160,14 +172,27 @@ class OrbitCameraController extends CameraController {
...
     _target += (_targetGoal - _target) * r;
-    node.lookAtFrom(_eyeFor(_target), _target);
+    var eyeDistance = _distance;
+    final projection = node.getComponent<CameraComponent>()?.projection;
+    if (projection is OrthographicProjection) {
+      final reference = _orthographicReference ??= (
+        distance: _distance,
+        zoom: projection.zoom,
+      );
+      projection.zoom = reference.zoom * reference.distance / _distance;
```
bdero: This rewrites `zoom` from the first captured reference on every update, so setting `projection.zoom` in code or `orthographicZoom` in the inspector, or assigning a new `OrthographicProjection`, is undone the next frame. The old reference zoom also carries over to a replaced projection.

packages/flutter_scene/lib/src/fscene/realize/builtin_codecs.dart
```diff
@@ -1357,25 +1382,103 @@ class CameraCodec extends DeclarativeComponentCodec<CameraComponent> {
     ComponentField.number(
       'near',
```
bdero, line 1383: Dropping `Range(0.0001, null)` here lets a perspective camera take `near <= 0` from the editor or a document, which asserts in `_matrix4Perspective` in debug and degenerates depth in release. Could perspective keep the constraint while orthographic allows negative values? `far` lost its range too.

bdero (commented):
A few issues with orthographic shading, reflections, and the orbit controller. Details inline.

packages/flutter_scene/shaders/view_vector.glsl
bdero: Restored `v_viewvector` to the true eye vector, so the cotangent frame derivatives are correct under every projection. The orthographic view direction now comes from a `ViewInfo` block behind `GetViewDirection()`, fixed in b8890d14.

bdero: Lighting, clearcoat, fog, and the sky sample now use `GetViewDirection()`, which returns `-forward` under ortho, and billboards and depth bias use a viewer-direction helper. A separate varying would have broken raw `ShaderMaterial` vertex stages, which must write exactly the seven engine varyings.

packages/flutter_scene/lib/src/render/planar_reflection.dart
bdero: Forward now comes from the cross product of the view-projection x and y rows, which the oblique clip and jitter never touch, so `Geometry.bind` stays unchanged for custom geometry. `ProjectionParams` also unwraps the oblique projection for the base near and far.

packages/flutter_scene/lib/src/camera_controllers/orbit_camera_controller.dart
bdero: Dolly now scales the current zoom by that frame's distance change only, so external zoom edits survive, and a newly assigned projection resets the controller state. Both have tests in b8890d14.

packages/flutter_scene/lib/src/fscene/realize/builtin_codecs.dart
bdero: Done in b8890d14. Perspective `near` and `far` keep their positive range, and orthographic has separate `orthographicNear` and `orthographicFar` fields.

packages/flutter_scene/lib/src/material/engine_lighting.dart
```diff
@@ -320,6 +334,36 @@ class EngineLightingUniforms {
   /// from [lighting]'s fog and directional light, and binds it on [shader].
   /// Every material shader declares `FogInfo`, so this is always bound; when
   /// there is no fog the `enabled` flag is 0 and `ApplyFog` is a no-op.
+  /// Binds the `ViewInfo` block `material_varyings.glsl` declares for
```
bdero, line 333: `bindViewInfo` landed between `bindFog`'s doc comment and its declaration, so this doc now opens with the `FogInfo` packing text and `bindFog` has none. Moving the four `FogInfo` lines back above `static void bindFog` fixes both.

bdero (commented):
The earlier comments all look resolved. One small doc fix inline.

packages/flutter_scene/lib/src/material/engine_lighting.dart
bdero: Moved the `FogInfo` doc back above `bindFog`, so each method has its own comment again. Done in 9a6f59e2.

PR #413, opened 2026-09-18
Web: upload glTF textures without `Image.toByteData`

xvrh:
Hi, I was running into some issues with our 3D viewer running on a web build with `--wasm`. I don't have any knowledge in 3D and WebGL, so this is all advised by AI :-) I leave you judging if this make sense. Thanks a lot for this awesome package.

### Problem

On the web, the glTF importer uploads a texture in five steps:

1. The browser decodes the image.
2. The result becomes a `ui.Image`.
3. `Image.toByteData` reads the pixels back.
4. `generateMipChain` builds the mips on the CPU.
5. Each level is uploaded with `texSubImage2D`.

This causes two problems on a Wasm build.

**The pixels can be wrong.** `toByteData` reads back through Flutter's own render surface. Flutter has two open bugs there:

- A read-back larger than 5760 × 5760 is clipped (flutter/flutter#182476).
- A frame drawn during a read-back corrupts it. An animated loading spinner is enough. flutter/flutter#192770 is a draft fix.

I hit both with a model that has two 8192² textures. It rendered in pieces. No error or warning was reported. The same model renders correctly with CanvasKit and on native.

**The main thread is blocked.** The web has no isolates, so every step runs on the main thread. I measured two steps alone, on a Wasm release build:

| texture | `toByteData` | `generateMipChain` |
|---|---|---|
| 2048² | not measured | 0.7 s |
| 4096² | 0.5 s | 2.2 s |
| 8192² | 8.1 s | 8.8 s |

### Change

The web backend already has its own WebGL2 context. This PR uploads the texture there directly: `createImageBitmap`, then `texSubImage2D`, then `generateMipmap`. Flutter is not involved. The other backends return null and keep the current `dart:ui` path.

The PR also adds `Node.fromGlbBytes(maxTextureSize:)`. A larger texture is scaled down while it is decoded. The web uses `resizeWidth`. The other backends use `instantiateImageCodecWithSize`. An 8192² RGBA texture with its mips uses 341 MB of GPU memory, so an app may want a limit.

### Results

Model: 157 MB GLB, two 8192² PNG textures, 1.35 M triangles. Release builds, Chrome 152, macOS.

| | before | after |
|---|---|---|
| Wasm, import time | 28 s | 4.3 s |
| Wasm, render | in pieces | correct |
| Wasm, longest main-thread stall | 14 s | 2.1 s |
| CanvasKit, import time | 17 s | 2.3 s |

The remaining 2.1 s stall comes from the mesh, not from the textures. It is the same at 2048² and at 8192². It does not appear with CanvasKit.

A model with one 2048² JPEG imports in about 150 ms.

During the "after" runs, a second spinner was animating and I was scrolling the page. The render stayed correct.

### Open question: mip quality

Color textures use `r8g8b8a8UNormInt`. The shader converts sRGB to linear when it samples. So `generateMipChain` averages in linear light. `gl.generateMipmap` averages the stored sRGB values instead. Minified color textures should come out slightly darker. I have not compared the two results visually.

If this is not acceptable, I see two options:

- Upload color textures as `SRGB8_ALPHA8`, and remove the shader conversion for them.
- Build the mips with a small downsample pass that converts to linear first.

Normal maps keep the CPU path, because their mips must be renormalized.

`maxTextureSize` does not depend on the new upload path. I can move it to its own PR.

### Notes

- Only the runtime glTF importer changes. `.fscene` textures (`resource_realizer.dart`) and `gpuTextureFromImage` also use `toByteData`. They have the same two problems on Wasm. I left them out to keep this PR small.
- Under dart2wasm, `gl.getParameter()` does not return a Dart `num`, so `value is num` is false. `gpu_context.dart` has this pattern for `MAX_COMBINED_TEXTURE_IMAGE_UNITS`. It falls back to 32 there, which is harmless.
- On this branch, the analyzer is clean and `flutter test` passes (2168 passed, 39 skipped). Those tests run without a GPU, so they do not exercise the new web path.
- I tested the web path by hand, on the 0.23.0 sources plus this commit. The numbers above come from those runs. I did not build a web app against `master`, and I did not run `examples/smoke_render`.
- CHANGELOG: two lines under 0.24.0.


bdero:
Thanks for this, the writeup and measurements were spot on. I pushed a follow-up commit on top of yours and merged. The web backend now builds the mip chain itself with the engine's own filter, so minified textures match the other backends and normal maps take the direct path too. `.fscene` textures go through the same route, `maxTextureSize` stayed in, and your note about `getParameter` under dart2wasm was right and is fixed as well.

PR #414, opened 2026-09-19
Normalize skin weights in the skinned vertex shaders

itsJoKr:
## Problem

glTF requires a vertex's four skin weights to sum to 1, but real exports miss by a fraction of a percent. three.js's [`Soldier.glb`](https://raw.githubusercontent.com/mrdoob/three.js/dev/examples/models/gltf/Soldier.glb) (the Mixamo Vanguard, one skinned mesh over a 49-joint rig) has **52 of 7,325 vertices** with weight sums as low as **0.9803**. It renders fine in three.js because `GLTFLoader` calls `mesh.normalizeSkinWeights()` on every skinned mesh; Unity and Unreal normalize at import.

In flutter_scene the same file grows long black spikes as the character moves away from the origin:

24 m, 3 km, 163 km

*(left to right: 163 km, 24 m, 3 km — same model, same frame of its run cycle)*

## Why it is worse here than in other engines

A joint matrix is `joint.globalTransform * inverseBindMatrices[i]`, so its **translation column carries the model's world position**, and the shader's weighted sum scales that translation along with the rotation. A vertex whose weights sum to 0.98 is placed at 98% of the world position:

| distance from origin | displacement of a 0.98 vertex |
|---|---|
| 24 m | 0.5 m (hides behind the character as a dark wedge) |
| 3 km | 59 m |
| 163 km | 3.2 km |

Engines that skin in model space and apply the world transform afterwards (Babylon, Unity, Unreal, Godot) shrink the same vertex by 2% of its *local* offset — a couple of centimetres anywhere in the world. Here the strays stay anchored near the origin while the body travels, so the mesh looks perfect at the start and progressively shatters. The joint matrices themselves were verified correct on the CPU (all 49 translations at 3000 ± 2 m at 3 km); the error is entirely the unnormalized sum.

## Fix

Normalize the four influences in `flutter_scene_skinned_body.glsl` (and identically in `flutter_scene_velocity_skinned.vert`, so the velocity pass deforms to the same positions and stray vertices don't report bogus motion vectors). A zero sum falls back to the first joint. A few ALU ops per skinned vertex.

Normalizing in the shader rather than at import defends against every asset, including ones loaded through the runtime glTF path, and matches what the loaders of the other engines already do on the user's behalf.

## Verification

Before/after at 3 km on the OpenGLES emulator (fp32, so this is not a precision effect):

before/after 3 km

Also verified at 24 m and 163 km, and on a Pixel 7a (Mali-G710) and OnePlus CPH2769 (Adreno 829), both on the Vulkan backend. The engine's frame-shot baselines near the origin changed only where the 0.5 m wedge had been.

To check a model offline: decode `WEIGHTS_0` and sum the four components per vertex; anything off 1.0 by more than ~1e-3 will show this at distance.


bdero:
Thanks, this landed in #418 with your commit and authorship intact, alongside a smoke scene that fails without the fix. It ships in 0.24.

PR #415, opened 2026-09-19
Give the joints texture a 16-texel floor

itsJoKr:
## Problem

A skin with few joints renders its later joints smeared across the scene. Seen with three.js's [`Soldier.glb`](https://raw.githubusercontent.com/mrdoob/three.js/dev/examples/models/gltf/Soldier.glb), which has a 49-joint body skin and a **2-joint visor skin** in the same file. The body rendered correctly; the visor's vertices weighted to its second joint drew as long dark blades:

visor before

`Skin._jointsTextureEdge` sizes the RGBA32F joints texture to the joint count: `max(4, nextPow2(ceil(sqrt(joints * 4))))`. The 49-joint skin gets 16×16; the 2-joint skin gets **4×4**, with joint 0 in row 0 and joint 1 in row 1. Everything past row 0 read wrong.

## What was established

- Hiding the visor node removed the blades; nothing else changed.
- Raising the floor to 16 fixed them with the visor visible, no other change.
- The 49-joint skin (already 16×16) was correct throughout.
- Reproduced on the **OpenGLES emulator at fp32**, so this is a data-layout problem, not precision.
- Vertex data checked offline: joint indices within range, weights summing to 1, layout `skinned_uv1_tangent` at 104 B/vertex, exactly 109 vertices — the input is correct.

Any skin under 17 joints used an edge below 16 (edge 4 up to 4 joints, edge 8 up to 16); full character rigs never did, which is why it surfaces on a secondary skin — a visor, a prop, a short bone chain — rather than the character.

## Mechanism (hypothesis, stated as such in the comment)

The likely cause is a row-pitch expectation in the texture upload path: at edge 16 an RGBA32F row is exactly 256 bytes, a common alignment unit, and every larger power of two is a multiple of it, while a 4×4 texture's rows are 64 bytes. I have **not** traced this through the backend, and edge 8 was not tested — the doc comment separates the observation and the fix (proven) from the mechanism (likely). If you know the actual constraint, the comment should say that instead.

## Fix

`max(4, …)` → `max(16, …)`. Costs 4 KiB per skin, which every skin of 17 or more joints already paid.

## Verification

Same frame after the change, visor visible:

visor after


bdero:
Thanks for digging into this one. I could not reproduce it, and I would like to before taking the floor, since the comment would otherwise be describing a mechanism we cannot see.

What I checked: the `skinned_animation` smoke scene already draws a 2-joint skin through a 4x4 joints texture with its second joint on row 1, and it renders correctly on every CI lane (Metal, Mesa llvmpipe, SwANGLE, SwiftShader Vulkan, WebGL2, Windows). The Soldier visor also renders correctly at 3 km on Metal and on the Android emulator with host GPU translation under both GLES and Vulkan, with only #414 applied. On the engine side, the Vulkan upload copies from a tightly packed staging buffer into an optimal-tiling image, and the GLES upload calls `glTexImage2D` with unpack alignment 1, so neither has a row-pitch dependency.

Could you share which host OS and emulator GPU mode (`-gpu host`, `swiftshader_indirect`, `swangle_indirect`) you were on, and whether you can reproduce the visor blades on the Pixel 7a or the OnePlus with #414 applied? A real-device reproduction, or a second host, would make this a documented driver workaround with a known scope and I would add it gladly. For now I am leaving it out of #418.

itsJoKr:
Yeah, I think this might be that #414 issue interfering and agent somehow got the wrong clue here. I couldn't reproduce it on any gpu in isolated test. Thanks for reviewing those other PRs, I'm closing this one.

PR #416, opened 2026-09-19
Write the normal and tangent varyings at unit length

itsJoKr:
## Problem

A model authored at centimetre scale under a 0.01 root node — the shape of many glTF exports, including three.js's [`Soldier.glb`](https://raw.githubusercontent.com/mrdoob/three.js/dev/examples/models/gltf/Soldier.glb) — lights as **solid black** on an Adreno 829 (OnePlus CPH2769, Vulkan backend): correct silhouette and shadow, only faint Fresnel rims. The same scene on the Android emulator (OpenGLES) renders correctly, as does a 45×-scaled model on the same phone. A Mali-G710 (Pixel 7a, Vulkan) is unaffected, so this is GPU-specific.

Adreno 829, before and after

## Diagnosis

The `world_normal` debug view — emulator left, Adreno right. The floor's normals are fine on both; the character's are **zero** on the Adreno.

world_normal, emulator vs phone

Ruled out first: `baseColorFactor` set to red reached the fragment shader (uniforms bind), removing the textures changed nothing (not sampling), the render targets scanned clean for NaN/Inf, vertex input locations and set/binding tables are identical between the GLES and Vulkan targets, and vertex colours are 1.0.

The chain:

1. `WorldNormalMatrix` returns the **cofactor** matrix, whose entries scale as the **square of the model scale**; its comment relies on "the later normalize" to remove the magnitude, and that normalize is in the fragment stage.
2. At a root scale of 0.01 (× a 0.98 fit) the world normal leaves the vertex stage at magnitude ≈ **1e-4**.
3. `v_normal` and `v_tangent` are **mediump varyings**; fp16's smallest normal value is 6.1e-5, so most components of a 1e-4 vector sit in the denormal range. The Adreno reads them back as zero, the fragment stage normalizes a zero vector, and every lighting term that depends on the normal collapses — only the specular Fresnel rims survive. The Mali keeps those values.

Metal and desktop GL stay fp32 and never show it. Skinning is incidental — any small-scaled model on that GPU does this.

## Fix

Write both varyings at unit length in the vertex stage (`UnitOrZero` in `normal_transform.glsl`, zero-preserving so a mesh without tangents keeps its "no tangent frame" signal), in both the unskinned and skinned bodies. Unit vectors are what fp16 represents well regardless of model scale. `PRECISION.md` gains the corresponding bullet.

## Verification

- Adreno 829: black → lit, same device and frame (above); also correct at 3 km.
- No regression on Mali-G710 at 24 m and 3 km, or on the OpenGLES emulator, where the engine's pinned frame-shot baselines were unchanged.


bdero:
Thanks, this landed in #418 with your commit and authorship intact. It ships in 0.24.

PR #418, opened 2026-09-19
Normalize skin weights and write unit-length normal varyings

bdero:
Supersedes #414 and #416, carrying both fixes with the contributor's commits. Skinned vertex shaders now normalize the four joint weights, since a joint matrix carries the model's world position and a weight sum under 1 pulls the vertex toward the origin by a growing distance as the model moves away from it. The vertex stage also writes the normal and tangent varyings at unit length, so a model authored at a small scale no longer lights black on GPUs that flush its tiny normals to zero in mediump varyings.

Adds a `skinned_weight_sum` smoke scene, a synthetic skinned mesh with weights summing to 0.98 placed 3 km from the origin, which fails without the first fix.


packages/flutter_scene/shaders/flutter_scene_unskinned_body.glsl
```diff
@@ -80,11 +80,12 @@ void main() {
       frame_info.camera_position, frame_info.depth_bias);
   gl_Position = frame_info.camera_transform * vec4(draw_position, 1.0);
   v_viewvector = frame_info.camera_position - vertex.world_position;
-  v_normal = vertex.world_normal;
+  // Unit length before interpolation (UnitOrZero, normal_transform.glsl).
+  v_normal = UnitOrZero(vertex.world_normal);
```
bdero, line 84: [P3] Update the vertex output contract

`material_vertex.glsl` still describes `v_normal` as not normalized, while this line and `MATERIALS.md` now define the vertex output as unit length before interpolation. Please update that declaration comment so future shader work does not follow the old contract.

bdero (commented):
Reviewed the shader math, material contract, synthetic fixture, and smoke results. The behavior is sound across the rendered backends. One stale declaration comment remains inline.

packages/flutter_scene/shaders/flutter_scene_unskinned_body.glsl
bdero: Done in cf01c4c6. The fragment-side declaration in material_varyings.glsl and the normal-length debug view comment said the same thing, so they follow.

PR #419, opened 2026-09-20
Composite display-referred surfaces past the tone curve

### LINKED ISSUE #382: WidgetComponent texture capture applies a non-linear tone curve, crushing dark colours (64 -> 35)
## Summary

Colours rendered through `WidgetComponent` arrive darker than the widget
painted them, and the loss is much worse for dark tones. A flat
`ColoredBox(color: grey)` hosted in a `WidgetComponent`, rendered by a
`SceneView`, and read back with `RenderRepaintBoundary.toImage` comes back:

| input grey | output | ratio |
|---|---|---|
| 255 | 240 | 0.941 |
| 187 | 180 | 0.963 |
| 128 | 116 | 0.906 |
| 64 | 35 | 0.547 |

The ratio is not constant, so this is not a uniform dimming a brightness
multiplier could correct. It collapses toward black, which is the usual
signature of an sRGB/linear mismatch somewhere between the widget
rasterization and the sampled texture.

## What we ruled out

- **The material.** Binding the identical captured texture to an opaque
  `UnlitMaterial` (`alphaMode = AlphaMode.opaque`) instead of the alpha-blended
  material `WidgetComponent` creates by default gives byte-identical output at
  all four inputs. The colour is already wrong when it is bound.
- **The platform.** Reproduced identically on macOS (Impeller) and on the web
  WebGL2 backend, which are separate implementations.
- **The machine.** Reproduced exactly, same four output values, on two
  machines with different CPU architectures and on Flutter 3.47.1 and 3.47.2.
- **Application code.** The reproduction below involves none: the inputs are a
  `ColoredBox` and a size.

## Reproduction

flutter_scene 0.22.2, Flutter 3.47.1 and 3.47.2, macOS Impeller and web
WebGL2. (0.22.2 rather than 0.23.0 because on 0.23.0 `WidgetComponent` draws
nothing at all for us, on both backends, which may deserve its own issue.)

The integration test below is self-contained (about sixty lines): for each
grey in {255, 187, 128, 64} it hosts a flat `ColoredBox` in a
`WidgetComponent` with `pixelRatio: 1`, renders it through a `SceneView`,
reads the pixels back with `toImage`, and takes the modal colour. It asserts
every output/input ratio stays above 0.85, and fails today at input 64 with
0.547.

Expected: a widget hosted in a `WidgetComponent` keeps its colours, the way
the same widget does when painted directly.

Actual: the table above.

## Why it matters

Any UI textured through `WidgetComponent` is hit hardest exactly where text
and line work live, near black. In our case near-black digits (value 27)
arrive at 2, on an otherwise pixel-faithful render.

## The test

<details>
<summary>texture_tone_test.dart, self-contained</summary>

```dart
import 'dart:typed_data';
import 'dart:ui' as ui;

import 'package:flutter/material.dart';
import 'package:flutter/rendering.dart';
import 'package:flutter_scene/scene.dart';
import 'package:flutter_test/flutter_test.dart';
import 'package:integration_test/integration_test.dart';
import 'package:vector_math/vector_math.dart' as vm;

void main() {
  IntegrationTestWidgetsFlutterBinding.ensureInitialized();

  testWidgets('measure what the texture path does to a colour', (tester) async {
    await tester.pumpWidget(const SizedBox());
    await Scene.initializeStaticResources();
    expect(Scene.isReadyToRender, isTrue,
        reason: 'without a GPU this measures nothing');

    /// Renders a flat [input] through a widget surface and returns the most
    /// common colour that comes back.
    Future<int> through(Color input, {bool opaque = false}) async {
      final scene = Scene();
      // Same capture either way. The only variable is the material the
      // texture is bound to, which separates a capture-side transform from a
      // sampling-side one: WidgetComponent owns an alpha-blended UnlitMaterial
      // when none is given, and blending is the obvious suspect for a
      // brightness change that worsens toward black.
      final surface = WidgetComponent(
        child: ColoredBox(color: input),
        size: const Size(200, 200),
        pixelRatio: 1,
        worldHeight: 2.0,
        update: WidgetUpdatePolicy.everyFrame,
        material: opaque
            ? (UnlitMaterial()..alphaMode = AlphaMode.opaque)
            : null,
      );
      scene.add(Node()..addComponent(surface));

      final key = GlobalKey();
      await tester.pumpWidget(MaterialApp(
        home: Scaffold(
          body: Center(
            child: RepaintBoundary(
              key: key,
              child: SizedBox(
                width: 200,
                height: 200,
                child: SceneView(
                  scene,
                  camera: PerspectiveCamera(
                    position: vm.Vector3(0, 0, 2.42),
                    target: vm.Vector3.zero(),
                  ),
                ),
              ),
            ),
          ),
        ),
      ));
      for (var i = 0; i < 20; i++) {
        await tester.pump(const Duration(milliseconds: 100));
      }

      final boundary =
          key.currentContext!.findRenderObject()! as RenderRepaintBoundary;
      late ByteData data;
      await tester.runAsync(() async {
        final image = await boundary.toImage();
        data = (await image.toByteData(format: ui.ImageByteFormat.rawRgba))!;
        image.dispose();
      });

      // The quad does not fill the box exactly, so take the mode rather than
      // a sample: the surface is the largest flat region.
      final counts = <int, int>{};
      for (var i = 0; i < data.lengthInBytes; i += 4) {
        final c = data.getUint32(i);
        counts[c] = (counts[c] ?? 0) + 1;
      }
      final top = counts.entries.reduce((a, b) => a.value >= b.value ? a : b);
      return top.key;
    }

    final ratios = <int, double>{};
    for (final v in const [255, 187, 128, 64]) {
      final input = Color.fromARGB(255, v, v, v);

      final blended = await through(input);
      final br = (blended >> 24) & 0xFF;
      ratios[v] = br / v;

      final opaque = await through(input, opaque: true);
      final orr = (opaque >> 24) & 0xFF;

      // ignore: avoid_print
      print('TONE in=$v  blend=$br (${(br / v).toStringAsFixed(3)})  '
          'opaque=$orr (${(orr / v).toStringAsFixed(3)})');
    }

    // Measured 2026-08-26 on flutter_scene 0.22.2:
    //   255 -> 240 (0.941), 187 -> 180 (0.963), 128 -> 116 (0.906),
    //    64 ->  35 (0.547)
    //
    // The ratio is not constant, so this is not a uniform dimming. It
    // collapses in the dark tones, which is the signature of a colour space
    // mismatch, and it is why the board's near-black digits arrive at 2
    // instead of 27.
    //
    // These assertions FAIL today on purpose. They describe what a correct
    // texture path would do, and the gap between them and the numbers above
    // is the defect. Do not relax them to make the suite green; that would
    // delete the only measurement of this.
    for (final entry in ratios.entries) {
      expect(entry.value, greaterThan(0.85),
          reason: 'input ${entry.key} came back at '
              '${(entry.value * 100).toStringAsFixed(0)} percent. The texture '
              'path is crushing this tone. Dark values suffer most, so digits '
              'and gridlines are affected more than the board surface.');
    }
  }, timeout: const Timeout(Duration(minutes: 8)));
}

```

</details>

bdero:
Resolves #382.

A captured widget is display-referred, but the renderer treated it as scene-referred radiance, so the tone mapping operator transformed it on the way out. The issue attributes the loss to an sRGB/linear mismatch in the capture path, but that path cancels exactly; the whole table is PBR Neutral's black-point term, which costs a near-black widget most of its value.

`Material.displayReferred` marks a surface whose color is already final. Those draws leave the opaque and translucent buckets for a layer drawn after the scene against its depth attachment, which a new display step composites onto the resolved image. `WidgetComponent` turns it on for the material it owns, so a captured widget keeps its colors by default; a widget bound to a material you supply is unchanged.

Pre-compensating by the inverse tone curve was the other candidate and does not work. The operators gamut-map, so saturated colors have no non-negative pre-image and come back worse than doing nothing (pure red renders as a washed pink). Compositing past the curve is exact for every color under every operator instead.


packages/flutter_scene/lib/src/render/display_referred_pass.dart
```diff
@@ -0,0 +1,102 @@
...
+
+/// Render-graph blackboard key for the display-referred layer [ScenePass]
+/// draws, which [DisplayReferredCompositePass] blends onto the resolved image.
+const String kDisplayReferredBlackboardKey = 'display_referred_color';
+
+/// Whether the scene holds a visible display-referred surface (see
+/// [Material.displayReferred]), which the frame pays the extra layer for.
+bool sceneHasDisplayReferred(RenderScene renderScene) {
+  for (final item in renderScene.items) {
+    if (item.visible && item.material.displayReferred) return true;
```
bdero: [P1] Include LOD materials in the activation scan

`SceneEncoder` selects `LodLevel.material`, but this scan only checks the highest-detail fallback `item.material`. If level 0 is scene-referred and the selected lower level is display-referred, the encoder removes that draw from both ordinary buckets while `displayReferredActive` stays false, so neither the layer nor composite runs and the object disappears. Scan the LOD materials too, ideally using the same per-view eligibility rules as the encoder.

packages/flutter_scene/lib/src/render/scene_pass.dart
```diff
@@ -426,6 +448,7 @@ class ScenePass extends RenderGraphPass {
         );
       }
       rendererSubmissions.submit(commandBuffer);
+      _encodeDisplayReferredLayer(context, encoder, depth, width, height);
```
bdero, line 458: [P1] Do not composite a jittered layer after TAA

When TAA is active this encoder was constructed with `currentJitteredViewProjection`. Flushing the display records here therefore produces a current-frame jittered layer, but TAA only resolves `kSceneColorBlackboardKey` and the layer is composited afterward. The widget will shift by the jitter every frame and its occlusion edge will not align with the resolved scene. Render this layer with an unjittered projection/depth path or include it in the temporal resolve.

packages/flutter_scene/lib/src/scene.dart
```diff
@@ -3240,6 +3246,15 @@ base class Scene implements SceneGraph {
       );
     }
 
+    // Display-referred surfaces composite after anti-aliasing so crisp UI is
```
bdero: [P2] Preserve edge antialiasing outside MSAA

This composite runs after FXAA/SMAA/TAA, while `_encodeDisplayReferredLayer` multisamples only when `_enableMsaa` is true. In every non-MSAA mode the layer's 3D silhouette is single-sampled and no later pass antialiases it, contrary to the `DisplayReferredCompositePass` comment. Rotated widget surfaces will gain jagged edges. Give the layer coverage antialiasing in these modes or place its silhouette before the selected antialiasing pass.

bdero (commented):
Found correctness gaps in LOD activation and TAA, plus unsupported antialiasing, debug, persistence, and CI cases. Details are inline.

packages/flutter_scene/shaders/flutter_scene_unlit.frag
```diff
@@ -46,12 +49,23 @@ void main() {
   // Linearize the sRGB-encoded base color so what we write to the
   // floating-point scene-color target is linear; the tone-mapping resolve
   // pass applies display encoding. Output is premultiplied by alpha.
-  vec3 rgb = SRGBToLinear(base.rgb) * vertex_color.rgb * frag_info.color.rgb;
+  //
+  // A display-referred surface skips that: it draws into the display-referred
+  // layer, which is composited after the resolve has already encoded, so its
+  // texel has to arrive unchanged.
+  bool display_referred = frag_info.display_referred > 0.5;
+  vec3 decoded = display_referred ? base.rgb : SRGBToLinear(base.rgb);
```
bdero, line 57: [P2] Keep debug inputs linear

`rgb` is display-encoded in this branch, but below it is assigned to `material.base_color`, and `DebugSurfaceOutput` treats that field as linear and calls `DebugColor`. A base-color or validation view will encode display-referred colors a second time, so 0.5 becomes about 0.735. Convert the composed display color back to linear for `MaterialInputs` while keeping `shaded` display-encoded.

packages/flutter_scene/lib/src/components/widget_component.dart
```diff
@@ -67,6 +67,7 @@ class WidgetComponent extends Component {
     Geometry? geometry,
     Material? material,
     void Function(gpu.Texture texture)? bind,
+    this.displayReferred = true,
```
bdero, line 70: [P2] Persist the widget setting

`WidgetCodec` serializes the configurable owned-surface fields but has no `displayReferred` field, and `create()` never reads one. A scene that opts out with `displayReferred: false` cannot round-trip this behavior and reloads with the new `true` default. Add it to the component schema and its round-trip test.

packages/flutter_scene/lib/src/material/unlit_material.dart
```diff
@@ -65,6 +65,21 @@ class UnlitMaterial extends Material {
...
+  /// resolved image, so an authored color arrives on screen unchanged. Use it
+  /// for captured widgets and other UI textured into the scene;
+  /// `WidgetComponent` sets it on the material it owns. The surface keeps its
+  /// depth test against the scene but takes no exposure, grading, tone
+  /// mapping, fog or bloom, and does not order against translucent geometry.
+  ///
+  /// With this set, [baseColorFactor] multiplies in display space rather than
+  /// linear space.
+  @override
+  bool displayReferred = false;
```
bdero, line 81: [P2] Persist the material setting

`_serializeUnlit` and `_unlit` do not write or read this flag. A display-referred `UnlitMaterial` saved as an fscene resource reloads scene-referred and silently changes both its render bucket and color. Add a boolean property with a false default and a round-trip test.

examples/smoke_render/integration_test/display_referred_fidelity_test.dart
```diff
@@ -0,0 +1,150 @@
+// Colour fidelity of a display-referred surface (issue #382). A flat sRGB
```
bdero: [P2] Wire this test into the smoke matrix

Every GitHub and Codemagic smoke job explicitly targets `integration_test/smoke_test.dart`, so none discovers this new file. It will not run in CI despite being the only pixel-level test for this feature. Fold it into `smoke_test.dart` or add this target to every backend job.

packages/flutter_scene/lib/src/render/display_referred_pass.dart
bdero: Fixed in b242c641. The scan now counts any level's material. The selected level depends on the view, which isn't knowable there, so it over-activates (an unused layer) rather than risk routing a draw out of both buckets.

packages/flutter_scene/lib/src/render/scene_pass.dart
bdero: Fixed in b242c641. The layer draws through a separate unjittered transform the encoder swaps in for that flush only, so culling and screen bounds stay on the scene's own frustum. Its depth test still runs against the jittered scene depth, which costs sub-pixel accuracy at the occlusion edge; noted in the code.

packages/flutter_scene/lib/src/scene.dart
bdero: Fixed in b242c641. The composite's placement is now conditional: after anti-aliasing under MSAA, before FXAA/SMAA otherwise. Non-MSAA modes trade some softening of the UI for a smoothed silhouette. The pass comment described the MSAA case as if it were universal and now says both.

packages/flutter_scene/shaders/flutter_scene_unlit.frag
bdero: Fixed in b242c641. The composed display color is converted back to linear before filling MaterialInputs, leaving `shaded` display-encoded.

packages/flutter_scene/lib/src/components/widget_component.dart
bdero: Fixed in b242c641. Added to the component schema with a true default matching the constructor, so an opt-out persists and an absent property does not flip. Round-trip coverage in ui_codecs_test.dart.

packages/flutter_scene/lib/src/material/unlit_material.dart
bdero: Fixed in b242c641. Written only when set and read back with a false default, with a round-trip test in test/fscene/unlit_display_referred_test.dart.

examples/smoke_render/integration_test/display_referred_fidelity_test.dart
bdero: Fixed in b242c641. Folded into smoke_test.dart rather than added as a target, so every backend job picks it up. It reuses the mutate-and-repump pattern from the depth_pairing test and runs fewer frames on Android.

